PR #25, opened 2022-11-23
remove autogenerated files

thomas-fossati:
Signed-off-by: Thomas Fossati <thomas.fossati@arm.com>


yogeshbdeshpande (approved):
LGTM!

nedmsmith:
Should the examples be updated too? This change will break working examples.

thomas-fossati:
> Should the examples be updated too? This change will break working examples.

mmm in theory it shouldn't: the examples Makefile targets are predicated on the existence of the autogen files.  Can you reproduce the problem?

nedmsmith:
> Can you reproduce the problem?

In the example cbor diag below from (https://github.com/ietf-rats-wg/draft-ietf-rats-corim/blob/main/cddl/examples/comid-1.diag), the array at  `/ measurement-map / [ {` and at the end  `} ]` should be removed since the array was removed from the triple definition.
```
},
      / measurement-map / [ {
        / comid.mval / 1 : {
          / comid.ver / 0 : {
            / comid.version / 0 : "1.0.0",
            / comid.version-scheme / 1 : 16384 / semver /
          },
          / comid.digests / 2 : [ [
            / hash-alg-id / 1, / sha256 /
            / hash-value / h'44aa336af4cb14a879432e53dd6571c7fa9bccafb75f488259262d6ea3a4d91b'
          ] ]
        }
      } ]
```

PR #57, opened 2023-02-27
svn-type-choice #8

### LINKED ISSUE #8: provide description of SVN
Describe SVN and "min" SVN.

nedmsmith:
Added prose to section describing svn and min-svn

Fix #8

### DIFF
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -825,6 +825,12 @@ $version-scheme /= int / text
 
 [^issue] https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/8
 
+The following describes details the security version number (svn) and the minimum security version number (min-svn). A security version number is used to track changes to an object that are security relevant. Rollback of a security relevant change is considered to be an attack vector, as such, svn values can't be decremented. If a security relevant flaw is discovered in an object, the object is updated with a fix and the svn value is incremented. 
+
+There may be several versions of an object that are in production use as a given time. If there are multiple versions having different svn values, the object with a lower svn value may or may not be in a security critical condition. The Endorser may specificy a minimum svn value (min-svn) to specify the lowest svn value that is acceptable. Objects with svn values that are equal to or greater than min-svn do not signal a secrity critical condition. Objects that are below the min-svn value are in a security critical condition that is unsafe for normal operations.
+
+The `svn-type-choice` measurement consists of a `tagged-svn` or `tagged-min-svn` value. The `tagged-svn` and `tagged-min-svn` tags are CBOR tags <tobe assigned by IANA CBOR Tag Registry> `#6.552` and `#6.553`. 
+
 ~~~ cddl
 {::include cddl/svn-type-choice.cddl}
 ~~~
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -823,19 +823,17 @@ $version-scheme /= int / text
 
 ###### Security Version Number {#sec-comid-svn}
 
-[^issue] https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/8
-
-The following describes details the security version number (svn) and the minimum security version number (min-svn).
+The following details the security version number (`svn`) and the minimum security version number (`min-svn`) statements.
 A security version number is used to track changes to an object that are security relevant.
-Rollback of a security relevant change is considered to be an attack vector, as such, svn values can't be decremented.
-If a security relevant flaw is discovered in an object, the object is updated with a fix and the svn value is incremented.
+Rollback of a security relevant change is considered to be an attack vector, as such, security version numbers can't be decremented.
+If a security relevant flaw is discovered in the Target Environment and subsequently fiexed, the `svn` value is typically incremented.
 
-There may be several versions of an object that are in production use at a given time.
-If there are multiple versions having different svn values, the object with a lower svn value may
-or may not be in a security critical condition. The Endorser may provide a minimum svn value using `min-svn`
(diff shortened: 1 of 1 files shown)


draft-ietf-rats-corim.md
```diff
@@ -825,6 +825,21 @@ $version-scheme /= int / text
 
 [^issue] https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/8
```
thomas-fossati, line 826: this should go now that we have text.

draft-ietf-rats-corim.md
```diff
@@ -825,6 +825,21 @@ $version-scheme /= int / text
...
 
+The following describes details the security version number (svn) and the minimum security version number (min-svn).
+A security version number is used to track changes to an object that are security relevant.
```
thomas-fossati, line 829: we should precisely define what we mean by "object" here (or possibly use another, more intuitive, term)

draft-ietf-rats-corim.md
```diff
@@ -825,6 +825,21 @@ $version-scheme /= int / text
 
 [^issue] https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/8
 
+The following describes details the security version number (svn) and the minimum security version number (min-svn).
```
thomas-fossati, line 828: There seems to be one or more missing piece(s) of English language here, but I can't tell exactly what :)

* The following describes _in_ details the security [...]
* The following details the security [...]
* The following describes _the_ details _of_ the security [...]

draft-ietf-rats-corim.md
```diff
@@ -825,6 +825,21 @@ $version-scheme /= int / text
...
+A security version number is used to track changes to an object that are security relevant.
+Rollback of a security relevant change is considered to be an attack vector, as such, svn values can't be decremented.
+If a security relevant flaw is discovered in an object, the object is updated with a fix and the svn value is incremented.
+
+There may be several versions of an object that are in production use as a given time.
```
thomas-fossati, line 833:
```suggestion
There may be several versions of an object that are in production use at a given time.
```

draft-ietf-rats-corim.md
```diff
@@ -825,6 +825,21 @@ $version-scheme /= int / text
...
+If there are multiple versions having different svn values, the object with a lower svn value may
+or may not be in a security critical condition. The Endorser may specificy a minimum svn value (min-svn)
+to specify the lowest svn value that is acceptable.
```
thomas-fossati, line 836:
```suggestion
or may not be in a security critical condition. The Endorser may provide a minimum svn value using `min-svn`
to specify the lowest svn value that is acceptable.
```

draft-ietf-rats-corim.md
```diff
@@ -825,6 +825,21 @@ $version-scheme /= int / text
...
+Objects with svn values that are equal to or greater than min-svn do not signal a secrity critical condition.
```
thomas-fossati, line 837:
```suggestion
Objects with svn values that are equal to or greater than `min-svn` do not signal a security critical condition.
```

draft-ietf-rats-corim.md
```diff
@@ -825,6 +825,21 @@ $version-scheme /= int / text
...
+Objects that are below the min-svn value are in a security critical condition that is unsafe for normal operations.
```
thomas-fossati, line 838:
```suggestion
Objects that are below the `min-svn` value are in a security critical condition that is unsafe for normal operations.
```

draft-ietf-rats-corim.md
nedmsmith: I'm open to suggestions. We use Target Environment and `environment-map` to refer to the component that is being attested. The Target Environment could have multiple measurements such as `digest` and `raw-value-group` but these measurements could be describing different objects within the Target Environment. I don't think we have terminology for the object from which a measurement is collected. I like "object" because it is neutral. Alternatives like "component", "device", "module" all have baggage. The O in OID can mean a variety of things without too much baggage. Object does however have baggage with regard to triples structure, but it seems this is obvious when in triples-map context.

thomas-fossati: Maybe we could provide one or more examples of objects, e.g.:
```
A security version number is used to track changes to
an object (e.g., OBJECT_INSTANCE_1, OBJECT_INSTANCE_2)
that are security relevant.
```
which avoids both defining it or finding another term.

nedmsmith: Fixed in branch - this thread is outdated.

draft-ietf-rats-corim.md
nedmsmith: fixed in branch - this thread is outdated.

draft-ietf-rats-corim.md
```diff
@@ -823,7 +823,20 @@ $version-scheme /= int / text
 
 ###### Security Version Number {#sec-comid-svn}
 
-[^issue] https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/8
+The following details the security version number (`svn`) and the minimum security version number (`min-svn`) statements.
+A security version number is used to track changes to an object (e.g., OBJECT_INSTANCE_1, OBJECT_INSTANCE_2, ...) that are security relevant.
```
thomas-fossati, line 827: I meant `OBJECT_INSTANCE_$i` as placeholders for real things.

```suggestion
A security version number is used to track changes to an object (e.g., a Secure Enclave) that are security relevant.
```

draft-ietf-rats-corim.md
thomas-fossati: Ned, if you address this we can merge.

thomas-fossati (approved):
thanks

PR #58, opened 2023-02-27
Tagged Integer Type description

### LINKED ISSUE #5: What is a "tagged integer"?
Provide a description of a "tagged integer", in particular what purposes it serves.

Also, consider maybe a more evocative name.

nedmsmith:
added description for tagged integer in class-id

Fix #5

### DIFF
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -229,10 +229,10 @@ identifier {{-cbor-oids}}.
 
 [^issue] https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/5
 
-Used in connection with `$class-id-type-choice /= tagged-int-type` when the class identifer is
-an integer value. The CBOR tag `#6.551` signifies the integer is to be used to idenfy a CoMID class.
-If the class definition is vendor specific, the `class-map` `vendor` field should be populated.
-
+Used as a class identifier for the environment.  It is expected that the
+integer value is vendor specific rather than globally meaningful.  Therefore,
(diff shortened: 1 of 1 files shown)


draft-ietf-rats-corim.md
```diff
@@ -229,6 +229,10 @@ identifier {{-cbor-oids}}.
 
 [^issue] https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/5
 
+Used in connection with `$class-id-type-choice /= tagged-int-type` when the class identifer is
+an integer value. The CBOR tag `#6.551` signifies the integer is to be used to idenfy a CoMID class.
+If the class definition is vendor specific, the `class-map` `vendor` field should be populated.
+
```
thomas-fossati, line 235: I don't think there's a reason for this to be other than vendor-specific.  Also, I reckon we need to make it a bit more normative.

```suggestion
Used as a class identifier for the environment.  It is expected that the
integer value is vendor specific rather than globally meaningful.  Therefore,
the sibling `vendor` field in the `class-map` MUST be populated to define the
namespace under which the value must be understood.
```

draft-ietf-rats-corim.md
```diff
@@ -229,6 +229,10 @@ identifier {{-cbor-oids}}.
 
 [^issue] https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/5
```
thomas-fossati, line 230: this should be expunged in this PR

thomas-fossati (approved):
thanks!

PR #66, opened 2023-03-07
Andrew draper evidence verification min

andrew-draper:
This change adds a description of verifier processing.

It is based on Yogesh's notes on the appraisal flow, on text in the TCG CoRIM spec and on the branch I wrote which we discussed a few weeks ago.

### DIFF
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -72,7 +72,11 @@ normative:
       ITU-T: Recommendation X.690
     target: https://www.itu.int/rec/T-REC-X.690
   DICE-Layering-Architecture:
-    target: https://trustedcomputinggroup.org/wp-content/uploads/DICE-Layering-Architecture-r19_pub.pdf
+    author:
+      org: "Trusted Computing Group"
+    title: "DICE Layering Architecture"
+     target: https://trustedcomputinggroup.org/wp-content/uploads/DICE-Layering-Architecture-r19_pub.pdf
+    date: July, 2020
   IANA.named-information: named-info
 
 informative:
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -1093,6 +1093,8 @@ All available Concise Reference Integrity Manifests (CoRIMs) are collected. A ve
 
 CoRIMs that are not within their validity period, or that cannot be associated with an authenticated and authorised source MUST be discarded.
 
+CoRIM that are secured by a cryptographic mechanism such as a signature MUST be discarded if the signature chain does not validate successfully.
+
 Other selection criteria MAY be applied.
 For example, if the Evidence format is known in advance, CoRIMs that do not match the expected profile can be readily discarded.
 
@@ -1102,18 +1104,18 @@ The selection process MUST yield at least one usable CoRIM.
 
 All the available Concise Bill Of Material tags (CoBOMs) are collected from the selected CoRIMs.
 
-The collection process MUST yield, at least, one CoBOM.
+If any CoBOM is present then other tags in the CoRIM which are not activated by the CoBOM MUST be ignored.
 
 ### Tags Identification and Validation
 
-The CoBOM(s) are used to locate and collect the listed tags -- i.e., Concise Module Identifiers (CoMID) and/or Concise Software Identifiers (CoSWID) and/or Concise Trust Anchor Stores (CoTS) -- from the selected CoRIMs.
+Potentially valid tags -- i.e., Concise Module Identifiers (CoMID) and/or Concise Software Identifiers (CoSWID) and/or Concise Trust Anchor Stores (CoTS) -- are chosen from the selected CoRIMs.
 
-All the listed tags MUST be successfully located and be syntactically and semantically valid.
+All the extracted tags MUST be successfully located and be syntactically and semantically valid.
 In particular, any internal cross-reference (e.g., CoMID-CoSWID linking triples) MUST be successfully resolved.
 
 ### Appraisal Context Construction
 
-All Reference Values, Endorsed Values and cryptographic verification key material found in the listed tags are extracted and loaded into the Appraisal Context.
+All of the potential tags are loaded into the Appraisal Context. The complete contents of each tag are loaded together, including Reference Values, Endorsed Values and cryptographic verification key material.
 
 This concludes the initialisation phase.
 
@@ -1143,73 +1145,49 @@ Independent of the specific method, the cryptographic integrity of Evidence MUST
 > * How key material is associated with the Attesting Environment
 > * How the Attesting Environment is identified in Evidence
 
+### The Accepted Claims Set
+
+At the end of the Evidence collection process evidence has been converted into a format suitable for appraisal. Verifiers are not required to use this as their internal state, but for the purposes of this document a sample verifier is discussed which uses this format.
+
+The accepted claims set will be matched against CoMID reference values, so this document describes an evidence structure which can be easily matched against these reference values. Each set of evidence contains an environment map providing a namespace, and a measurement-values-map containing one or more entries.
+
+Each entry in the measurement-values-map is a separate piece of evidence describing the environment named by the environment-map. An attestor can provide multiple environment-maps each containing a single values measurement-values map, a single environment-map containing multiple entries in its measurement-values-map, or a combination of these approaches.
+
+If evidence from different sources has the same environment-map then the measurement-values-maps are merged. If both measurement-value-maps being merged contain the same key then the values associated with that key MUST be binary identical. 
+
 ### Transforming SPDM Evidence to a format usable for matching
 
 ***Evidence Binding For SPDM*** describes the process by which evidence in a SPDM MEASUREMENTS response is converted to Evidence suitable for matching using the rules below. The converted evidence is held in evidence triples which have a similar format to reference-triples (their semantics is described below).
 
 ### Transforming DICE Evidence to a format usable for matching
 
-DICE Evidence appears in certificates in the TcbInfo or MultiTcbInfo extension. Each TcbInfo, and each entry in the MultiTcbInfo, is converted to an evidence triple using the rules below.
-
-The verifier SHALL translate each DICE TcbInfo into an evidence-triple as described in this section. The verifier SHALL split each DICE MultiTcbInfo extension into separate TcbInfo object and translate each one into an evidence-triple in the same way.
+DICE Evidence appears in certificates in the TcbInfo or MultiTcbInfo extension. Each TcbInfo, and each entry in the MultiTcbInfo, is converted to an evidence triple using the rules in this section. In a MultiTcbInfo each entry in the sequence is treated as independent and translated into a separate evidence object.
 
 The verifier SHALL translate each field in the TcbInfo into a field in the created endorsed-triple-record
 
-- The TcbInfo `type` field SHALL be copied to the field named `environment-map / comid.class / comid.class-id`
-- The TcbInfo `vendor` field SHALL be copied to the field named `environment-map / comid.class / comid.vendor`
-- The TcbInfo `model` field SHALL be copied to the field named `environment-map / comid.class / comid.model`
-- The TcbInfo `layer` field SHALL be copied to the field named `environment-map / comid.class / comid.layer`
-- The TcbInfo `index` field SHALL be copied to the field named `environment-map / comid.class / comid.index`
+- The TcbInfo `type` field SHALL be copied to the field named `environment-map / class / class-id`
+- The TcbInfo `vendor` field SHALL be copied to the field named `environment-map / class / vendor`
+- The TcbInfo `model` field SHALL be copied to the field named `environment-map / class / model`
+- The TcbInfo `layer` field SHALL be copied to the field named `environment-map / class / layer`
+- The TcbInfo `index` field SHALL be copied to the field named `environment-map / class / index`
 
-- The TcbInfo `version` field SHALL be translated to the field named `measurement-map / comid.mval / comid.ver`
-> TODO: Details of translation
-- The TcbInfo `svn` field SHALL be copied to the field named `measurement-map / comid.mval / comid.svn`
-- The TcbInfo `fwids` field SHALL be translated to the field named `measurement-map / comid.mval / comid.digests`
-> TODO: Details of translation
-- The TcbInfo `flags` field SHALL be translated to the field named `measurement-map / comid.mval / comid.flags`
-> TODO: Details of translation
-- The TcbInfo `vendorInfo` SHALL shall be copied to the field named `measurement-map / comid.mval / comid.raw-value`
-
-
->>> TODO: Insert DICE/SPDM specific stuff here
+- The TcbInfo `version` field SHALL be translated to the field named `measurement-map / mval / version / version`
+- The TcbInfo `svn` field SHALL be copied to the field named `measurement-map / mval / svn`
+- The TcbInfo `fwids` field SHALL be translated to the field named `measurement-map / mval / digests`
+  - Each digest within fwids is translated to a CoMID digest object, with an appropriate algorithm identifier
+- The TcbInfo `flags` field SHALL be translated to the field named `measurement-map / mval / flags`
+  - Each flag is translated independently 
+- The TcbInfo `vendorInfo` SHALL shall be copied to the field named `measurement-map / mval / raw-value`
 
 If there are multiple evidence triples with the same environment map then ...
-
-### The Accepted Claims Set
-
->>> Start review this part
-
-At the end of the Evidence collection process evidence has been converted into a format suitable for appraisal. Verifiers are not required to use this as their internal state, but for the purposes of this document a sample verifier is discussed which uses this format.
-
-The Internal structure is based on ...
 
 ## Evidence appraisal
 
 In the Evidence Appraisal phase, a CoRIM Appraisal Context and an Evidence Appraisal Policy are used to convert the received Evidence from its raw form into a more usable form. This phase may be repeated multiple times .
 The outcome of the appraisal process is summarised in an Attestation Result.
-
->>> Start need to review
-
-The Attestation Result provides a set of trust metrics associated with the appraised Evidence together with any information that can be derived by the Verifier about the Attester via supply chain endorsements.
 The Relying Party application uses the content of the Attestation Result to make its own policy decisions.
 
-We make no assumptions on the specific shape of the Attestation Result, except for its optional ability to include Endorsed Values associated with the appraised Attester that the Verifier has been able to infer from Evidence and the Appraisal Context.
-
-### Matching Evidence against Reference Values
-
-The second step consists in matching Reference Values in the Appraisal Context against Evidence.
-
-The Target Environment associated with Evidence needs to be identified.
-Typically, this is done by extracting the relevant claims from the Evidence claims-set.
-The exact mechanics depend on the specific Evidence format.
-
-For example:
-In PSA, the Verifier extracts the `euid` and `psa-implementation-id` claims from the PSA claims-set.
-In DICE, the Target Environment identifiers are sourced from the `DiceTcbInfo` X.509 extension.
-
-The Target Environment identifier is used to lookup the relevant Reference Values from the Appraisal Context.
-
-Evidence MUST match the identified Reference Values.
+We make no assumptions on the specific shape of the Attestation Result, except for its optional ability to include Evidence from the attestor and Endorsed Values that the Verifier has been able to infer from Evidence and the Appraisal Context.
 
 > A CoRIM profile MUST describe:
 >
@@ -1218,28 +1196,51 @@ Evidence MUST match the identified Reference Values.
 > * How the Target Environment is identified in Evidence
 > * How measurements from Evidence are matched against Reference Values
 
-### Adding Endorsed Values to the Accepted Claims
-
->>> TODO
-
-The third step consists in matching Endorsed Values in the Appraisal Context with Evidence.
-
-The Target and Attested Environment associated with Evidence need to be identified.
-
-Again, this is typically done by extracting the relevant claims from the Evidence claims-set.
-The exact mechanics depend on the specific Evidence format.
-
-Alongside the environment identifiers, the matched Reference Values MAY also be used in selecting applicable Endorsed Values.
-
-Any matched Endorsed Value can be added to the Attestation Result.
-
 > A CoRIM profile MUST describe:
 >
 > * How Endorsed Values are conveyed (e.g., using Endorsed Value triples, or Device Identity triples)
 > * How Endorsed Values are associated with the Target and Attesting Environments
 > * How the Target and Attesting Environments are identified in Evidence
 
->>> End need to review
+The sections below describe the algorithm used to compare CoMID tags.
+
+### Comparing and processing CoMID tags
+
+There can be multiple passes of the CoMID evidence appraisal process.
+
+In each pass, each potential CoMID tag in the Appraisal Context is compared against the accepted claims set (which has been initialised with Evidence from the attestor(s) ). If the reference values in the CoMID tag match the Accepted Claims Set then Endorsements and other values from the tag are added to the Accepted Claims Set.
+
+If any tags matched in a pass through the CoMID appraisal process then another pass is performed. Tags which matched in a previous pass are ignored in later passes of the same appraisal.
+
+A later pass may match Reference Values from the CoMID against Endorsements from matched tags in an earlier pass, this allows CoRIM authors to create hierarchical descriptions of a system.
+
+### Matching Evidence against Reference Values
+
+This section describes the process performed by the verifier to determine whether a candidate CoMID tag matches the Accepted Claims Set. If a tag does not match the accepted claims set then it is silently ignored for this pass, processing continues with the remaining candidate tags.
+
+The verifier enumerates over the reference values in the CoMID tag. If the Accepted Claims Set does not contain an entry with the same environment-map as the reference claim then the tag does not match. Comparison of environment-map is performed using a binary comparison. Verifiers SHOULD convert environment-map to the canonical format before performing the binary comparison.
+
+The verifer locates the corresponding measurement-values-map in the reference value and the Accepted Claims Set. The verifier enumerates over the entries in the reference value measurement-values-map. If the Accepted Claims set measurement-values-map does not have an entry with the same key as the reference value measurement-values-map, or if the entries do not match, then the tag does not match.
+
+The algorithm used to compare the measurement-values-map entries together depends on whether the reference value is tagged, and on the key which identifies the entry.
+
+If the reference value is tagged then the verifier must use the protocol described by the tag to match the entries.
+
+If the reference value is not tagged and the key is a special value described in the sub-sections below, then the protocol appropriate to that key is used to match the entries.
+
+If the reference valus is not tagged, and the key is not a special value described below, then the entries are compared with binary comparison.
+
+#### Comparing raw-value entries
+
+#### Comparing svn entries
+
+#### Comparing digest entries
+
+### Adding CoMID Endorsed Values to the Accepted Claims Set
+
+If a CoMID tag matches the Accepted Claims Set then the verifier must attempt to add Endorsed values from that tag to the Accepted Claims Set.
+
+If Endorsed values being added have the same name and measurement-values-map key, but a different measurement-value-map entry from the existing value.
  
 # Implementation Status
 
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -1115,7 +1115,7 @@ After the verifier has processed all CoBOMs it MUST discard any tags which have 
 
 ### Tags Identification and Validation
 
-The verifier chooses tags -- including Concise Module ID Tags ([CoMID]({sec-comid})), Concise Software ID Tags ([CoSWID]), and/or Concise Trust Anchor Stores ([CoTS](I-D.ietf-rats-concise-ta-stores)) -- from the selected CoRIMs.
+The verifier chooses tags -- including Concise Module ID Tags (CoMID, {{sec-comid}}), Concise Software ID Tags (CoSWID, {{-coswid}}), and/or Concise Trust Anchor Stores (CoTS, {{?I-D.ietf-rats-concise-ta-stores}}) -- from the selected CoRIMs.
 
 The verifier MUST discard all tags which are not syntactically and semantically valid.
 In particular, any internal cross-reference (e.g., CoMID-CoSWID linking triples) MUST be successfully resolved.
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -1109,7 +1109,7 @@ The selection process MUST yield at least one usable tag.
 
 All the available Concise Bill Of Material (CoBOMs) tags are then collected from the selected CoRIMs.
 
-The verifier MUST check each CoBOM to see whether it can be activated. The verifier MUST activate all tags referenced by an activated CoBOM.
+The verifier MUST activate all tags referenced by a CoBOM.
 
 After the verifier has processed all CoBOMs it MUST discard any tags which have not been activated by a CoBOM.
 
@@ -1118,7 +1118,7 @@ After the verifier has processed all CoBOMs it MUST discard any tags which have 
 The verifier chooses tags -- including Concise Module ID Tags (CoMID, {{sec-comid}}), Concise Software ID Tags (CoSWID, {{-coswid}}), and/or Concise Trust Anchor Stores (CoTS, {{?I-D.ietf-rats-concise-ta-stores}}) -- from the selected CoRIMs.
 
 The verifier MUST discard all tags which are not syntactically and semantically valid.
-In particular, any internal cross-reference (e.g., CoMID-CoSWID linking triples) MUST be successfully resolved.
+In particular, any cross-referenced triples (e.g., CoMID-CoSWID linking triples) MUST be successfully resolved.
 
 ### Appraisal Context Construction
 
@@ -1144,7 +1144,7 @@ For example:
 In DICE, a proof of liveness is performed on the final key in the certificate chain. If this passes then a suitable certification path anchored on a trusted root certificate is looked up -- e.g., based on linking information obtained from the DeviceID certificate (see Section 9.2.1 of {{DICE-Layering-Architecture}})-- in the Appraisal Context.  If found, then usual X.509 certificate validation is performed.
 In PSA, the verification public key is looked up in the appraisal context using the `euid` claim found in the PSA claims-set (see {{Section 4.2.1 of -psa-token}}).  If found, COSE Sign1 verification is performed accordingly.
 
-Independent of the specific method, the cryptographic integrity of Evidence MUST be successfully verified.
+Independent of the specific integrity protection method used, the integrity of Evidence MUST be successfully verified.
 
 > A CoRIM profile MUST describe:
 >
@@ -1175,7 +1175,7 @@ In the Evidence Appraisal phase, a CoRIM Appraisal Context and an Evidence Appra
 The outcome of the appraisal process is summarised in an Attestation Result.
 The Relying Party application uses the content of the Attestation Result to make its own policy decisions.
 
-We make no assumptions on the specific shape of the Attestation Result, except for its optional ability to include Evidence from the attestor and Endorsed Values that the Verifier has been able to infer from Evidence and the Appraisal Context.
+This specification makes no assumptions on the specific shape of the Attestation Result, except for its optional ability to include Evidence from the attestor and Endorsed Values that the Verifier has been able to infer from Evidence and the Appraisal Context.
 
 > A CoRIM profile MUST describe:
 >
@@ -1223,7 +1223,7 @@ Note that while specifications may extend the matching semantics using tags, the
 
 #### Comparing svn entries
 
-#### Comparing digest entries
+#### Comparing digests entries
 
 #### Defining handling for new tags
 
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -1121,7 +1121,7 @@ This concludes the initialisation phase.
 
 ## Evidence Collection
 
-In the evidence collection phase the verifier communicates with attesters to collect evidence. 
+In the evidence collection phase the verifier communicates with attesters to collect evidence.
 
 The first part of the evidence collection phase does not perform any crypographic validation. This allows verifiers to use untrusted code for their initial evidence collection.
 
@@ -1153,7 +1153,7 @@ The accepted claims set will be matched against CoMID reference values, so this 
 
 Each entry in the measurement-values-map is a separate piece of evidence describing the environment named by the environment-map. An attestor can provide multiple environment-maps each containing a single values measurement-values map, a single environment-map containing multiple entries in its measurement-values-map, or a combination of these approaches.
 
-If evidence from different sources has the same environment-map then the measurement-values-maps are merged. If both measurement-value-maps being merged contain the same key then the values associated with that key MUST be binary identical. 
+If evidence from different sources has the same environment-map then the measurement-values-maps are merged. If both measurement-value-maps being merged contain the same key then the values associated with that key MUST be binary identical.
 
 ### Transforming SPDM Evidence to a format usable for matching
 
@@ -1176,7 +1176,7 @@ The verifier SHALL translate each field in the TcbInfo into a field in the creat
 - The TcbInfo `fwids` field SHALL be translated to the field named `measurement-map / mval / digests`
   - Each digest within fwids is translated to a CoMID digest object, with an appropriate algorithm identifier
 - The TcbInfo `flags` field SHALL be translated to the field named `measurement-map / mval / flags`
-  - Each flag is translated independently 
+  - Each flag is translated independently
 - The TcbInfo `vendorInfo` SHALL shall be copied to the field named `measurement-map / mval / raw-value`
 
 If there are multiple evidence triples with the same environment map then ...
@@ -1241,7 +1241,7 @@ If the reference valus is not tagged, and the key is not a special value describ
 If a CoMID tag matches the Accepted Claims Set then the verifier must attempt to add Endorsed values from that tag to the Accepted Claims Set.
 
 If Endorsed values being added have the same name and measurement-values-map key, but a different measurement-value-map entry from the existing value.
- 
+
 # Implementation Status
 
 This section records the status of known implementations of the protocol
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -1126,19 +1126,19 @@ The selection process MUST yield at least one usable tag.
 All the available Concise Bill Of Material (CoBOMs) tags are then collected
 from the selected CoRIMs.
 
-The verifier MUST activate all tags referenced by a CoBOM.
+The Verifier MUST activate all tags referenced by a CoBOM.
 
-After the verifier has processed all CoBOMs it MUST discard any tags which have
+After the Verifier has processed all CoBOMs it MUST discard any tags which have
 not been activated by a CoBOM.
 
 ### Tags Identification and Validation
 
-The verifier chooses tags -- including Concise Module ID Tags (CoMID, {{sec-comid}}),
+The Verifier chooses tags -- including Concise Module ID Tags (CoMID, {{sec-comid}}),
 Concise Software ID Tags (CoSWID, {{-coswid}}),
 and/or Concise Trust Anchor Stores (CoTS, {{?I-D.ietf-rats-concise-ta-stores}}) --
 from the selected CoRIMs.
 
-The verifier MUST discard all tags which are not syntactically and semantically
+The Verifier MUST discard all tags which are not syntactically and semantically
 valid.
 In particular, any cross-referenced triples (e.g., CoMID-CoSWID linking triples)
 MUST be successfully resolved.
@@ -1153,12 +1153,12 @@ This concludes the initialisation phase.
 
 ## Evidence Collection
 
-In the evidence collection phase the verifier communicates with attesters to
+In the evidence collection phase the Verifier communicates with attesters to
 collect evidence.
 
 The first part of the Evidence collection phase does not perform any
 cryptographic validation.
-This allows verifiers to use untrusted code for their initial Evidence collection.
+This allows Verifiers to use untrusted code for their initial Evidence collection.
 
 The results of the evidence collection are protocol specific data and transcripts
 which are used as input to appraisal by the Verifier.
@@ -1196,20 +1196,32 @@ Evidence MUST be successfully verified.
 ### The Accepted Claims Set
 
 At the end of the Evidence collection process evidence has been converted into
-a format suitable for appraisal.
+a format suitable for appraisal. This document describes an `accepted-claims-set`
+format and the algorithms used to compare it against CoMID reference values.
+
+    accepted-claims-set = {​
+        &(ce.evidence-triples: 0) => [ + reference-triple-record ]​
+      ? &(ce.identity-triples: 1) => [ + identity-triple-record ]​
+      ? &(ce.dependency-triples: 2) => [ + domain-dependency-triple-record ]
+      ? &(ce.domain-membership-triples: 3) => [ + domain-membership-triple-record ]
+      ? &(ce.coswid-triples: 4) => [ + ev-coswid-triple-record ]
+      * $$ev-triples-map-extension​
+    }
+
 Verifiers are not required to use this as their internal state, but for the
-purposes of this document a sample verifier is discussed which uses this format.
+purposes of this document a sample Verifier is discussed which uses this format.
 
 The Accepted Claims Set will be matched against CoMID reference values, as per
-the appraisal policy of the verifier.
+the appraisal policy of the Verifier.
 This document describes an example evidence structure which can be easily
 matched against these reference values.
 Each set of evidence contains an `environment-map` providing a namespace, and
-a `measurement-values-map` containing one or more entries.
+a non empty `measurement-values-map`.
 
 Each entry in the `measurement-values-map` is a separate piece of evidence
 describing the environment named by the `environment-map`.
- An attestor can provide multiple `environment-map`s each containing a
+
+ An Attester can provide multiple `environment-map`s each containing a
  `measurement-values-map` with one entry;  a single `environment-map` containing
  multiple entries in its `measurement-values-map`; or a combination of
  these approaches.
@@ -1219,30 +1231,38 @@ If evidence from different sources has the same `environment-map` then the
 If both measurement-value-maps being merged contain the same key then the
 values associated with that key MUST be binary identical.
 
+### Accepted Claims Set Initialisation
+
+The Accepted Claims Set is initialised with cryptographically verified Evidence
+from the Attestation Environments.
+
 > A CoRIM profile MUST describe:
 >
 > * How evidence is converted to a format suitable for appraisal
 
-{{sec-dice-spdm}} provides information on how evidence collected using
+Section {sec-dice-spdm} provides information on how evidence collected using
 DICE and SPDM protocols is added to the Accepted Claims Map.
 
 
-## Evidence appraisal
+## Accepted Claims Map extension using CoMID tags
 
-In the Evidence Appraisal phase, a CoRIM Appraisal Context and an Evidence
-Appraisal Policy are used by the Verifier to appraise the received evidence.
-If the evidence is acceptable then the CoRIM can supplement it with a more
-concise description of the attestor.
+In the Accepted Claims Map extension phase, a CoRIM Appraisal Context and
+an Evidence Appraisal Policy are used by the Verifier to find CoMID tags which
+match the Attester. Tags which match are accepted, and the Accepted Claims Map
+is extended using Endorsements etc. from the accepted tags.
+
 For example, a CoRIM provided by a firmware author might match against the hash
 of a piece of firmware and provide the version number of that firmware.
-This phase may be repeated multiple times.
+
+The Accepted Claims Map extension phase may be repeated multiple times.
+
 The outcome of the appraisal process is summarised in an Attestation Result.
 The Relying Party application uses the content of the Attestation Result to
 make its own policy decisions.
 
 This specification makes no assumptions on the specific shape of the
 Attestation Result, except for its optional ability to include Evidence from
-the attestor and Endorsed Values that the Verifier has been able to infer from
+the Attester and Endorsed Values that the Verifier has been able to infer from
 Evidence and the Appraisal Context.
 
 > A CoRIM profile MUST describe:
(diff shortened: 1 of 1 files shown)


draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1066,182 @@ The following describes each member of the `concise-bom-tag` map.
...
+
+Note that what follows is not meant to describe a real implementation.
+In particular, it is expected that the cost associated with the initialisation phase can be amortised across multiple appraisals.
+Implementers are free to do as they please, as long as the stated invariants are not broken.
+
+In the following, if a MUST-level requirement is violated, the entire procedure is aborted.
+
+## Appraisal Context initialisation
+
+The goal of the initialisation phase is to load the CoRIM Appraisal Context with objects such as CoRIM tags, cryptographic validation key material (e.g., raw public keys, root certificates, intermediate CA certificate chains, etc.) etc. that will be used in the subsequent Evidence Appraisal phase.
```
nedmsmith, line 1088: It is helpful is long sentences are broken into separate lines that don't wrap around.
the use of ... etc. ) etc. ...  is clunky. I don't think the sentence looses any meaning if the second 'etc.' is omitted.

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1066,182 @@ The following describes each member of the `concise-bom-tag` map.
...
+### Tags Identification and Validation
+
+Potentially valid tags -- i.e., Concise Module Identifiers (CoMID) and/or Concise Software Identifiers (CoSWID) and/or Concise Trust Anchor Stores (CoTS) -- are chosen from the selected CoRIMs.
+
+All the extracted tags MUST be successfully located and be syntactically and semantically valid.
+In particular, any internal cross-reference (e.g., CoMID-CoSWID linking triples) MUST be successfully resolved.
+
+### Appraisal Context Construction
+
+All of the potential tags are loaded into the Appraisal Context. The complete contents of each tag are loaded together, including Reference Values, Endorsed Values and cryptographic verification key material.
```
nedmsmith, line 1118: Suggest "All of the validated and potentially useful tags are..."
'loaded together' is ambiguous - I don't have a suggestion per se, but wording to the effect that they are available to the Verifier for appraisal.

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1066,182 @@ The following describes each member of the `concise-bom-tag` map.
...
+
+## Evidence Collection
+
+In the evidence collection phase the verifier communicates with attesters to collect evidence. 
+
+The first part of the evidence collection phase does not perform any crypographic validation. This allows verifiers to use untrusted code for their initial evidence collection.
+
+The results of the evidence collection are protocol specific data and transcripts which can be processed by the verifier.
+
+### Cryptographic validation of Evidence
```
nedmsmith, line 1130: The RATS Architecture refers to this as 'appraisal'. We should use that term unless we're describing something else.

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1066,182 @@ The following describes each member of the `concise-bom-tag` map.
...
+
+If the authenticity of Evidence is secured by a cryptographic mechanism such as a signature, the first step in the Evidence Appraisal is to perform cryptographic validation of the Evidence.
+
+The exact cryptographic signature validation mechanics depend on the specific Evidence collection protocol.
+
+For example:
+In DICE, a proof of liveness is performed on the final key in the certificate chain. If this passes then a suitable certification path anchored on a trusted root certificate is looked up -- e.g., based on linking information obtained from the DeviceID certificate (see Section 9.2.1 of [DICE Layering Architecture](https://trustedcomputinggroup.org/wp-content/uploads/DICE-Layering-Architecture-r19_pub.pdf))-- in the Appraisal Context.  If found, then usual X.509 certificate validation is performed.
```
nedmsmith, line 1137: break up long sentences into non-wraparound lines.
'proof of liveness' is ambiguous - maybe 'proof of possession' or 'key authentication' is intended?
DICE uses the term 'alias key' or 'alias certificate' to refer to the last layer in a DICE layering. 
References should use either [1] or [RFC0000] convention.

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1066,182 @@ The following describes each member of the `concise-bom-tag` map.
...
+In PSA, the verification public key is looked up in the appraisal context using the `euid` claim found in the PSA claims-set (see [Section 4.2.1 of I-D.tschofenig-rats-psa-token](https://www.ietf.org/archive/id/draft-tschofenig-rats-psa-token-10.html#name-instance-id).  If found, COSE Sign1 verification is performed accordingly.
+
+Independent of the specific method, the cryptographic integrity of Evidence MUST be successfully verified.
```
nedmsmith, line 1140: "specific method" is ambiguous - suggest "specific integrity protection method used, the integrity of Evidence..."

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1066,182 @@ The following describes each member of the `concise-bom-tag` map.
...
+
+> A CoRIM profile MUST describe:
+>
+> * How cryptographic verification key material is conveyed (e.g., using Attestation Keys triples, or CoTS tags)
```
nedmsmith, line 1144: "conveyed" is misleading since RATS Arch uses 'conveyance' to mean how files such as manifests are communicated from one role / entity to another. This usage seems to be describing how verification key material is to be represented by Endorsements.

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1066,182 @@ The following describes each member of the `concise-bom-tag` map.
...
+> * How key material is associated with the Attesting Environment
+> * How the Attesting Environment is identified in Evidence
+
+### The Accepted Claims Set
+
+At the end of the Evidence collection process evidence has been converted into a format suitable for appraisal. Verifiers are not required to use this as their internal state, but for the purposes of this document a sample verifier is discussed which uses this format.
```
nedmsmith, line 1150: 'evidence' to 'Evidence'

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1066,182 @@ The following describes each member of the `concise-bom-tag` map.
...
+
+The accepted claims set will be matched against CoMID reference values, so this document describes an evidence structure which can be easily matched against these reference values. Each set of evidence contains an environment map providing a namespace, and a measurement-values-map containing one or more entries.
```
nedmsmith, line 1152: 'reference values' to 'Reference Values'
'evidence' to 'Evidence'
'environment map' to '`environment-map`'
'measurement-values-map' to '`measurement-values-map`'
More explanation is needed that describes how Evidence, in other encodings, is translated into a representation that follows a CoMID triple. Possibly elevate the section below or cross-reference it.

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1066,182 @@ The following describes each member of the `concise-bom-tag` map.
...
+
+Each entry in the measurement-values-map is a separate piece of evidence describing the environment named by the environment-map. An attestor can provide multiple environment-maps each containing a single values measurement-values map, a single environment-map containing multiple entries in its measurement-values-map, or a combination of these approaches.
+
+If evidence from different sources has the same environment-map then the measurement-values-maps are merged. If both measurement-value-maps being merged contain the same key then the values associated with that key MUST be binary identical.
```
nedmsmith, line 1156: suggest "... the measurement-values-maps are..." to "... the measurements (e.g., `measurement-values-map`) are merged.
'key' is ambiguous - suggest 'code point' 
suggest adding a simple example - "(e.g., if the 'name' code point has the value "my environment" and 'name' is found in multiple sets of measurements being merged, then each value must be "my environment" or the merge will fail."

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1066,182 @@ The following describes each member of the `concise-bom-tag` map.
...
+
+Each entry in the measurement-values-map is a separate piece of evidence describing the environment named by the environment-map. An attestor can provide multiple environment-maps each containing a single values measurement-values map, a single environment-map containing multiple entries in its measurement-values-map, or a combination of these approaches.
+
+If evidence from different sources has the same environment-map then the measurement-values-maps are merged. If both measurement-value-maps being merged contain the same key then the values associated with that key MUST be binary identical. 
+
+### Transforming SPDM Evidence to a format usable for matching
+
+***Evidence Binding For SPDM*** describes the process by which evidence in a SPDM MEASUREMENTS response is converted to Evidence suitable for matching using the rules below. The converted evidence is held in evidence triples which have a similar format to reference-triples (their semantics is described below).
```
nedmsmith, line 1160: Should use a placeholder reference e.g., [TCG_SPDM-TBD]

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1066,182 @@ The following describes each member of the `concise-bom-tag` map.
...
+
+### Transforming DICE Evidence to a format usable for matching
+
+DICE Evidence appears in certificates in the TcbInfo or MultiTcbInfo extension. Each TcbInfo, and each entry in the MultiTcbInfo, is converted to an evidence triple using the rules in this section. In a MultiTcbInfo each entry in the sequence is treated as independent and translated into a separate evidence object.
```
nedmsmith, line 1164: Note: There are additional formats. Should cite TCG specs and use DiceTcbInfo and DiceMultiTcbInfo as examples.

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1066,182 @@ The following describes each member of the `concise-bom-tag` map.
...
+
+- The TcbInfo `version` field SHALL be translated to the field named `measurement-map / mval / version / version`
+- The TcbInfo `svn` field SHALL be copied to the field named `measurement-map / mval / svn`
+- The TcbInfo `fwids` field SHALL be translated to the field named `measurement-map / mval / digests`
+  - Each digest within fwids is translated to a CoMID digest object, with an appropriate algorithm identifier
+- The TcbInfo `flags` field SHALL be translated to the field named `measurement-map / mval / flags`
+  - Each flag is translated independently 
+- The TcbInfo `vendorInfo` SHALL shall be copied to the field named `measurement-map / mval / raw-value`
+
+If there are multiple evidence triples with the same environment map then ...
```
nedmsmith, line 1182: I assume ... means TBD?
suggest: "... the measurements are merged under the common environment."

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1066,182 @@ The following describes each member of the `concise-bom-tag` map.
...
+
+## Evidence appraisal
+
+In the Evidence Appraisal phase, a CoRIM Appraisal Context and an Evidence Appraisal Policy are used to convert the received Evidence from its raw form into a more usable form. This phase may be repeated multiple times .
+The outcome of the appraisal process is summarised in an Attestation Result.
+The Relying Party application uses the content of the Attestation Result to make its own policy decisions.
+
+We make no assumptions on the specific shape of the Attestation Result, except for its optional ability to include Evidence from the attestor and Endorsed Values that the Verifier has been able to infer from Evidence and the Appraisal Context.
```
nedmsmith, line 1190: 'We make no...' to 'This specification makes no...'

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1066,182 @@ The following describes each member of the `concise-bom-tag` map.
...
+
+> A CoRIM profile MUST describe:
```
nedmsmith, line 1192: These few lines seem to be repeated from above.

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1066,182 @@ The following describes each member of the `concise-bom-tag` map.
...
+>
+> * How Endorsed Values are conveyed (e.g., using Endorsed Value triples, or Device Identity triples)
+> * How Endorsed Values are associated with the Target and Attesting Environments
+> * How the Target and Attesting Environments are identified in Evidence
+
+The sections below describe the algorithm used to compare CoMID tags.
+
+### Comparing and processing CoMID tags
+
+There can be multiple passes of the CoMID evidence appraisal process.
```
nedmsmith, line 1209: Lots of inconsistent use of case for RATS Architecture defined terms that are upper case terms.
CoMID defines Reference Values and Endorsements but not Evidence.  Maybe remove 'CoMID'?

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1066,182 @@ The following describes each member of the `concise-bom-tag` map.
...
+
+In each pass, each potential CoMID tag in the Appraisal Context is compared against the accepted claims set (which has been initialised with Evidence from the attestor(s) ). If the reference values in the CoMID tag match the Accepted Claims Set then Endorsements and other values from the tag are added to the Accepted Claims Set.
```
nedmsmith, line 1211: Note: this process assumes the Evidence, Reference Values and Endorsed Values have a common environment.

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1066,182 @@ The following describes each member of the `concise-bom-tag` map.
...
+
+If any tags matched in a pass through the CoMID appraisal process then another pass is performed. Tags which matched in a previous pass are ignored in later passes of the same appraisal.
+
+A later pass may match Reference Values from the CoMID against Endorsements from matched tags in an earlier pass, this allows CoRIM authors to create hierarchical descriptions of a system.
```
nedmsmith, line 1215: This needs more discussion.

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1066,182 @@ The following describes each member of the `concise-bom-tag` map.
...
+
+### Matching Evidence against Reference Values
+
+This section describes the process performed by the verifier to determine whether a candidate CoMID tag matches the Accepted Claims Set. If a tag does not match the accepted claims set then it is silently ignored for this pass, processing continues with the remaining candidate tags.
+
+The verifier enumerates over the reference values in the CoMID tag. If the Accepted Claims Set does not contain an entry with the same environment-map as the reference claim then the tag does not match. Comparison of environment-map is performed using a binary comparison. Verifiers SHOULD convert environment-map to the canonical format before performing the binary comparison.
```
nedmsmith, line 1221: 'enumerates' to 'iterates'

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1066,182 @@ The following describes each member of the `concise-bom-tag` map.
...
+
+The verifer locates the corresponding measurement-values-map in the reference value and the Accepted Claims Set. The verifier enumerates over the entries in the reference value measurement-values-map. If the Accepted Claims set measurement-values-map does not have an entry with the same key as the reference value measurement-values-map, or if the entries do not match, then the tag does not match.
```
nedmsmith, line 1223: Can we describe the method using more generic terminology such as "corresponding measurements, (e.g., `measurement-values-map`)" so that the specific corim statements are exemplary.

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1066,182 @@ The following describes each member of the `concise-bom-tag` map.
...
+
+The algorithm used to compare the measurement-values-map entries together depends on whether the reference value is tagged, and on the key which identifies the entry.
+
+If the reference value is tagged then the verifier must use the protocol described by the tag to match the entries.
```
nedmsmith, line 1227: 'tagged' is ambiguous - suggest 'tagged (e.g., IANA CBOR tags registry [REF-TBD])..."
'protocol' may be ambiguous
'tag' may be ambiguous

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1066,182 @@ The following describes each member of the `concise-bom-tag` map.
...
+
+If the reference value is not tagged and the key is a special value described in the sub-sections below, then the protocol appropriate to that key is used to match the entries.
```
nedmsmith, line 1229: 'key' may be ambiguous

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1066,182 @@ The following describes each member of the `concise-bom-tag` map.
...
+
+#### Comparing svn entries
+
+#### Comparing digest entries
+
+### Adding CoMID Endorsed Values to the Accepted Claims Set
+
+If a CoMID tag matches the Accepted Claims Set then the verifier must attempt to add Endorsed values from that tag to the Accepted Claims Set.
+
+If Endorsed values being added have the same name and measurement-values-map key, but a different measurement-value-map entry from the existing value.
```
nedmsmith, line 1243: 'name' may be ambiguous

nedmsmith (commented):
Manifold comments inline.

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1066,182 @@ The following describes each member of the `concise-bom-tag` map.
...
+Other selection criteria MAY be applied.
+For example, if the Evidence format is known in advance, CoRIMs that do not match the expected profile can be readily discarded.
+
+The selection process MUST yield at least one usable CoRIM.
+
+### CoBOMs Extraction
+
+All the available Concise Bill Of Material tags (CoBOMs) are collected from the selected CoRIMs.
+
+If any CoBOM is present then other tags in the CoRIM which are not activated by the CoBOM MUST be ignored.
```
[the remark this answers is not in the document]
nedmsmith: CDDL for `corim-map` doesn't make use of CoBOM tag mandatory. But the prose makes it mandatory - see [1]. Given [1], the language in 1107 can describe 'unactivated tags' as ignored. Not sure it needs to be normative MUST since [1] already contains normative. 

[1]
A Concise Bill of Material (CoBOM) object represents the signal for the verifier to activate the listed tags. Data contained in a tag MUST NOT be used for appraisal until a CoBOM which activates that tag has been received and successfully processed. All the tags listed in the CoBOM must be activated in the same transaction, i.e., either all or none.

draft-ietf-rats-corim.md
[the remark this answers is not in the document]
nedmsmith: RATS Arch uses 'Claims collection' to mean the Attesting Environment directly measures the Target Environment. It also uses 'Evidence collection' to describe a lead Attester that assembles multiple Evidences and prepares them for conveyance. 'Evidence' means the Claims are integrity protected, so we don't need to describe the various forms of Claims integrity protection here.

nedmsmith:
Note: the CI tools will fail if a line ends in whitespace.

draft-ietf-rats-corim.md
yogeshbdeshpande, line 1107:
```suggestion
If any CoBOM is present then other tags in the CoRIM which are not activated by the CoBOM MUST be ignored.
```
I think, this statement is ambiguous. In fact if any CoBOM is present, then any CoMID/CoSWID tags present in CoRIM but absent in CoBOM MUST be ignored.

draft-ietf-rats-corim.md
yogeshbdeshpande, line 1152:
```suggestion
The accepted claims set will be matched against CoMID reference values, as per the appraisal policy of the verifier. This document describes an example evidence structure which can be easily matched against these reference values. Each set of evidence contains an environment map providing a namespace, and a measurement-values-map containing one or more entries.
```

draft-ietf-rats-corim.md
yogeshbdeshpande, line 1182: Line 1182 is not complete? Can you please complete this....

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1066,182 @@ The following describes each member of the `concise-bom-tag` map.
...
+  - Each digest within fwids is translated to a CoMID digest object, with an appropriate algorithm identifier
+- The TcbInfo `flags` field SHALL be translated to the field named `measurement-map / mval / flags`
+  - Each flag is translated independently 
+- The TcbInfo `vendorInfo` SHALL shall be copied to the field named `measurement-map / mval / raw-value`
+
+If there are multiple evidence triples with the same environment map then ...
+
+## Evidence appraisal
+
+In the Evidence Appraisal phase, a CoRIM Appraisal Context and an Evidence Appraisal Policy are used to convert the received Evidence from its raw form into a more usable form. This phase may be repeated multiple times .
```
yogeshbdeshpande, line 1186:
```suggestion
In the Evidence Appraisal phase, a CoRIM Appraisal Context and an Evidence Appraisal Policy are used to convert the received Evidence from its raw form into a more usable form. This phase may be repeated multiple times.
```

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1066,182 @@ The following describes each member of the `concise-bom-tag` map.
...
+- The TcbInfo `fwids` field SHALL be translated to the field named `measurement-map / mval / digests`
+  - Each digest within fwids is translated to a CoMID digest object, with an appropriate algorithm identifier
+- The TcbInfo `flags` field SHALL be translated to the field named `measurement-map / mval / flags`
+  - Each flag is translated independently 
+- The TcbInfo `vendorInfo` SHALL shall be copied to the field named `measurement-map / mval / raw-value`
+
+If there are multiple evidence triples with the same environment map then ...
+
+## Evidence appraisal
+
```
yogeshbdeshpande, line 1185: I think the statement on Line 1186 is not accurate. 
In the Evidence Appraisal phase, a CoRIM Appraisal Context and an Evidence Appraisal Policy are used by the Verifier to appraise the received evidence to deduce the trustworthiness of the received Evidence.

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1066,182 @@ The following describes each member of the `concise-bom-tag` map.
...
+The Relying Party application uses the content of the Attestation Result to make its own policy decisions.
+
+We make no assumptions on the specific shape of the Attestation Result, except for its optional ability to include Evidence from the attestor and Endorsed Values that the Verifier has been able to infer from Evidence and the Appraisal Context.
+
+> A CoRIM profile MUST describe:
+>
+> * How Reference Values are conveyed (e.g., using Reference Value triples, or CoSWID tags)
+> * How Reference Values are associated with the Target Environment
+> * How the Target Environment is identified in Evidence
+> * How measurements from Evidence are matched against Reference Values
```
yogeshbdeshpande, line 1197: I do not think, Line 1197 is part of scope of CoRIM Profile only, but very much also depends on the Appraisal policy

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1066,182 @@ The following describes each member of the `concise-bom-tag` map.
...
+
+The sections below describe the algorithm used to compare CoMID tags.
+
+### Comparing and processing CoMID tags
+
+There can be multiple passes of the CoMID evidence appraisal process.
+
+In each pass, each potential CoMID tag in the Appraisal Context is compared against the accepted claims set (which has been initialised with Evidence from the attestor(s) ). If the reference values in the CoMID tag match the Accepted Claims Set then Endorsements and other values from the tag are added to the Accepted Claims Set.
+
+If any tags matched in a pass through the CoMID appraisal process then another pass is performed. Tags which matched in a previous pass are ignored in later passes of the same appraisal.
```
yogeshbdeshpande, line 1213: I think, this needs more discussion and clarification as well!

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1066,182 @@ The following describes each member of the `concise-bom-tag` map.
...
+
+A later pass may match Reference Values from the CoMID against Endorsements from matched tags in an earlier pass, this allows CoRIM authors to create hierarchical descriptions of a system.
+
+### Matching Evidence against Reference Values
+
+This section describes the process performed by the verifier to determine whether a candidate CoMID tag matches the Accepted Claims Set. If a tag does not match the accepted claims set then it is silently ignored for this pass, processing continues with the remaining candidate tags.
```
yogeshbdeshpande, line 1219: Line 1217 correctly say that we are matching Evidence against Reference Values, however, the description at line 1219, explains as if we are matching CoMID against Accepted Claims Set.

I suppose your intention was to match "Accepted Claims Set" against the given reference values present in the CoMID tags?

yogeshbdeshpande (commented):
Thank you for the PR.
Left few comments and clarifications

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1066,182 @@ The following describes each member of the `concise-bom-tag` map.
...
+# CoRIM-based Evidence Verification
+
+The verification procedure is divided into three separate phases:
+
+* Appraisal Context initialisation
+* Evidence collection
+* Evidence appraisal
+
+At a few well-defined points in the procedure, the Verifier behaviour will depend on the specific CoRIM profile.
+Each CoRIM profile SHALL fully describe the expected Verifier behaviour for each of those well-defined points.
```
henkbirkholz, line 1078:
```suggestion
Each CoRIM profile MUST provide a description of the expected Verifier behavior for each of those well-defined points.
```

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1066,182 @@ The following describes each member of the `concise-bom-tag` map.
...
+
+Note that what follows is not meant to describe a real implementation.
```
henkbirkholz, line 1080:
```suggestion
The following text illustrates an example.
```

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1066,182 @@ The following describes each member of the `concise-bom-tag` map.
...
+In particular, it is expected that the cost associated with the initialisation phase can be amortised across multiple appraisals.
```
henkbirkholz, line 1081: 'cost' is not introduced. What does that mean?

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1066,182 @@ The following describes each member of the `concise-bom-tag` map.
...
+Implementers are free to do as they please, as long as the stated invariants are not broken.
```
henkbirkholz, line 1082: I would recommend to refrain from 'implementers guidance' and go with 'examples'

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1066,182 @@ The following describes each member of the `concise-bom-tag` map.
...
+
+In the following, if a MUST-level requirement is violated, the entire procedure is aborted.
```
henkbirkholz, line 1084: this is always true. Could have been removed, if this would be normative text. But is an example. Doing both is possible (actually...) but confusing. Normative sections have to be separated from informative sections.

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1066,182 @@ The following describes each member of the `concise-bom-tag` map.
   The `$$concise-bom-tag-extension` extension socket is empty in this
   specification.
 
+# CoRIM-based Evidence Verification
```
henkbirkholz, line 1069: TL;DR

is this intended to be normative or an example? Please see confused comments below.

draft-ietf-rats-corim.md
```diff
@@ -71,6 +71,8 @@ normative:
     seriesinfo:
       ITU-T: Recommendation X.690
     target: https://www.itu.int/rec/T-REC-X.690
+  DICE-Layering-Architecture:
+    target: https://trustedcomputinggroup.org/wp-content/uploads/DICE-Layering-Architecture-r19_pub.pdf
```
thomas-fossati, line 75: DICE-Layering-Architecture:
    author:
      org: "Trusted Computing Group"
    title: "DICE Layering Architecture"
     target: https://trustedcomputinggroup.org/wp-content/uploads/DICE-Layering-Architecture-r19_pub.pdf
    date: July, 2020

draft-ietf-rats-corim.md
```diff
@@ -71,6 +71,12 @@ normative:
     seriesinfo:
       ITU-T: Recommendation X.690
     target: https://www.itu.int/rec/T-REC-X.690
+  DICE-Layering-Architecture:
+    author:
+      org: "Trusted Computing Group"
+    title: "DICE Layering Architecture"
+     target: https://trustedcomputinggroup.org/wp-content/uploads/DICE-Layering-Architecture-r19_pub.pdf
+    date: July, 2020
   IANA.named-information: named-info
```
thomas-fossati, line 80:
```suggestion
  DICE-Layering-Architecture:
    author:
      org: "Trusted Computing Group"
    title: "DICE Layering Architecture"
    target: https://trustedcomputinggroup.org/wp-content/uploads/DICE-Layering-Architecture-r19_pub.pdf
    date: July, 2020
  IANA.named-information: named-info
```

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1072,199 @@ The following describes each member of the `concise-bom-tag` map.
...
+The verification procedure is divided into three separate phases:
+
+* Appraisal Context initialisation
+* Evidence collection
+* Evidence appraisal
+
+At a few well-defined points in the procedure, the Verifier behaviour will depend on the specific CoRIM profile.
+Each CoRIM profile MUST provide a description of the expected Verifier behavior for each of those well-defined points.
+
+In particular, it is expected that the resources used during the initialisation phase can be amortised across multiple appraisals.
```
thomas-fossati, line 1086:
```suggestion
Note that what follows is not meant to describe a real implementation.
In particular, it is expected that the resources used during the initialisation phase can be amortised across multiple appraisals.
```

draft-ietf-rats-corim.md
thomas-fossati: > this is always true. Could have been removed, if this would be normative text. But is an example. Doing both is possible (actually...) but confusing. Normative sections have to be separated from informative sections.

We could debate on whether it's a necessity -- there could be more complex control flow decisions based on a requirement violation than just "bail out!".
But more generally, it doesn't hurt to make an explicit statement here, at the very beginning of a long and complex set of actions.

thomas-fossati: Actually, I want to correct myself.  I think we should drop L1089 altogether not because it is redundant, but because it is misleading: for example, when later on we say _"corims outside their validity period MUST be discarded"_ we don't want to kill the appraisal if a corim is found to be past its expiration, we just want to avoid pulling it in...

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1072,199 @@ The following describes each member of the `concise-bom-tag` map.
...
+Verifiers claiming compliance with this specification must exhibit the same externally visible behavior as described here, they are not required to use the same internal data structures.
+
+In the following, if a MUST-level requirement is violated, the entire procedure is aborted.
```
thomas-fossati, line 1089:
```suggestion
```

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1072,199 @@ The following describes each member of the `concise-bom-tag` map.
...
+
+The goal of the initialisation phase is to load the CoRIM Appraisal Context with objects such as CoRIM tags, cryptographic validation key material (e.g., raw public keys, root certificates, intermediate CA certificate chains), etc. that will be used in the subsequent Evidence Appraisal phase.
+
+### CoRIM Selection
+
+All available CoRIMs are collected. A Verifier may be pre-configured with a large number of tags describing many types of device. All CoRIMs are loaded at this stage, later stages will select the CoRIMs appropriate to the Evidence Appraisal step.
+
+CoRIMs that are not within their validity period, or that cannot be associated with an authenticated and authorised source MUST be discarded.
+
+CoRIM that are secured by a cryptographic mechanism such as a signature MUST be discarded if the signature chain does not validate successfully.
```
thomas-fossati, line 1101: the chaining to a known and authorised entity is taken care of in L1099.

```suggestion
CoRIM that are secured by a cryptographic mechanism such as a signature MUST be discarded if the signature does not validate successfully.
```

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1072,199 @@ The following describes each member of the `concise-bom-tag` map.
...
+
+Other selection criteria MAY be applied.
+
+For example, if the Evidence format is known in advance, tags that do not match the expected profile can be readily discarded.
+
+The selection process MUST yield at least one usable tag.
+
+### CoBOM Extraction
+
+CoBOMs are processed after all tags have been loaded from CoRIMs. All the available Concise Bill Of Material tags (CoBOMs) are collected from the selected CoRIMs.
```
thomas-fossati, line 1111: suggest streamlining a bit:

```suggestion
All the available Concise Bill Of Material tags (CoBOMs) are then collected from the selected CoRIMs.
```

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1072,199 @@ The following describes each member of the `concise-bom-tag` map.
...
+
+The verifier MUST check each CoBOM to see whether it can be activated. The verifier MUST activate all tags referenced by an activated CoBOM.
```
thomas-fossati, line 1113: not clear what the "activation criteria" for a CoBOM would be.

The idea is that, at this stage, all CoBOMs found are taken into consideration for subsequent tags identification and validation.

Probably this bit does not deserve a section on its own and can be merged with the following sub-section.

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1072,199 @@ The following describes each member of the `concise-bom-tag` map.
...
+
+After the verifier has processed all CoBOMs it MUST discard any tags which have not been activated by a CoBOM.
+
+### Tags Identification and Validation
+
+The verifier chooses tags -- including Concise Module ID Tags ([CoMID]({sec-comid})), Concise Software ID Tags ([CoSWID]), and/or Concise Trust Anchor Stores ([CoTS](I-D.ietf-rats-concise-ta-stores)) -- from the selected CoRIMs.
```
thomas-fossati, line 1118: I see you are still a young Padawan in the art of [kramdown-rfc references](https://github.com/cabo/kramdown-rfc#references) :-)

```suggestion
The verifier chooses tags -- including Concise Module ID Tags (CoMID, {{sec-comid}}), Concise Software ID Tags (CoSWID, {{-coswid}}), and/or Concise Trust Anchor Stores (CoTS, {{I-D.ietf-rats-concise-ta-stores}}) -- from the selected CoRIMs.
```

draft-ietf-rats-corim.md
nedmsmith: We can just strike "activated" from activated CoBOM without loss of meaning. The Verifier trusts CoBOM tags based on its trust anchor policy.

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1072,198 @@ The following describes each member of the `concise-bom-tag` map.
...
+
+The selection process MUST yield at least one usable tag.
+
+### CoBOM Extraction
+
+All the available Concise Bill Of Material (CoBOMs) tags are then collected from the selected CoRIMs.
+
+The verifier MUST check each CoBOM to see whether it can be activated. The verifier MUST activate all tags referenced by an activated CoBOM.
+
+After the verifier has processed all CoBOMs it MUST discard any tags which have not been activated by a CoBOM.
```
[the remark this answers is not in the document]
yogeshbdeshpande: @andrew-draper Just checking, this comment is marked as "Resolved" by you without:
* either changing the content in the PR 
OR
* Providing a suitable explaination as to why you think, you wanted to retain the same text.
Could you please clarify. 

Apologies if I am looking at the wrong set of changes? Please point to the right set of files then?

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1072,202 @@ The following describes each member of the `concise-bom-tag` map.
...
+> * How key material is associated with the Attesting Environment
+> * How the Attesting Environment is identified in Evidence
+
+### The Accepted Claims Set
+
+At the end of the Evidence collection process evidence has been converted into a format suitable for appraisal. Verifiers are not required to use this as their internal state, but for the purposes of this document a sample verifier is discussed which uses this format.
+
+The Accepted Claims Set will be matched against CoMID reference values, as per the appraisal policy of the verifier. This document describes an example evidence structure which can be easily matched against these reference values. Each set of evidence contains an `environment-map` providing a namespace, and a `measurement-values-map` containing one or more entries.
+
+Each entry in the `measurement-values-map` is a separate piece of evidence describing the environment named by the `environment-map`. An attestor can provide multiple `environment-map`s each containing a single values measurement-values map, a single `environment-map` containing multiple entries in its `measurement-values-map`, or a combination of these approaches.
```
yogeshbdeshpande, line 1162:
```suggestion
Each entry in the `measurement-values-map` is a separate piece of evidence describing the environment named by the `environment-map`. An attester can provide multiple `environment-map`s each containing a single values measurement-values map, a single `environment-map` containing multiple entries in its `measurement-values-map`, or a combination of these approaches.
```

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1072,202 @@ The following describes each member of the `concise-bom-tag` map.
...
+>
+> * How evidence is converted to a format suitable for appraisal
+
+Section {sec-dice-spdm} provides information on how evidence collected using DICE and SPDM protocols is added to the Accepted Claims Map.
+
+
+## Evidence appraisal
+
+In the Evidence Appraisal phase, a CoRIM Appraisal Context and an Evidence Appraisal Policy are used by the Verifier to appraise the received evidence.
+If the evidence is acceptable then the CoRIM can supplement it with a more concise description of the attestor.
```
yogeshbdeshpande, line 1176:
```suggestion
If the evidence is acceptable then the CoRIM can supplement it with a more concise description of the attester.
```

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1072,202 @@ The following describes each member of the `concise-bom-tag` map.
...
+For example, a CoRIM provided by a firmware author might match against the hash of a piece of firmware and provide the version number of that firmware.
+This phase may be repeated multiple times.
+The outcome of the appraisal process is summarised in an Attestation Result.
+The Relying Party application uses the content of the Attestation Result to make its own policy decisions.
+
+This specification makes no assumptions on the specific shape of the Attestation Result, except for its optional ability to include Evidence from the attestor and Endorsed Values that the Verifier has been able to infer from Evidence and the Appraisal Context.
```
yogeshbdeshpande, line 1182:
```suggestion
This specification makes no assumptions on the specific shape of the Attestation Result, except for its optional ability to include Evidence from the attester and Endorsed Values that the Verifier has been able to infer from Evidence and the Appraisal Context.
```

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1072,202 @@ The following describes each member of the `concise-bom-tag` map.
...
+> * How Endorsed Values are conveyed (e.g., using Endorsed Value triples, or Device Identity triples)
+> * How Endorsed Values are added to the Accepted Claims Map
+
+The sections below describe the algorithm used to compare CoMID tags.
+
+### Comparing and processing CoMID tags
+
+There can be multiple passes of the CoMID evidence appraisal process.
+
+In each pass, each potential CoMID tag in the Appraisal Context is compared against the accepted claims set (which has been initialised with Evidence from the attestor(s) ). If the reference values in the CoMID tag match the Accepted Claims Set then Endorsements and other values from the tag are added to the Accepted Claims Set.
```
yogeshbdeshpande, line 1202:
```suggestion
In each pass, each potential CoMID tag in the Appraisal Context is compared against the accepted claims set (which has been initialised with Evidence from the attester(s) ). If the reference values in the CoMID tag match the Accepted Claims Set then Endorsements and other values from the tag are added to the Accepted Claims Set.
```

draft-ietf-rats-corim.md
thomas-fossati: I don't think this comments have been addressed:

1. remove "activate"
2. merge the two adjacent subsections

draft-ietf-rats-corim.md
```diff
@@ -1080,84 +1080,112 @@ The verification procedure is divided into three separate phases:
...
-At a few well-defined points in the procedure, the Verifier behaviour will depend on the specific CoRIM profile.
-Each CoRIM profile MUST provide a description of the expected Verifier behavior for each of those well-defined points.
+At a few well-defined points in the procedure, the Verifier behaviour will
+depend on the specific CoRIM profile.
+Each CoRIM profile MUST provide a description of the expected Verifier behavior
+for each of those well-defined points.
 
 Note that what follows describes a simplified implementation.
-Verifiers claiming compliance with this specification must exhibit the same externally visible behavior as described here,
+Verifiers claiming compliance with this specification must exhibit the same
```
thomas-fossati, line 1089: this must is normative:

```suggestion
Verifiers claiming compliance with this specification MUST exhibit the same
```

draft-ietf-rats-corim.md
```diff
@@ -1080,84 +1080,112 @@ The verification procedure is divided into three separate phases:
...
 ## Appraisal Context initialisation
 
-The goal of the initialisation phase is to load the CoRIM Appraisal Context with objects such as CoRIM tags,
-cryptographic validation key material (e.g., raw public keys, root certificates, intermediate CA certificate chains), etc.
-that will be used in the subsequent Evidence Appraisal phase.
+The goal of the initialisation phase is to load the CoRIM Appraisal Context
+with objects such as CoRIM tags,
+cryptographic validation key material (e.g., raw public keys, root certificates,
+intermediate CA certificate chains), etc. that will be used in the subsequent
+Evidence Appraisal phase.
```
thomas-fossati, line 1101: An observation: the term "CoRIM tag" does not look right.  A CoRIM _has_ tags (i.e., CoMIDs, CoSWIDs, CoBOMs, CoTS), but it's not itself a tag.

In this context seems more appropriate to talk about endorsed values and reference values rather than a generic CoRIM, which includes the crypto material.

```suggestion
with objects such as CoRIM tags,
cryptographic validation key material (e.g., raw public keys, root certificates,
intermediate CA certificate chains), etc. that will be used in the subsequent
Evidence Appraisal phase.
```

draft-ietf-rats-corim.md
```diff
@@ -1080,84 +1080,112 @@ The verification procedure is divided into three separate phases:
...
+All CoRIMs are loaded at this stage, later stages will select the CoRIMs
+appropriate to the Evidence Appraisal step.
 
-CoRIMs that are not within their validity period, or that cannot be associated with an authenticated and authorised source MUST be discarded.
+CoRIMs that are not within their validity period, or that cannot be associated
+with an authenticated and authorised source MUST be discarded.
 
-CoRIM that are secured by a cryptographic mechanism such as a signature which does not pass validation MUST be discarded.
+CoRIM that are secured by a cryptographic mechanism such as a signature which
+does not pass validation MUST be discarded.
```
thomas-fossati, line 1115: this seems to repeat L1112, specifically:

_"[...] that cannot be associated with an authenticated [...] source MUST be discarded."_

draft-ietf-rats-corim.md
```diff
@@ -1080,84 +1080,112 @@ The verification procedure is divided into three separate phases:
...
-The verifier MUST discard all tags which are not syntactically and semantically valid.
-In particular, any cross-referenced triples (e.g., CoMID-CoSWID linking triples) MUST be successfully resolved.
+The verifier MUST discard all tags which are not syntactically and semantically
+valid.
+In particular, any cross-referenced triples (e.g., CoMID-CoSWID linking triples)
+MUST be successfully resolved.
 
 ### Appraisal Context Construction
 
 All of the validated and potentially useful tags are loaded into the Appraisal Context.
```
thomas-fossati, line 1148: I suggest we drop _"potentially useful"_ as it's not necessary (and a bit distracting)

draft-ietf-rats-corim.md
```diff
@@ -1080,84 +1080,112 @@ The verification procedure is divided into three separate phases:
...
+collect evidence.
 
-The first part of the Evidence collection phase does not perform any cryptographic validation.
+The first part of the Evidence collection phase does not perform any
+cryptographic validation.
 This allows verifiers to use untrusted code for their initial Evidence collection.
 
-The results of the evidence collection are protocol specific data and transcripts which can be used as input to appraisal by a trusted Verifier.
+The results of the evidence collection are protocol specific data and transcripts
+which can be used as input to appraisal by a trusted Verifier.
```
thomas-fossati, line 1164: streamline

```suggestion
which are used as input to appraisal by the Verifier.
```

draft-ietf-rats-corim.md
```diff
@@ -1080,84 +1080,112 @@ The verification procedure is divided into three separate phases:
...
-Each tag is loaded into the Appraisal Context as a single unit, later stages will accept or ignore the whole tag.
+Each tag is loaded into the Appraisal Context as a single unit, later stages
+will accept or ignore the whole tag.
 
 This concludes the initialisation phase.
 
 ## Evidence Collection
 
-In the evidence collection phase the verifier communicates with attesters to collect evidence.
+In the evidence collection phase the verifier communicates with attesters to
```
thomas-fossati, line 1156: missing capitalisation (Evidence, Verifier, Attester)

draft-ietf-rats-corim.md
```diff
@@ -1080,84 +1080,112 @@ The verification procedure is divided into three separate phases:
...
-In DICE, a proof of liveness is performed on the final key in the certificate chain.
-If this passes then a suitable certification path anchored on a trusted root certificate is looked up
--- e.g., based on linking information obtained from the DeviceID certificate (see Section 9.2.1 of {{DICE-Layering-Architecture}})--
-in the Appraisal Context.  If found, then usual X.509 certificate validation is performed.
-In PSA, the verification public key is looked up in the appraisal context using the `euid` claim found in the PSA claims-set (see {{Section 4.2.1 of -psa-token}}).
+In DICE, a proof of liveness is performed on the final key in the certificate
+chain.
+If this passes then a suitable certification path anchored on a trusted root
+certificate is looked up -- e.g., based on linking information obtained from
+the DeviceID certificate (see Section 9.2.1 of {{DICE-Layering-Architecture}})--
```
thomas-fossati, line 1180:
```suggestion
the DeviceID certificate (see Section 9.2.1 of {{DICE-Layering-Architecture}}) --
```

thomas-fossati (commented):
committing a first batch of comments before leaving office for the w/e.  I'll restart on Monday from L1196.

draft-ietf-rats-corim.md
nedmsmith: +1

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1072,328 @@ The following describes each member of the `concise-bom-tag` map.
...
+## Evidence Collection
+
+In the evidence collection phase the verifier communicates with attesters to
+collect evidence.
+
+The first part of the Evidence collection phase does not perform any
+cryptographic validation.
+This allows verifiers to use untrusted code for their initial Evidence collection.
+
+The results of the evidence collection are protocol specific data and transcripts
```
nedmsmith, line 1163: Inconsistent use of terms: 'Evidence collection' or 'evidence collection' but not both.

draft-ietf-rats-corim.md
nedmsmith: 'can be' seems appropriate since Evidence can be protected by a conveyance protocol or by a conveyance mechanism such as COSE wrapper.

nedmsmith (approved):

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1072,328 @@ The following describes each member of the `concise-bom-tag` map.
...
+Endorsements and other values from the tag are added to the Accepted Claims Set.
+
+If any tags matched in a pass through the CoMID appraisal process then another
+pass is performed.
+Tags which matched in an earlier pass are ignored in later passes of the same
+appraisal.
+
+A later pass may match Reference Values from the CoMID against Endorsements
+added by matched tags in an earlier pass, this allows CoRIM authors to create
+hierarchical descriptions of a system.
```
thomas-fossati, line 1279: Complex behaviour as the one described above should be encoded in a triple rather than trickling through the whole process.

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1072,328 @@ The following describes each member of the `concise-bom-tag` map.
...
+
+### Matching Evidence against Reference Values
+
+This section describes the process performed by the verifier to determine
+whether a candidate CoMID tag matches the Accepted Claims Set.
+If a tag does not match the Accepted Claims Set then it is silently ignored
+for this pass, the Verifier continues to iterate over the remaining candidate
+tags.
```
thomas-fossati, line 1287: It doesn't seem to be specified what it mean for a tag to match?  All the triples must match?  Or one is sufficient?  If the latter, then I'd argue that the unit of matching is not the CoMID, but the triple.

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1072,328 @@ The following describes each member of the `concise-bom-tag` map.
...
+
+The Verifier iterates over the Reference Values in the CoMID tag.
+If the Accepted Claims Set does not contain an entry with the same `environment-map`
+as the Reference Value then the tag does not match.
+Comparison of `environment-map` is performed using a binary comparison.
+A Verifier SHOULD convert `environment-map` to the canonical format before
```
thomas-fossati, line 1293: We don't seem to define what is the _"canonical format"_?

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1072,328 @@ The following describes each member of the `concise-bom-tag` map.
...
+If any tags matched in a pass through the CoMID appraisal process then another
+pass is performed.
+Tags which matched in an earlier pass are ignored in later passes of the same
+appraisal.
+
+A later pass may match Reference Values from the CoMID against Endorsements
+added by matched tags in an earlier pass, this allows CoRIM authors to create
+hierarchical descriptions of a system.
+
+### Matching Evidence against Reference Values
```
thomas-fossati, line 1281: It looks like at least half of this section (i.e., the part that discusses the details of value matching) belongs to the measurement-value-map section.

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1072,328 @@ The following describes each member of the `concise-bom-tag` map.
...
+
+[^issue]: Content missing. Tracked at https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/71
+
+#### Defining handling for new tags
+
+[^issue]: Content missing. Tracked at https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/71
+
+### Adding CoMID Endorsed Values to the Accepted Claims Set
+
+If a CoMID tag matches the Accepted Claims Set then the verifier must attempt
```
thomas-fossati, line 1345: Not sure what we meant by _"must attempt"_.  It this a best-effort operation?  (I don't think so.)

thomas-fossati (commented):
review completed.  Thanks for your patience.

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1072,328 @@ The following describes each member of the `concise-bom-tag` map.
...
+> * How Endorsed Values are conveyed (e.g., using Endorsed Value triples, or Device Identity triples)
+> * How Endorsed Values are added to the Accepted Claims Map
+
+The sections below describe the algorithm used to compare CoMID tags.
+
+### Comparing and processing CoMID tags
+
+There can be multiple passes of the CoMID evidence appraisal process.
+
+In each pass, each potential CoMID tag in the Appraisal Context is compared
```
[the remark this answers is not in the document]
nedmsmith: Does "activated" mean the same as "accepted" in the context of the discussion related to "accepted Claims"?

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1072,328 @@ The following describes each member of the `concise-bom-tag` map.
...
+appraisal.
+
+A later pass may match Reference Values from the CoMID against Endorsements
+added by matched tags in an earlier pass, this allows CoRIM authors to create
+hierarchical descriptions of a system.
+
+### Matching Evidence against Reference Values
+
+This section describes the process performed by the verifier to determine
+whether a candidate CoMID tag matches the Accepted Claims Set.
```
nedmsmith, line 1284: Given the discussion above, the candidate "Claim" is "accepted" if the rules for appraisal are satisfied, then the Claim is added to the Accepted Claims set. 
A CoMID "tag" can contain multiple triples that describe multiple components having multiple Claims. It seems trying to describe an appraisal process at the 'tag' granularity may be difficult.

draft-ietf-rats-corim.md
nedmsmith: The Verifier relies on Reference Values to specify the 'canonical' or expected format of key-value expressions. If Evidence is in a different (non canonical) form, then the Verifier can do a conversion. For example, endieness, words vs. bytes, ascii vs. UTF8 etc.

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1072,328 @@ The following describes each member of the `concise-bom-tag` map.
...
+with the same key as the Reference Value `measurement-values-map`,
+or if the entries do not match, then the tag does not match.
+
+The algorithm used to match the `measurement-values-map` entries together
+depends on whether the reference value is tagged,
+and on the `measurement-values-map` key which identifies the entry.
+
+If the Reference Value `measurement-values-map` value starts with a CBOR tag
+then the verifier MUST use the algorithm associated with that tag to match
+the entries.
```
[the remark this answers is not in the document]
nedmsmith: CBOR tags could have multiple meanings: (a) data format, (b) function that is applied to the data.  I don't know if we have defined cddl that implies (b) but if we do, we should discuss the implications of this. In the case of `min-svn` there is an expectation that a 'min' function is applied when evaluating an `svn` with a `min-svn`.  However, a tagged-min-svn only indirectly implies the 'min' function semantics as the data type itself implies function semantics. 
It is reasonable to expect that function semantics could be explicitly described by the cddl expression contained in the 'value' side of a 'key-value' expression. The `min-svn` example encodes the function semantics in the 'key' rather than the 'value'.

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1072,328 @@ The following describes each member of the `concise-bom-tag` map.
...
+
+[^issue]: Content missing. Tracked at https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/71
+
+#### Defining handling for new tags
+
+[^issue]: Content missing. Tracked at https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/71
+
+### Adding CoMID Endorsed Values to the Accepted Claims Set
+
+If a CoMID tag matches the Accepted Claims Set then the verifier must attempt
```
nedmsmith, line 1345:
```suggestion
If a CoMID tag matches the Accepted Claims Set then the Verifier must attempt
```

draft-ietf-rats-corim.md
nedmsmith: (as suggested above) matching should be described as the Claim level of granularity. For a CoMID tag to "match" Endorsed values, the tag must contain triples that contain Endorsed Values. The Environment that is the subject of the triple (aka the name portion of the 'name-value' statement) must equal the name portion of a 'name-value' statement found in ACS. This is closer to what it means to "match" something. The measurements in the 'value' portion of the Endorsed Values claim is copied into the ACS at the place where the name match occurred. 
Note: If there are Reference Value claims that didn't match Evidence then the node in the ACS where the Endorsement claim would have been copied into will be removed including any claims that might have shown up in advance, be it from processing Endorsements or from processing Attester Evidence claims are are the superset of the Reference Values claims.  
By describing the acceptance and rejection criteria, the explanation can avoid having to be careful about building up state of the ACS over multiple passes and keeping track to apply the right pass at the right time.

draft-ietf-rats-corim.md
```diff
@@ -1066,6 +1072,328 @@ The following describes each member of the `concise-bom-tag` map.
...
+to add Endorsed values from that tag to the Accepted Claims Set.
+
+If any of the Endorsed values being added have the same `environment-map` and
+`measurement-values-map` key, but a different`measurement-values-map` value
+from the value in the Accepted Claims Map then the verifier MUST NOT add any
+part of the tag to the Accepted Claims Map.
```
nedmsmith, line 1351:
```suggestion
part of the Claim to the Accepted Claims Map.
```

nedmsmith: This paragraph assumes 'exact match' semantics. Maybe this assumption should be stated clearly as part of the intro to this subsection?

nedmsmith (commented):
Multiple comments

draft-ietf-rats-corim.md
```diff
@@ -1095,7 +1095,7 @@ phase can be amortised across multiple appraisals.
 ## Appraisal Context initialisation
 
 The goal of the initialisation phase is to load the CoRIM Appraisal Context
-with objects such as CoRIM tags,
+with objects such as tags (CoMID, CoSWID etc.) from CoRIM files,
```
thomas-fossati, line 1098:
```suggestion
with objects such as tags (CoMID, CoSWID, etc.) from CoRIM files,
```

PR #75, opened 2023-03-31
remove spurious link definition

thomas-fossati:
xml2rfc laments about an absurd link that is created because square brackets are not escaped in an entry of the CBOR tag registration table.


nedmsmith (approved):

yogeshbdeshpande (approved):
LGTM!

PR #76, opened 2023-03-31
talk a bit more about profiles

thomas-fossati:


draft-ietf-rats-corim.md
```diff
@@ -387,8 +387,18 @@ The following describes each child element of this type.
 
 ### Profile Types {#sec-corim-profile-types}
 
-A profile specifies which of the optional parts of a CoRIM are required, which
-are prohibited and which extension points are exercised and how.
+Profiling is the mechanism that allows the base CoRIM schema to be customised to fit a specific Attester or Attesters' family.
```
nedmsmith, line 390: not sure 'Attester's family' is a concrete concept. It may generate more questions.

draft-ietf-rats-corim.md
```diff
@@ -387,8 +387,18 @@ The following describes each child element of this type.
...
+A profile defines which of the optional parts of a CoRIM are required, which are prohibited and which extension points are exercised and how.
+A profile MUST NOT alter the syntax or semantics of an already defined CoRIM type.
+A profile MAY constrain the values of a given CoRIM type to a subset of that type.
+
+CoRIM profiles SHOULD be specified in a publicly available document.
+
+A CoRIM profile can use one of the base CoRIM media types defined in {{sec-mt-corim-signed}} and {{sec-mt-corim-unsigned}} with the `profile` parameter set to the appropriate value.
+Alternatively, it MAY define and register its own media type.
+
+A profile identifier is either a OID {{-cbor-oids}} or a URL {{-uri}}.
```
nedmsmith, line 401:
```suggestion
A profile identifier is either an OID {{-cbor-oids}} or a URL {{-uri}}.
```

draft-ietf-rats-corim.md
```diff
@@ -387,8 +387,18 @@ The following describes each child element of this type.
...
-are prohibited and which extension points are exercised and how.
+Profiling is the mechanism that allows the base CoRIM schema to be customised to fit a specific Attester or Attesters' family.
+
+A profile defines which of the optional parts of a CoRIM are required, which are prohibited and which extension points are exercised and how.
+A profile MUST NOT alter the syntax or semantics of an already defined CoRIM type.
+A profile MAY constrain the values of a given CoRIM type to a subset of that type.
+
+CoRIM profiles SHOULD be specified in a publicly available document.
+
+A CoRIM profile can use one of the base CoRIM media types defined in {{sec-mt-corim-signed}} and {{sec-mt-corim-unsigned}} with the `profile` parameter set to the appropriate value.
```
nedmsmith, line 398:
```suggestion
A CoRIM profile can use one of the base CoRIM media types defined in {{sec-mt-corim-signed}} and
{{sec-mt-corim-unsigned}} with the `profile` parameter set to the appropriate value.
```

draft-ietf-rats-corim.md
```diff
@@ -387,8 +387,18 @@ The following describes each child element of this type.
 
 ### Profile Types {#sec-corim-profile-types}
 
-A profile specifies which of the optional parts of a CoRIM are required, which
-are prohibited and which extension points are exercised and how.
+Profiling is the mechanism that allows the base CoRIM schema to be customised to fit a specific Attester or Attesters' family.
+
+A profile defines which of the optional parts of a CoRIM are required, which are prohibited and which extension points are exercised and how.
```
nedmsmith, line 392:
```suggestion
A profile defines which of the optional parts of a CoRIM are required,
which are prohibited and which extension points are exercised and how.
Exercised extension points should preserve the intent of the original map semantics.
```

draft-ietf-rats-corim.md
```diff
@@ -387,8 +387,18 @@ The following describes each child element of this type.
...
+A profile MUST NOT alter the syntax or semantics of an already defined CoRIM type.
+A profile MAY constrain the values of a given CoRIM type to a subset of that type.
```
yogeshbdeshpande, line 394:
```suggestion
A profile MAY constrain the set of a given CoRIM type to a subset of that type.
A profile MAY extend the set of a given CoRIM type to meet the requirement of a specific profile.
```

nedmsmith (commented):
suggested changes inline.

draft-ietf-rats-corim.md
yogeshbdeshpande, line 390:
```suggestion
Profiling is the mechanism that allows the base CoRIM schema to be customised to fit a specific Attester or Attesters' class.
```

draft-ietf-rats-corim.md
nedmsmith: I think 'values' is more intuitive than 'set' as set implies the type structure follows mathematical sets. But 'subset' isn't inferring mathematical sets.

yogeshbdeshpande: I felt it a bit like key value semantics hence suggested, with the set.
Perhaps `choice` may fit better, isn't it?

nedmsmith: maybe less is more?

thomas-fossati: "set" alone doesn't sound good.  If you were instead suggesting "set of values", I concur with Ned that "values" alone does the job.

draft-ietf-rats-corim.md
thomas-fossati: I'm OK removing "Attesters' family" altogether.
In alternative, use Ned's suggestion in https://github.com/ietf-rats-wg/draft-ietf-rats-corim/pull/76#discussion_r1154636121

thomas-fossati: I'm going with option 1

thomas-fossati: dropped in [cee17f1](https://github.com/ietf-rats-wg/draft-ietf-rats-corim/pull/76/commits/cee17f19773f8503d87dc204bc00a0f924acfe75)

draft-ietf-rats-corim.md
yogeshbdeshpande: ok, fine. However what about  the other aspect of extend also ? It seems to say that we only Constrain the types.

yogeshbdeshpande: I suppose the native types which are `$type-choices` can be extended, isn't it ?

thomas-fossati: Not necessarily, only when we say so.

nedmsmith: > Not necessarily, only when we say so.

In the context of defining profile rules, the profile mechanism is a way to extend / restrict the standard (base) schema in a non-standard way. By requiring the profile to be public, there is an approximation of a 'defacto' standardization. This flexibility is desirable even necessary as a way to accommodate the wide range of product differences. However, profiles are better than fully proprietary approaches. That is the value that should be highlighted. Profiles may be useful for identifying features that can / should be adopted into a standard / base schema over time. I think this is what is meant by "only when we say so".

nedmsmith: > ok, fine. However what about the other aspect of extend also ? It seems to say that we only Constrain the types.

CDDL is a language of 'types'. Even maps and records are typing information. A schema is a type structure that is used to check a data structure for type compliance. It's OK to refer to the schema as a collection of 'type' structures IMHO. Otherwise, its ok to use 'schema', as in: "A profile MAY constrain or extend CoRIM schema statements."

draft-ietf-rats-corim.md
```diff
@@ -387,8 +387,22 @@ The following describes each child element of this type.
...
-are prohibited and which extension points are exercised and how.
+Profiling is the mechanism that allows the base CoRIM schema to be customised to fit a specific Attester.
+
+A profile defines which of the optional parts of a CoRIM are required,
+which are prohibited and which extension points are exercised and how.
+Exercised extension points should preserve the intent of the original map semantics.
+A profile MUST NOT alter the syntax or semantics of an already defined CoRIM type.
+A profile MAY constrain the values of a given CoRIM type to a subset of the values.
+A profile MAY extend the set of a given CoRIM type to meet the requirement of a specific profile using the defined extension points.
+
```
thomas-fossati, line 398:
```suggestion
A profile MUST NOT alter the syntax or semantics of an already defined CoRIM type.
A profile MAY constrain the values of a given CoRIM type to a subset of the values.
A profile MAY extend the set of a given CoRIM type to meet the requirement of a specific profile using the defined extension points.
Exercised extension points should preserve the intent of the original map semantics.

```

draft-ietf-rats-corim.md
```diff
@@ -387,8 +387,22 @@ The following describes each child element of this type.
...
 
-A profile specifies which of the optional parts of a CoRIM are required, which
-are prohibited and which extension points are exercised and how.
+Profiling is the mechanism that allows the base CoRIM schema to be customised to fit a specific Attester.
+
+A profile defines which of the optional parts of a CoRIM are required,
+which are prohibited and which extension points are exercised and how.
+A profile MUST NOT alter the syntax or semantics of an already defined CoRIM type.
+A profile MAY constrain the values of a given CoRIM type to a subset of the values.
+A profile MAY extend the set of a given CoRIM type to meet the requirement of a specific profile using the defined extension points.
```
thomas-fossati, line 396:
```suggestion
A profile MAY extend the set of a given CoRIM type using the defined extension points.
```

draft-ietf-rats-corim.md
```diff
@@ -387,8 +387,22 @@ The following describes each child element of this type.
...
-A profile specifies which of the optional parts of a CoRIM are required, which
-are prohibited and which extension points are exercised and how.
+Profiling is the mechanism that allows the base CoRIM schema to be customised to fit a specific Attester.
+
+A profile defines which of the optional parts of a CoRIM are required,
+which are prohibited and which extension points are exercised and how.
+A profile MUST NOT alter the syntax or semantics of an already defined CoRIM type.
+A profile MAY constrain the values of a given CoRIM type to a subset of the values.
+A profile MAY extend the set of a given CoRIM type using the defined extension points.
+Exercised extension points should preserve the intent of the original map semantics.
```
thomas-fossati, line 397:
```suggestion
Exercised extension points should preserve the intent of the original semantics.
```

draft-ietf-rats-corim.md
```diff
@@ -387,8 +387,22 @@ The following describes each child element of this type.
 
 ### Profile Types {#sec-corim-profile-types}
 
-A profile specifies which of the optional parts of a CoRIM are required, which
-are prohibited and which extension points are exercised and how.
+Profiling is the mechanism that allows the base CoRIM schema to be customised to fit a specific Attester.
+
+A profile defines which of the optional parts of a CoRIM are required,
+which are prohibited and which extension points are exercised and how.
+A profile MUST NOT alter the syntax or semantics of an already defined CoRIM type.
```
yogeshbdeshpande, line 394:
```suggestion
A profile MUST NOT alter the syntax or semantics of a CoRIM type defined in base schema.
```

draft-ietf-rats-corim.md
```diff
@@ -387,8 +387,22 @@ The following describes each child element of this type.
...
+A profile MAY extend the set of a given CoRIM type using the defined extension points (see {{sec-extensibility}}).
+Exercised extension points should preserve the intent of the original semantics.
+
+CoRIM profiles SHOULD be specified in a publicly available document.
+
+A CoRIM profile can use one of the base CoRIM media types defined in {{sec-mt-corim-signed}} and
+{{sec-mt-corim-unsigned}} with the `profile` parameter set to the appropriate value.
+Alternatively, it MAY define and register its own media type.
+
+A profile identifier is either an OID {{-cbor-oids}} or a URL {{-uri}}.
```
yogeshbdeshpande, line 405: A profile is fixed once it is released, any subsequent change will lead to a new profile version.

thomas-fossati: done in [48d4739](https://github.com/ietf-rats-wg/draft-ietf-rats-corim/pull/76/commits/48d47398efef6db443417ce410811ef6bae23d9b)

yogeshbdeshpande (approved):
LGTM!

Thank you for all the work!

Issue #84, opened 2023-04-26
Verifier behavior for tagged-thumbprint-type


nedmsmith:
The `tagged-thumbprint-type` is used by `$crypto-key-type-choice` which is used in the `identity-triple-record`, `attest-key-triple-record` and `authorized-by` code point in various other structures. The verifier behavior for `tagged-thumbprint-type` could be interpreted different ways that are not documented. Both the identity and attest-key triples' descriptions indicate use of the key is a prerequisite. A key digest can't be used directly to verify private key usage. The verifier needs the public key. There could be two cases where the key is verified, (a) verified directly by the verifier who must use the digest to find the key, (b) verified by split verifiers where a first verifier does a signature check and a second verifier completes the signature check by checking CRLs and does certificate path validation. For (b), the first verifier may use `tagged-thumbprint-type` to identify the partially verified key such that the second verifier completes the process. It's assumed that the second verifier has a way to trust the first verifier.

The `tagged-thumbprint-type` definition says the digest could be of a raw key or a certificate. This implies the verifier maintains a lookup table for both or that there is enough context in the triple for the verifier to know which type of lookup is appropriate. Possibly, it makes sense to define a new thumbprint type for certificate and constrain the current thumbprint type to keys only to avoid verifier having to have additional context?

`tagged-thumbprint-type = #6.557(digest)` is described as a key thumbprint
`tagged-cert-thumbprint-type = #6.559(digest)` is described as a certificate thumbprint

Language for tagged thumbprints should describe the split verifier scenario where the first verifier performs the signature check and the second verifier performs CRL / cert status and path validation checks.

yogeshbdeshpande:
Please also refer to : https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/82

nedmsmith:
Fixed by #85

PR #85, opened 2023-04-26
Crypto key issues

nedmsmith:
PR that addresses issues #83 and #84

### DIFF
--- a/cddl/cose-key.cddl
+++ b/cddl/cose-key.cddl
@@ -1,0 +1,10 @@
+COSE_KeySet = [ + COSE_Key ]
+
+COSE_Key = {
+    1 => tstr / int
+    ? 2 => bstr 
+    ? 3 => tstr / int 
+    ? 4 => [+ (tstr / int) ]
+    ? 5 => bstr
+    * tstr => any
+}
--- a/cddl/cose-key.cddl
+++ b/cddl/cose-key.cddl
@@ -6,5 +6,5 @@ COSE_Key = {
     ? 3 => tstr / int 
     ? 4 => [+ (tstr / int) ]
     ? 5 => bstr
-    * tstr => any
+    * cose-label => cose-value
 }
--- a/cddl/crypto-key-type-choice.cddl
+++ b/cddl/crypto-key-type-choice.cddl
@@ -4,6 +4,7 @@ $crypto-key-type-choice /= tagged-pkix-base64-cert-path-type
 $crypto-key-type-choice /= tagged-cose-key-type
 $crypto-key-type-choice /= tagged-thumbprint-type
 $crypto-key-type-choice /= tagged-cert-thumbprint-type
+$crypto-key-type-choice /= tagged-cert-path-thumbprint-type
 
 tagged-pkix-base64-key-type = #6.554(tstr)
 tagged-pkix-base64-cert-type = #6.555(tstr)
@@ -11,3 +12,4 @@ tagged-pkix-base64-cert-path-type = #6.556(tstr)
 tagged-thumbprint-type = #6.557(digest)
 tagged-cose-key-type = #6.558(COSE_KeySet / COSE_Key)
 tagged-cert-thumbprint-type = #6.559(digest)
+tagged-cert-path-thumbprint-type = #6.560(digest)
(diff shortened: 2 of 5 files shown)


cddl/cose-key.cddl
```diff
@@ -0,0 +1,10 @@
+COSE_KeySet = [ + COSE_Key ]
```
shnwc, line 1: I don't see yet why `COSE_KeySet` should be supported.

cddl/crypto-key-type-choice.cddl
```diff
@@ -1,9 +1,13 @@
...
 $crypto-key-type-choice /= tagged-pkix-base64-cert-path-type
+$crypto-key-type-choice /= tagged-cose-key-type
 $crypto-key-type-choice /= tagged-thumbprint-type
+$crypto-key-type-choice /= tagged-cert-thumbprint-type
 
 tagged-pkix-base64-key-type = #6.554(tstr)
 tagged-pkix-base64-cert-type = #6.555(tstr)
 tagged-pkix-base64-cert-path-type = #6.556(tstr)
 tagged-thumbprint-type = #6.557(digest)
+tagged-cose-key-type = #6.558(COSE_KeySet / COSE_Key)
```
shnwc, line 12:
```suggestion
tagged-cose-key-type = #6.558(COSE_Key)
```
I don't see yet why COSE_KeySet should be supported. Everywhere `$crypto-key-type-choice` is used, it already is contained in an array. Also every `$crypto-key-type-choice` definition contains data for a single key or cert, it would be good to keep the new definition `tagged-cose-key-type` consistent.

cddl/cose-key.cddl
shnwc: How about adding the comments in the COSE_Key definition in [RFC8152] (https://datatracker.ietf.org/doc/html/rfc8152#section-7) to here?

cddl/crypto-key-type-choice.cddl
shnwc: Since RFC8152 defines both `COSE_Key` and `COSE_KeySet`, after further consideration, I think it makes sense to support both, to align with the RFC.

draft-ietf-rats-corim.md
```diff
@@ -86,6 +86,7 @@ normative:
       ITU-T: Recommendation X.690
     target: https://www.itu.int/rec/T-REC-X.690
   IANA.named-information: named-info
+  RFC8152: cose
```
thomas-fossati, line 89: we already have an entry for COSE (as STD96), please use that (which also points to the COSEbis document)

thomas-fossati (approved):
LGTM modulo adjusting the COSE ref

nedmsmith:
I added a thumbprint for certificate path where a derived cert path may need to be represented as a digest

draft-ietf-rats-corim.md
```diff
@@ -1008,10 +1009,24 @@ A cryptographic key can be one of the following formats:
...
+
+A cryptographic key digest can be one of the following formats:
+
+* `tagged-thumbprint-type`: a `digest` of a raw public key. The digest value may
+  be used to find the public key if contained in a lookup table.
+
+* `tagged-cert-thumbprint-type`: a `digest` of a certificate.
+  The digest value may be used to find the certificate if contained in a lookup table.
+
+* `tagged-cert-path-thumbprint-type`: a `digest` of a certificate path.
```
yogeshbdeshpande, line 1023:
```suggestion
* `tagged-cert-path-thumbprint-type`: a `digest` of a certification path.
```

yogeshbdeshpande (approved):
LGMT! Please incorporate review comments

draft-ietf-rats-corim.md
```diff
@@ -87,6 +87,7 @@ normative:
     target: https://www.itu.int/rec/T-REC-X.690
   IANA.named-information: named-info
 
+
```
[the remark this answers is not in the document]
henkbirkholz: If we add that reference, I'd recommend to add below line 59.
Where will that reference be used in text?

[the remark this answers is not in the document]
nedmsmith: I assume it would be informative (not sure what "below line 59" means). A reference to a pkix- thing might cite it or not based on whether or not it is a normative reference correct?

cddl/crypto-key-type-choice.cddl
nedmsmith: Since RFC8152 defines the structure, the COSE_Key CDDL doesn't need to be incorporated into the prose. However, the cddl file is useful for building examples.

Issue #63, opened 2023-03-03
flags-map extensions


nedmsmith:
A TCG specification extends the flags-map as follows:
```
;
; Flags map extensions
;
$$flags-map-extension //= (
  ? &(runtime-meas: 6) => bool
  ? &(immutable-meas: 7) => bool
  ? &(tcb-meas: 8) => bool
)
```

Should CoRIM incorporate this extension into the base (as several of the other extension have already been incorporated such as `domain-dependency-triples`)?

nedmsmith:
Fixed by PR #94

PR #94, opened 2023-05-23
Fixes issue #63 and #100

nedmsmith:
Fixes issue #63

### DIFF
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -957,29 +957,29 @@ If a `flags-map` value is not specified, then the operational mode is unknown.
 
 The following describes each member of the `flags-map`:
 
-* `configured` (index 0): If the flag is true, the measured environment is fully configured for normal operation.
+* `is-configured` (index 0): If the flag is true, the measured environment is fully configured for normal operation.
 
-* `secure` (index 1): If the flag is true, the measured environment's configurable
+* `is-secure` (index 1): If the flag is true, the measured environment's configurable
 security settings are fully enabled.
 
-* `recovery` (index 2): If the flag is true, the measured environment is NOT in recovery
+* `is-recovery` (index 2): If the flag is true, the measured environment is in recovery
 mode.
 
-* `debug` (index 3): If the flag is true, the measured environment is in a debug enabled
+* `is-debug` (index 3): If the flag is true, the measured environment is in a debug enabled
 mode.
 
-* `replay-protected` (index 4): If the flag is true, the measured environment is
+* `is-replay-protected` (index 4): If the flag is true, the measured environment is
 protected from replay by a previous image that differs from the current image.
 
-* `integrity-protected` (index 5): If the flag is true, the measured environment is
+* `is-integrity-protected` (index 5): If the flag is true, the measured environment is
 protected from unauthorized update.
 
-* `runtime-meas` (index 6): If the flag is true, the measured environment is measured
+* `is-runtime-meas` (index 6): If the flag is true, the measured environment is measured
 after being loaded into memory.
 
-* `immutable` (index 7): If the flag is true, the measured environment is immutable.
+* `is-immutable` (index 7): If the flag is true, the measured environment is immutable.
 
-* `tcb` (index 8): If the flag is true, the measured environment is a trusted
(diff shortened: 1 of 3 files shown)


draft-ietf-rats-corim.md
```diff
@@ -957,24 +957,30 @@ If a `flags-map` value is not specified, then the operational mode is unknown.
...
+* `configured` (index 0): If the flag is true, the measured environment is fully configured for normal operation.
 
-* `secure` (index 1): The measured environment's configurable security settings
-  are fully enabled if the flag is true.
+* `secure` (index 1): If the flag is true, the measured environment's configurable
+security settings are fully enabled.
 
-* `recovery` (index 2): The measured environment is NOT in a recovery state if
-  the flag is true.
+* `recovery` (index 2): If the flag is true, the measured environment is NOT in recovery
```
yogeshbdeshpande, line 965: Just wondering how come we reversed the logic of this flag.

Ideally if `recovery` is true then measured environment should be in the recovery mode isn't it?

nedmsmith: > Just wondering how come we reversed the logic of this flag.

Originally, the first 3 flags were defined by TCG Attestation Architecture spec. They were described in ASN.1 as a bit array. They wanted the bit array to be zero if no security significant states were present. Interestingly, the debug flag in comid didn't preserve the reverse logic. Subsequent additions to the flags haven't preserved this convention. The most recent definition of flags in the asn.1 structure includes a mask and a max-size marker making it less important for the bit array to be zero for no security relevant concerns.

It seems odd that recovery has negative logic. Does it make sense to change this? What is the impact?

draft-ietf-rats-corim.md
```diff
@@ -957,24 +957,30 @@ If a `flags-map` value is not specified, then the operational mode is unknown.
...
 
-* `integrity-protected` (index 5): The measured environment is protected from
-  unauthorized update if the flag is true.
+* `integrity-protected` (index 5): If the flag is true, the measured environment is
+protected from unauthorized update.
+
+* `runtime-meas` (index 6): If the flag is true, the measured environment is measured
+after being loaded into memory.
+
+* `immutable` (index 7): If the flag is true, the measured environment is immutable.
```
shnwc, line 980: As shown in issue #63, the names in TCG extension are `immutable-meas` and `tcb-meas`. Why use different names here? Or how about change them to be easier to understand, like `runtime-measured`, `is-immutable`, and `is-tcb`.

By the way, is `tcb` redundant? Since `replay-protected` and `integrity-protected` already implies that the attester is a TCB.

I think we need new flag `confidentiality-protected`.

nedmsmith: > I think we need new flag `confidentiality-protected`.

I think this should be a new issue.

draft-ietf-rats-corim.md
thomas-fossati: `++massive_cognitive_dissonance;`

Not sure who's responsible for fixing this though.

We could drop it from this PR and make an issue to solve this with upstream, or put it in anyway but still have an explicit, inline warning flag and associated issue.

Happy either way.

draft-ietf-rats-corim.md
thomas-fossati: > Or how about change them to be easier to understand, like `runtime-measured`, `is-immutable`, and `is-tcb`.

:+1:

draft-ietf-rats-corim.md
yogeshbdeshpande: Either way, we need to raise this issue. @nedmsmith : I am OK, if you track this as a defect to fix it downstream in all appropriate documents...

All I do not want this `massive cognitive dissonance` to percolate to exist forever and be pain to readers/interpreters.

nedmsmith: Fixed in latest push.

draft-ietf-rats-corim.md
nedmsmith: fixed in latest push.

thomas-fossati (approved):
LGTM

I guess we could ping-pong forever on naming choices, but luckily the code point is what matters in the end :-)

PR #118, opened 2023-08-14
Instances of accepted claims map were changed to accepted claims set

nedmsmith:
Changed uses of accepted claims map to accepted claims set to align with other uses of accepted claims set.


henkbirkholz (approved):
lgtm. Claims Set is the term to go with. Map would have been more precise, but that's where we are today :-)

yogeshbdeshpande (approved):

nedmsmith:
Lets be consistent. If we decide to change to accepted claims map then we can change all occurances.

PR #123, opened 2023-08-16
Address issue 96 multiple CoBOMs

andrew-draper:


draft-ietf-rats-corim.md
```diff
@@ -1292,9 +1292,10 @@ from the selected CoRIMs.
 CoBOMs which are not within their validity period, or which reference tags
 not available to the verifier, are discarded.
 
-The Verifier MUST activate all tags referenced by a CoBOM.
-
-ISSUE: What does the verifier do if there are multiple CoBOMs available
+If there are multiple CoBOMs available then the Verifier MUST process each
+CoBOM independently. The Verifier MUST activate all tags referenced by
+any valid CoBOM. It is valid to have multiple CoBOMs activating the same
+tag.
```
thomas-fossati, line 1298: suggested rewrite based on f2f discussion:

```suggestion
The Verifier processes all CoBOMs that are valid at the point in time verification occurs, and activates all tags referenced therein.

Note that the Verifier may decide to discard some of the available, valid CoBOMs depending on any locally configured authorisation policies modelling the trust relationships between the Verifier Owner and the relevant supply chain actors.
(Any such authorisation policy is out of scope of CoRIM.)

For example, a composite device ({{Section 3.3 of -rats-arch}}) is likely to be fully described by multiple CoRIMs, each signed by a different supply chain actor.
In such case, the Verifier Owner may instruct the Verifier to only process CoBOMs signed by a trusted integrator and discard any "leaf" CoBOMs signed by ISVs describing the individual components.
```

nedmsmith: c/authorisation/authorization
We discussed 'activation' as being something the supplier does as it means the cobom is the way the supplier describes the composition of the object (product). If there is a 2-stage activation, one part done by suppliers and a 2nd part done by verifiers, then this should be described more clearly.
Can we agree on /supplier/supply chain entity/supply chain actor/ and consistently use the term while also being sensitive to readability?
It's more confusing to introduce "leaf" cobom then define what is meant by leaf. If the description is clear, that should be sufficient. But I don't think the description is clear. Maybe "...discard tags activated by supplier CoBOMs that are not activated by the trusted integrator."

thomas-fossati: > c/authorisation/authorization

done with a sneak edit

> Can we agree on /supplier/supply chain entity/supply chain actor/ and consistently use the term while also being sensitive to readability?

I am ok with any in that list.  Which one should I use?  Is it "supplier?"

thomas-fossati: > [...] Maybe "...discard tags activated by supplier CoBOMs that are not activated by the trusted integrator."

great suggestion, and third sneak edit in a row.

thomas-fossati: > We discussed 'activation' as being something the supplier does as it means the cobom is the way the supplier describes the composition of the object (product). If there is a 2-stage activation, one part done by suppliers and a 2nd part done by verifiers, then this should be described more clearly.

I have probably lost some of the discussion context, if so apologies.

I don't think there are two stages of activation, in my head there's only one stage which is actioned by the verifier on the CoBOM signal coming from the supplier.
The only thing we are adding here is a hi-pass filter based on local policy that the Verifier Owner can inject in the CoBOM processing.

yogeshbdeshpande: @thomas-fossati I do not think the thought process you have in mind is reflected in the PR.

The original intent of the PR was: 
1) Can Multiple authorities allowed to issue CoBOM's ?

2) If yes, then when they issue Delta from the old one, does the new one add to the old one or replaces the old ones.

3. To me the Verifier processes a CoBOM as soon as it receives it.

thomas-fossati: I am slightly confused:

> The original intent of the PR was:
> 
> 1. Can Multiple authorities allowed to issue CoBOM's ?

OK. The text I put forward says: yes.

> 2. If yes, then when they issue Delta from the old one, does the new one add to the old one or replaces the old ones.

I am not sure what a "delta CoBOM" and I don't think it was mentioned -- at least in the context of this PR.  Can you clarify?

> 3. To me the Verifier processes/activates a CoBOM as soon as it receives it.

What about the validity period?

nedmsmith: Normally, the signer of a document is "speaking". The document contains the assertions asserted by the signer.  The signer asserts the validity as well as the bill of materials - which says these "materials belong together". There is contradictory text in the current description of CoBOM.
"A Concise Bill of Material (CoBOM) object represents the signal for the Verifier to activate the listed tags. Data contained in a tag MUST NOT be used for appraisal until a CoBOM which activates that tag has been received and successfully processed."
The first sentence says the Verifier activates the CoBOM. The second sentence says the CoBOM activates the tag. CoBOMs are signed by RIM signers which can be suppliers. The second sentence says the Verifier can't use the CoBOM until the signature (signer) is "processed" which implies the Verifier authenticated the signer somehow. In essence, the Verifier's "activation" is simply accepting what the supplier said. 

The appraisal processing question therefore is "what if multiple suppliers say different things" and the Verifier trusts multiple suppliers? How does the Verifier not get into an undecidability quandary? 

This is the question that we should focus on.

Most often, trust is qualified by "trusted for what?" If the Verifier is asked to appraise a laptop computer, then there should be a policy that identifies the laptop supplier. The signer of a CoRIM/CoBOM by the "laptop vendor" determines which CoBOM to trust.

Since a composition (description logic) is expected to be hierarchical, the laptop description will contain other suppliers' components each qualified by the type of thing that it is e.g., "motherboard supplier" or "NIC supplier". The Verifier would rely on the policies having component / composition qualifiers to avoid undecidability cases.

Note that since CoBOM relies on (comid?) tags as the unit of disambiguation, for what is in the BOM. The (comid?) tags cannot contain descriptions of multiple objects (say a NIC and a GPS) - that is unless the laptop supplier intended to include both NIC and GPS. Otherwise, there is no way for the laptop supplier to remove the GPS related triples using CoBOMs. 

A simple solution is to constrain (comid) tags to describe one logical thing such as a laptop, or a NIC, or a motherboard. If the NIC supplier doubles as a GPS supplier and it offers a product that is a combo NIC+GPS component, then this is an indivisible thing. If the laptop OEM doesn't want the GPS, then the supplier MUST create a tag for just the NIC.

henkbirkholz: Slight change in wording.

```suggestion
suggested rewrite based on f2f discussion:

```suggestion
The Verifier processes all CoBOMs that are valid at the point in time appraisal of Evidence occurs, and activates all tags referenced therein.

A Verifier may decide to discard some of the available and valid CoBOMs depending on any locally configured authorization policies.
(Such policies model the trust relationships between the Verifier Owner and the relevant suppliers, and are out of scope of the present document.)

For example, a composite device ({{Section 3.3 of -rats-arch}}) is likely to be fully described by multiple CoRIMs, each signed by a different supplier.
In such case, the Verifier Owner may instruct the Verifier to discard tags activated by supplier CoBOMs that are not activated by the trusted integrator.
```

nedmsmith: I'm OK with this wording. Maybe a little verbose, but that can be fixed later.

nedmsmith (approved):

draft-ietf-rats-corim.md
henkbirkholz: ack

henkbirkholz: lgtm for today's merge interval

yogeshbdeshpande (approved):
One minor comment, but in general LGTM!

Issue #126, opened 2023-08-21
Key measurements that don't have prescribed semantics


nedmsmith:
The `identity-triple-record` and `attest-key-triple-record` have encoded key usage semantics into the triple predicate. 

There are vendor specific or otherwise additional possible key usages that differ. 

One way to overload the above triples with different key usage semantics is to allow the `environment-map` to carry vendor-specific key usage semantics. For example, a `class-id` OID might define an arc for software update (.111.0) and enumerate keys from the arc (.111.1, .111.2, ..., .111.n) etc.

Alternatively, the $crypto-key-type-choice could be included in `measurement-values-map` with the semantics of "this key belongs to this environment". Omitting key usage semantics.

The option to encode key usage semantics into the `environment-map` is still available. 

If, for example, there is a SW update use case tied to attestation, the SW update Relying Party code will recognize the use case context and as needed check the environment context or identify an alternative form (such as a certificate) that contains additional key usage context.

The minimal attestation statement is "this key belongs to this environment". 

`measurement-values-map` becomes:
```
measurement-values-map = non-empty<{
  ? &(version: 0) => version-map
  ? &(svn: 1) => svn-type-choice
  ? &(digests: 2) => [ + digest ]
  ? &(flags: 3) => flags-map
  ? (
      &(raw-value: 4) => $raw-value-type-choice,
      ? &(raw-value-mask: 5) => raw-value-mask-type
    )
  ? &(mac-addr: 6) => mac-addr-type-choice
  ? &(ip-addr: 7) =>  ip-addr-type-choice
  ? &(serial-number: 8) => text
  ? &(ueid: 9) => ueid-type
  ? &(uuid: 10) => uuid-type
  ? &(name: 11) => text
  ? &(keyid: 12) => $crypto-key-type-choice
  * $$measurement-values-map-extension
}>
```
Identity and attest key triples can remain in `triples-map` since the triple predicate as additional semantics over what is proposed.
```
...
  ? &(identity-triples: 2) =>
    [ + identity-triple-record ]
  ? &(attest-key-triples: 3) =>
    [ + attest-key-triple-record ]
...
```

nedmsmith:
`identity-triples` and `attest-key-triples` should be updated with text that describes the verifier behavior when these triples are asserted as Endorsed Values and as Evidence/Reference Values.

nedmsmith:
See also PR #135

nedmsmith:
[...]

PR #127, opened 2023-08-21
Adds crypto key measurement type

nedmsmith:
Issue #126  calls for a $crypto-key-type-choice measurement that isn't stereotyped for either identity or attestation, but can support a wide range of key usages. It also allows authorized-by context to be added and works with conditional endorsement.

### DIFF
--- a/cddl/measurement-values-map.cddl
+++ b/cddl/measurement-values-map.cddl
@@ -13,5 +13,6 @@ measurement-values-map = non-empty<{
   ? &(ueid: 9) => ueid-type
   ? &(uuid: 10) => uuid-type
   ? &(name: 11) => text
+  ? &(cryptokeys: 12) => [ + $crypto-key-type-choice ]
   * $$measurement-values-map-extension
 }>
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -941,7 +941,7 @@ The following describes each member of the `measurement-values-map`.
 
 * `cryptokeys` (index 12): identifies cryptographic keys that are protected by the Target Environment
   See {{sec-crypto-keys}} for the supported formats.
-  An Attesting Environment determines that keys are protected as part of Claims Collection.
+  An Attesting Environment determines that keys are protected as part of Claims collection.
   Appraisal verifies that, for each value in `cryptokeys`, there is a matching Reference Value entry.
 
 ###### Version {#sec-comid-version}
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -1654,7 +1654,7 @@ are compared with the bytes following the CBOR tag from the Accepted Claims Set 
 If the byte strings match, and there is another array entry,
 then the next entry from the Reference Values array is likewise
 compared with the next entry of the Accepted Claims Set array.
-If all entrys of the Reference Values array matches a corresponding entry in the Accepted Claims Set array, then the `cryptokeys` Reference Value matches.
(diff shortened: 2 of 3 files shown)


jdbeaney (approved):

cddl/measurement-values-map.cddl
```diff
@@ -13,5 +13,6 @@ measurement-values-map = non-empty<{
   ? &(ueid: 9) => ueid-type
   ? &(uuid: 10) => uuid-type
   ? &(name: 11) => text
+  ? &(cryptokeys: 12) => [ + $crypto-key-type-choice ]
```
andrew-draper, line 16: I think making this an array makes things more complicated. Can we make this a singleton?

nedmsmith, line 16: This is an array because there can be multiple keys that are protected by the environment. The verifier is checking that the same list of keys reported in Evidence is included in Reference Values. The Verifier will not use or verify the keys unless the RP or a downstream verifier instructs this. But we can say the same for any measurement (that RP or downstream verifier may post process the claims). The default Verifier action is to compare each Evidence key at an array position with each Reference Value key at the same array position.

yogeshbdeshpande:
@nedmsmith we need more description of how this is relevant, what use cases and context this would be used!

nedmsmith:
> @nedmsmith we need more description of how this is relevant, what use cases and context this would be used!

See #126 discussion

nedmsmith:
Note: This PR is 17 commits behind main. Description of verifier behavior should be applied to sections introduced into main. Possibly need to rebase or apply this PR and add text via another PR.

nedmsmith:
There's an interesting question around $crypto-key-type-choice measurements and traditional digest measurements in terms of whether the verifier should apply different appraisal behaviors to each. If a simple byte-wise compare is wanted, the key object could be digested and placed in a `digests` measurement. But as a key, it (should?) apply crypto key matching semantics. 
If a cert chain is supplied, then the key could be in an alternate form (e.g., Subject Public Key in X.509) vs. say a digest of the key. The verifier may to cert path validation on the certificate to weed out invalid keys. (This seems equivalent to offline key verification checks called for by PR #135). 
Note that key as a measurement doesn't require the key to be exercised (as in signing evidence), but presumably, the key provenance found in the ACS is useful to RPs for downstream applications.

yogeshbdeshpande (approved):
LGTM!

draft-ietf-rats-corim.md
```diff
@@ -1632,6 +1641,21 @@ which stores [expect-raw-value raw-value-mask] in an array?
...
+The array contains one or more entries in sequence.
+
+The CBOR tag of the first entry of the Reference Value `cryptokeys` array is compared with
+the CBOR tag of the first entry of the Accepted Claims Set `cryptokeys` value.
+If the CBOR tags match, then the bytes following the CBOR tag from the Reference Value entry
+are compared with the bytes following the CBOR tag from the Accepted Claims Set entry.
+If the byte strings match, and there is another array entry,
+then the next entry from the Reference Values array is likewise
+compared with the next entry of the Accepted Claims Set array.
+If all entrys of the Reference Values array matches a corresponding entry in the Accepted Claims Set array, then the `cryptokeys` Reference Value matches.
```
thomas-fossati, line 1656:
```suggestion
If all entries of the Reference Values array match a corresponding entry in the Accepted Claims Set array, then the `cryptokeys` Reference Value matches.
```

thomas-fossati (approved):
LGTM.  I've inlined a few, purely editorial, comments.

Issue #132, opened 2023-08-31
Collision for CBOR tag 560


setrofim:
CBOR tag `#6.560` is used for both `tagged-cert-path-thumbprint-type` and `tagged-bytes` types (See: https://ietf-rats-wg.github.io/draft-ietf-rats-corim/draft-ietf-rats-corim.html#name-raw-values-types and https://ietf-rats-wg.github.io/draft-ietf-rats-corim/draft-ietf-rats-corim.html#name-crypto-keys).

nedmsmith:
There are earmarks for specific CBOR tags for xcorim (525-527) and concise-evidence (570-571). Should these be included in the table? Should the repo maintain a local registry?

nedmsmith:
fixed by PR #133

PR #133, opened 2023-08-31
fixed duplicate use of cbor tag 560

nedmsmith:
changed $crypto-keys-type-choice to use 561. Fixes issue #132


babongo:
Thanks @setrofim for spotting and @nedmsmith for the patch.

I'd go further here and add a bit of logics to the pipeline to automatically check for duplications, so that we don't end up in the same situation in the future.

nedmsmith:
I'm open to suggestions for how to do that.

thomas-fossati:
> I'm open to suggestions for how to do that.

Whenever we add a new tag we should re-run the `cbor-tags.txt` target to update the tags catalogue.

We could couple that with a check for duplicates in the `cbor-tags.txt` file, e.g.:
```
grep -E -h -o '#6\.[0-9]{3}' cbor-tags.txt | uniq -cd
```
and exit non-zero if there's at least one line.

setrofim (approved):

cddl/Makefile
```diff
@@ -57,3 +58,9 @@ clean: ; rm -f $(CLEANFILES)
 
 # Extract the CBOR tags defined by CoRIM/CoMID (i.e., those in the 5xx space)
 cbor-tags.txt: $(wildcard *.cddl) ; grep -h '#6\.5' *cddl | sort -u -t'=' -k2 > $@
+
+# Return an error if there are duplicat CBOR Tags defined
+cbor-tags-unique: $(wildcard *.cddl) ; \
+	 grep -h '#6\.5' *cddl | sort -u -t'=' -k2 > cbor-tags.txt ; \
```
thomas-fossati, line 64: I guess this target should depend explicitly on `cbor-tags.txt` -- this way L64 becomes redundant

cddl/Makefile
```diff
@@ -57,3 +58,9 @@ clean: ; rm -f $(CLEANFILES)
...
+	 grep -E -h -o '#6\.[0-9]{3}' cbor-tags.txt | uniq -cd ; \
+	 if [ ! -z $1 ]; then exit 1; else echo "CBOR tags are unique"; fi
```
thomas-fossati, line 66: `$1` is a positional parameter, I don't think this will expand to anything useful.

thomas-fossati (commented):
Thanks for taking a stab!  I'll provide a couple of tweaks

thomas-fossati (approved):

nedmsmith:
LGTM

Issue #140, opened 2023-09-06
Verifier processing of `authorized-by`


nedmsmith:
PR #107 line 1528, presumes there is no authority context without use of `authorized-by` statements. However, given all RVs are signed / authenticated, all RVs have an authorized-by context. The same is true for Evidence / ACS. All Evidence is signed / authenticated, hence has an authorized-by context. I don't think the conditional wording is needed, but rather to explain better the authorized-by context.

Namely, that all claims are entered into the ACS with authority context (either by assigning the corim signer `corim-signer-map`, or by `authorized-by` in `measurement-map`.

Note: `coswid-triples`, `identity-triples` and `attest-key-triples` need to be updated to include `authorized-by` since they don't use `measurement-map`.

andrew-draper:
I agree this should be done, and we need to define authorized by better.
One thing to note is that we need description of use cases for this header, and to describe how authorized by is filled in when evidence/endorsements are added to the ACS

nedmsmith:
A partial solution to this issue involves augmenting `identity-triples` and `attest-key-triples` with `authorized-by`. The solution proposed here builds on PR #127 which adds ` ? &(cryptokeys: 12) => [ + $crypto-key-type-choice ]` measurement to `measurement-values-map`. 
The current `identity-triples` uses:
```
identity-triple-record = [ 
     environment-map, 
     [ + $crypto-key-type-choice ] 
]
```
which doesn't include `authorized-by`.
```
triples-map = {
...
  ? &(identity-triples: 2) =>
    [ + identity-triple-record ]
  ? &(attest-key-triples: 3) =>
    [ + attest-key-triple-record ]
...
}
```
Instead, identity-triple-record is expanded:
```
identity-triple-record = [ 
     environment-map, 
     [ + $crypto-key-type-choice ] / measurement-map
]
```
and
```
attest-key-triple-record = [ 
     environment-map, 
     [ + $crypto-key-type-choice ] / measurement-map
]
```
Since, measurement-values-map contains `[ + $crypto-key-type-choice ]` the same key information can be described in the measurement value, but since measurement-map is included, the `authorized-by` context can be supplied.

Retaining the current definition prevent breaking backward compatibility.

nedmsmith:
The coswid triple could be instrumented with `authroized-by` similarly:
```
coswid-triple-record = [
  environment-map
  [ + concise-swid-tag-id ]
]

[...]

PR #145, opened 2023-09-08
Add verifier behaviour for conditional triples

andrew-draper:


draft-ietf-rats-corim.md
```diff
@@ -1491,36 +1491,97 @@ and how the verifier matches a Reference Value against the Accepted Claims Set
...
+The Verifier SHALL try to match all Reference Values contained within all
+`reference-triple-record`s within the group against Evidence and Endorsements
+in the Accepted Claims Set. {{sec-match-one-ref-val}} describes how each
 Reference Value is matched against the Accepted Claims Set.
 
+If all Reference Values in a group match entries in the Accepted Claims Set
+then all Endorsed Values from `endorsed-triple-record`s in the group are added
+to the Accepted Claims Set (see {{sec-add-to-acs}}).
+
 If any Reference Value in a group does not match the Accepted Claims Set then
```
nedmsmith, line 1517: I think "Accepted Claims Set"  is ambiguous given the above statement that ACS may contain Endorsements as well as Evidence. The algorithm for accepting Evidence applies to Reference Values. If the Attester supplies no Evidence that is vetted by Reference Values, then no Endorsed values should be attributed IMO. Simple example, if Attester asserts it is an HSM, but no Ref Vals substantiate this, and a FIPS certifying body produces an Endorsement that says HSM is EAL4+, then the Attester has successfully achieved privilege escalation.

draft-ietf-rats-corim.md
```diff
@@ -1491,36 +1491,97 @@ and how the verifier matches a Reference Value against the Accepted Claims Set
...
+This section describes how a Reference Value is matched against Evidence or Endorsements
+in the Accepted Claims Set.
+If any part of the processing indicates that the Reference Value does not match
+then the remaining steps in this section are skipped for that group.
 
-A Reference Value consists of an `environment-map` plus a `measurement-map`. In the
-`reference-values-triple-record` these are packaged together. In other triples multiple
-Reference Values are represented more compactly by letting one `environment-map`
-apply to multiple `measurement-map`s.
+A Reference Value consists of an `environment-map` plus a `measurement-map`. It is
```
nedmsmith, line 1583: I still think this is too narrowly stated. If taken literally, the identity-triples, attest-key-triples, conditional triples and coswid-triples have to be discarded because their "measurements" aren't describes as `measurement-map`.

draft-ietf-rats-corim.md
```diff
@@ -1491,36 +1491,97 @@ and how the verifier matches a Reference Value against the Accepted Claims Set
...
+>> *TODO: Is a Conditional Endorsement Series Triple affected by
+`reference-triple-record`s in the same group? I think not.
+Does this need to be mentioned?*
 
 #### Matching a Reference Value against the Accepted Claims Set {#sec-match-one-ref-val}
 
-This section describes how a Reference Value is matched against Evidence in the Accepted
-Claims Set.
-If any part of the processing indicates that the Reference Value does not match then the remaining steps in this section are skipped for that group.
+This section describes how a Reference Value is matched against Evidence or Endorsements
```
nedmsmith, line 1578: What are some use cases for Ref Values matching Endorsements?

draft-ietf-rats-corim.md
```diff
@@ -1491,36 +1491,97 @@ and how the verifier matches a Reference Value against the Accepted Claims Set
...
+SHALL attempt to match the Reference Value part of each entry against entries
+in the Accepted Claims Set (see {{sec-match-one-ref-val}}).
+If the Reference Value matches then the verifier SHALL add the corresponding
+Endorsed Value to the Accepted Claims Set (see {{sec-add-to-acs}}) and
+SHALL skip processing the remaining entries in the array.
 
-Each group is processed independently of other groups. If a group fails to match
-the Accepted Claims Set then this does not affect the processing of other groups.
+>> *TODO: Is a Conditional Endorsement Series Triple affected by
+`reference-triple-record`s in the same group? I think not.
```
nedmsmith, line 1573: It could be if the stateful-environment-record mentions endorsement claims that were added via a reference triple.

draft-ietf-rats-corim.md
```diff
@@ -1491,36 +1491,97 @@ and how the verifier matches a Reference Value against the Accepted Claims Set
...
+If it matches then the verifier SHALL add the Endorsed value to the
+Accepted Claims Set (see {{sec-add-to-acs}}).
+
+>> *TODO: Is a Conditional Endorsement Triple affected by `reference-triple-record`s
+in the same group? I think not. Does this need to be mentioned?*
+
+#### Matching a Conditional Endorsement Series Triple against the Accepted Claims Set
+
+A Conditional Endorsement Series Triple {{sec-comid-triple-cond-series}} is an efficient
+encoding of a series of alternative Reference Values, each with a corresponding
```
nedmsmith, line 1545: I think this is misstated. The series claims are endorsed values, of which one in the series is selected for inclusion in the ACS. The RVs are not a "series" but a stateful environment.

draft-ietf-rats-corim.md
```diff
@@ -1491,36 +1491,97 @@ and how the verifier matches a Reference Value against the Accepted Claims Set
...
+Endorsed Value. The Verifier adds the Endorsed Values corresponding to the first
```
nedmsmith, line 1546:
```suggestion
Endorsed Value.
The Verifier adds the first Endorsed Value in the series that matches the stateful environment.
```

draft-ietf-rats-corim.md
```diff
@@ -1491,36 +1491,97 @@ and how the verifier matches a Reference Value against the Accepted Claims Set
...
+Reference Value which matches.
```
nedmsmith, line 1547:
```suggestion
The stateful enviroment acts like a Reference Value which matches Claims contained in the Accepted Claims Set.
```

draft-ietf-rats-corim.md
```diff
@@ -1491,36 +1491,97 @@ and how the verifier matches a Reference Value against the Accepted Claims Set
...
+
+All Reference Values and Endorsed Values in the triple use the same `environment-map`
```
nedmsmith, line 1549: I don't know what these two lines are trying to say, but it seems incorrect.

draft-ietf-rats-corim.md
```diff
@@ -1491,36 +1491,97 @@ and how the verifier matches a Reference Value against the Accepted Claims Set
...
+Each group is processed independently of other groups. If `reference-triple-record`s
+in a group do not match the Accepted Claims Set then this does not affect
+the processing of other groups.
+
+#### Matching a Conditional Endorsement Triple against the Accepted Claims Set
+
+A Conditional Endorsement Triple {{sec-comid-triple-cond-end}} is an efficient
+encoding of a Reference Value and an Endorsed Value.
+
+The triple's `stateful-environment-record` field contains both the `environment-map` and `measurement-map` parts of the Reference Value.
```
nedmsmith, line 1529:
```suggestion
The triple's `stateful-environment-record` field constitutes the Reference Value which contains both `environment-map` and `measurement-map` Claims.
```

draft-ietf-rats-corim.md
```diff
@@ -1491,36 +1491,97 @@ and how the verifier matches a Reference Value against the Accepted Claims Set
...
+all `endorsed-triple-record`s in the group are silently ignored.
+
+Each group is processed independently of other groups. If `reference-triple-record`s
+in a group do not match the Accepted Claims Set then this does not affect
+the processing of other groups.
+
+#### Matching a Conditional Endorsement Triple against the Accepted Claims Set
+
+A Conditional Endorsement Triple {{sec-comid-triple-cond-end}} is an efficient
+encoding of a Reference Value and an Endorsed Value.
```
nedmsmith, line 1527:
```suggestion
encoding of a Reference Value that describes the result of previously accepted Claims as the condition for accepting additional Endorsed Values.
```

draft-ietf-rats-corim.md
```diff
@@ -1491,36 +1491,97 @@ and how the verifier matches a Reference Value against the Accepted Claims Set
...
+
+The triple's `stateful-environment-record` field contains both the `environment-map` and `measurement-map` parts of the Reference Value.
+
+The Endorsed Value is encoded using the same `environment-map`, plus the
```
nedmsmith, line 1531: This looks incorrect. The triple object is `measurement-values-map` not an environment-map.

draft-ietf-rats-corim.md
```diff
@@ -1491,36 +1491,97 @@ and how the verifier matches a Reference Value against the Accepted Claims Set
...
+`measurement-value-map` field within the Conditional Endorsement Triple.
+
+For each Conditional Endorsement Triple, the verifier SHALL attempt to match the
```
nedmsmith, line 1534:
```suggestion
For each Conditional Endorsement Triple, the verifier SHALL match the
```

draft-ietf-rats-corim.md
```diff
@@ -1491,36 +1491,97 @@ and how the verifier matches a Reference Value against the Accepted Claims Set
...
+Reference Value against entries in the Accepted Claims Set (see {{sec-match-one-ref-val}}).
```
nedmsmith, line 1535:
```suggestion
Reference Value, consisting of a `stateful-environment-record`, against entries in the Accepted Claims Set (see {{sec-match-one-ref-val}}).
```

draft-ietf-rats-corim.md
```diff
@@ -1491,36 +1491,97 @@ and how the verifier matches a Reference Value against the Accepted Claims Set
...
+If it matches then the verifier SHALL add the Endorsed value to the
+Accepted Claims Set (see {{sec-add-to-acs}}).
+
+>> *TODO: Is a Conditional Endorsement Triple affected by `reference-triple-record`s
+in the same group? I think not. Does this need to be mentioned?*
```
nedmsmith, line 1540: It can be if the reference-triple-record contains claims that are in the stateful-environment-record. The expectation is the reference-triples are processed before the conditional triples. (I think).

nedmsmith (commented):
multiple comments

draft-ietf-rats-corim.md
andrew-draper: We should probably discuss this more. My take is that Evidence is automatically added to the ACS with the authorization of the attester. Or more precisely with the authority key which authorized the attester (eg Intel DICE root if the verifier trusts that key).
For most relying parties, that evidence means nothing without a CoRIM to add a conditional endorsement. The FIPS certifying body could say that an entity with measurements A=A1 and B=B1 authorized by key K is FIPS certified.
As we discussed, I think authorized by deserves its own section.

draft-ietf-rats-corim.md
andrew-draper: As you see below I think the spec needs a name for `environment-map` plus `measurement-map` when it appears in conditional triples as well as when it appears in reference-values-triple-record.
I am quite happy to use a new name for EM+MM.
What should it be?

draft-ietf-rats-corim.md
andrew-draper: The main use case is hierarchical CoRIMs. 
- CoMID A has reference values which are expected to match against measurements A and B, if they match it adds endorsement E. 
- CoMID C contains reference values which are expected to match against measurements C and D, and another which is expected to match against endorsement E if authorized by A. If all match then it adds endorsement F.
- The relying party is looking for endorsement F.
I'll put this in a separate PR

draft-ietf-rats-corim.md
andrew-draper: I agree that the reference value part of Conditional Endorsement Series Triple could match against an endorsement added by an earlier group.
I guess what I was asking is whether CEST is in a group or not. If it's in a group then a mismatched reference value in that group could stop it from being processed. That seems odd to me, so I would rather say that CEST is not in a group, but I want to make sure everyone is happy with what I'm writing.

draft-ietf-rats-corim.md
andrew-draper: The is the introductory description of what the triple does, IMHO it should describe the rest of the section in informative words so that a reader knows what is coming next.
I used the word "series" because the verifier processes them in order.
I haven't found a non-circular definition of "stateful environment" so I don't think that is a good term to use instead. It might be if we define it appropriately. Perhaps what I called "Reference Value" is what you think of as "stateful environment"?

draft-ietf-rats-corim.md
andrew-draper: I am wondering whether to do s/Reference Value/stateful environment/ or something like that throughout the whole diff

draft-ietf-rats-corim.md
andrew-draper: I may be misunderstanding your expectations for CET

draft-ietf-rats-corim.md
andrew-draper: I think a change similar to what you propose is good, but it depends on which name we use to replace "Reference Value"

draft-ietf-rats-corim.md
andrew-draper: The next couple of lines say that, so I don't think it needs to say it here as well.

draft-ietf-rats-corim.md
andrew-draper: We need to add the MVM to an EM somewhere. As I understand it, the MVM is added to the same EM as the one in the stateful-environment-record.

draft-ietf-rats-corim.md
andrew-draper: agree

draft-ietf-rats-corim.md
andrew-draper: Again, I am asking about whether CET is a group of its own or is part of another group

draft-ietf-rats-corim.md
andrew-draper: I will address this as part of the patch for issue #140

draft-ietf-rats-corim.md
andrew-draper: Possible suggestions:
environment-measurement-pair
claim tuple

nedmsmith: > environment-measurement-pair

Fine! but probably would not use hyphens with pair. E.g.: "... an environment-measurement pair..." or "... an environment-measurement pairing..."

andrew-draper:
Replaced by #175

Issue #161, opened 2023-10-02
Extensible RIM freshness


nedmsmith:
CoRIM defines a freshness mechanism as:
```
? &(rim-validity: 4) => validity-map
validity-map = {
  ? &(not-before: 0) => time
  &(not-after: 1) => time
}
```
However, `validity-map` is not extensible. Hence, alternative freshness mechanisms such as those described by RFC9334 including epoch markers, can't be used with CoRIM.

If `validity-map` was extensible, then support for alternatives could be added by a profile or later by the base schema.

This issue proposes the following change to base schema:
```
validity-map = {
  ? &(not-before: 0) => time
  &(not-after: 1) => time
  * $$validity-map-extension
}
```

thomas-fossati:
I don't think this is the right place where to add extensibility.

The reason why `validity-map` has very scoped semantics is because it is reused in different contexts (`corim-map`, `corim-meta-map` and `concise-bom-tag`).  Allowing extensibility there makes it hard to track all the subtle semantics differences.

If the use case is Concise Evidence / Concise AR, a freshness indicator can already be added via the `$$corim-map-extension`.

PR #164, opened 2023-10-04
Fix issue #161

nedmsmith:
Adds extension point to validity-map

### DIFF
--- a/cddl/examples/corim-validity.diag
+++ b/cddl/examples/corim-validity.diag
@@ -1,0 +1,51 @@
+/ corim / 500(
+  / corim-map / 501({
+    / corim.id / 0 : h'284e6c3e5d9f4f6b851f5a4247f243a7',
+    / corim.tags / 1 : [
+      / concise-mid-tag / 506( <<
+        / concise-mid-tag / {
+          / comid.tag-identity / 1 : {
+            / comid.tag-id / 0 : h'3f06af63a93c11e4979700505690773f'
+          },
+          / comid.entity / 2 : [ {
+            / comid.entity-name / 0 : "ACME Inc.",
+            / comid.reg-id / 1 : 32("https://acme.example"),
+            / comid.role / 2 : [ 0 ] / tag-creator /
+          } ],
+          / comid.triples / 4 : {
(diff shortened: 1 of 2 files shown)


yogeshbdeshpande (approved):
LGTM!

thomas-fossati (commented):
The semantics of this must be documented in §1.3.3 -- and possibly in all the places it is `validity-map` is reused.

Besides, since this introduces an extension point, we need a new IANA sub-registry.

But more generally, I don't think this is the right design.  (See https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/161#issuecomment-1760732672 for the rationale.)

nedmsmith:
> I am a bit confused by this PR (next to style & syntax).
> 
> This seems to break the straight forward semantics of current validity map. But more importantly, I cannot see how Epoch Markers (that were the example in #161 would solve a problem here. Do they really?

The validity period assumes verifier has trusted time. I'm looking for a solution that doesn't require trusted time. Epoch markers could be used in this sort of environment.

andrew-draper (changes requested):
This looks good, but I think it also needs text which says that if a verifier doesn't understand any of the fields within validity-map then that map is not valid.

nedmsmith:
This branch was rebased to pick up changes in main.

nedmsmith:
> Besides, since this introduces an extension point, we need a new IANA sub-registry.

We can tolerate not having an IANA sub-registry until there is interest in standardizing extensions. In the meantime, a vendor may leverage a profile context to define a vendor-specific approach using negative code points. 

The lack of a defined registry creates a friction point that limits use of the extention point for standard use, thereby limiting possible backwards compatibility issues should another approach become preferred.

nedmsmith:
Move to "won't fix" based on the justification that an alternative form of validity can be added by extending at the corim-map and opting not to use validity?

deeglaze:
Sgrm

yogeshbdeshpande:
Awaiting some actions in EPOCH markers draft to introduce an Extensibility points. @Henk to follow up on this and then only close this Pull Request!

PR #168, opened 2023-10-19
Group Membership & MEC Endorsment

### LINKED ISSUE #136: Need definition of what is a group for the purpose of reference value matching
The text uses the placeholder term "group" to describe reference values and endorsed values which are processed together.
There are different ways proposed to encode the group concept in the CoRIM file
We need to agree how the "group" is encoded, and probably rewrite the document to replace the term "group" with new text

henkbirkholz:
fixes #136


draft-ietf-rats-corim.md
```diff
@@ -1186,6 +1186,78 @@ Evidence.
...
+
+#### Group Membership Triple {#sec-comid-triple-group-membership}
+
+In order to model hierarchical device composition, CoRIM authors need to identify all components in the composite device (see Composite Device in {{-rats-arch}}). A group describes all target environments that a certain attesting environment is generating evidence about. Consequently, all target environments in question are associated with a common domain via the group-membership-triple.
+
+~~~ cddl
+group-membership-triple-record = [
+  relative-top-layer-root: environment-map
+  relative-sub-nodes: [ + environment-map ]
+]
```
andrew-draper, line 1198: I think this doesn't work when different revisions of CoMID have different definitions of what is in an environment

draft-ietf-rats-corim.md
```diff
@@ -1186,6 +1186,78 @@ Evidence.
...
+> "IF accepted state matches the `cond` value, THEN the `env` is associated with the endorsed value `val`."
+
+~~~ cddl
+multi-env-conditional-endorsement-triple-record = [
+  env: environment-map                    ; the environment to which the endorsed value
+                                          ; applies
+  val: measurement-values-map             ; the endorsed value
+  cond: [ + stateful-environment-record ] ; all target environments along with
+                                          ; their expected state that need to match
+]
```
andrew-draper, line 1242: I have a version of this which adds multiple conditional endorsements (with the same condition for each). I think that is likely to be widely used, and for the cost of 1 additional byte is probably worthwhile.

andrew-draper (changes requested):
I think this needs work before we commit it. I have a similar proposal.

yogeshbdeshpande:
> I think this needs work before we commit it. I have a similar proposal.

Yes, Me and Andrew today precisely worked on a better improved version of the same `Conditional Endorsement Triple`

I would let Andrew come up with the First Draft and then we can review it together in the coming meeting!

thomas-fossati:
> > I think this needs work before we commit it. I have a similar proposal.
> 
> Yes, Me and Andrew today precisely worked on a better improved version of the same `Conditional Endorsement Triple`
> 
> I would let Andrew come up with the First Draft and then we can review it together in the coming meeting!

If the only difference is the array around `val` I think this can be trivially accommodated within this PR.  At least I would be happy to.

draft-ietf-rats-corim.md
nedmsmith: > I think this doesn't work when different revisions of CoMID have different definitions of what is in an environment

What doesn't work?

nedmsmith:
> I think this needs work before we commit it. I have a similar proposal.

I don't see the value in having multiple competing PRs. If there isn't consensus for how to solve a problem, then the issue is still in progress and not ready for a PR.

andrew-draper:
I have closed PR #169 - I don't think we have reached agreement on the goals. If we rush for a change before the deadline then we may make mistakes.

draft-ietf-rats-corim.md
andrew-draper: I hadn't realised that this is just a rename of domain-membership-triple-record to group-membership-triple-record. I am not convinced that rename makes the document better.

thomas-fossati: tentatively renamed to "environment bundle"

draft-ietf-rats-corim.md
thomas-fossati: added multiple cardinality to `val`

cddl/mec-endorsement-triple-record.cddl
```diff
@@ -0,0 +1,5 @@
+multi-env-conditional-endorsement-triple-record = [
+  env: environment-map
+  ends: [ + measurement-values-map ]
+  cond: [ + stateful-environment-record ]
+]
```
henkbirkholz, line 5:
```suggestion
multi-env-conditional-endorsement-triple-record = [
  cond: [ + stateful-environment-record ]
  env: environment-map
  ends: [ + measurement-values-map ]
]
]
```

henkbirkholz:
#177 now points to a wiki page illustrating the approach.

PR #170, opened 2023-10-24
Updates

yogeshbdeshpande:


draft-ietf-rats-corim.md
```diff
@@ -1249,6 +1249,25 @@ applies to all measurements in the triple, including those in `measurement-value
...
+The CDDL below, provides a schema that describes the update flow.
+
+~~~ cddl
+{::include cddl/update-record.cddl}
+~~~
+
+#### X-Triples {#sec-x-triple-rec}
+For the Supply Chain Systems where only incremental updates are possible, it is essential that obsolete or security vulnerable Reference Values and/or Endorsed Values must be removed using x-ref-endor-triple-record.
+
+The CCDL below, provides schema for remving Reference Values and/or Endorsed Values from a Verifier.
```
yogeshbdeshpande, line 1264:
```suggestion
The CCDL below, provides a schema for removing Reference Values and/or Endorsed Values from a Verifier.
```

cddl/update-record.cddl
```diff
@@ -0,0 +1,12 @@
+update-triple-record = [
+&(subj: 0) => $subject-type-choice
+& (mval: 1) => [+measurement-values-map] ; object of the update
```
nedmsmith, line 3: Why call it `mval` if it means "object"?

cddl/update-record.cddl
```diff
@@ -0,0 +1,12 @@
...
+? &(authorized-by: 2) => [ + $crypto-key-type-choice ]
+&(action: 3) => $update-action-type-choice
+]
+
+$subject-type-choice /= [+environment-map] ; One or more environment that comprise the subject
+$subject-type-choice /= [+stateful-environment-record] ;  One or more stateful-environments that comprise the subject
+
+$update-action-type-choice /= add ; The new triple causes a new revision identified by measurement-values-map , to be simply updated/added in Verifier
```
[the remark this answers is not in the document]
yogeshbdeshpande: @nedmsmith : The goal of `$update-action-type-choice /= add` is NOT to remove anything from the Verifier, but to increment.

However, as we are using a `common` `update-triple` for any Reference Value and Endorsed Value we  have two options:

(a) Specify what is that is been updated by specifying the type - RefVal/EndVal/CondEnd pertaining to the Subject

(b) Rely on the Supply Chain to exactly issue the same `triple` (like RefVal Triple) OR an EndVal Triple with the new `&(action: 3) => $update-action-type-choice` and 
indicate whether it is an `ADD` or a `REPLACE`

**Choice:**
Option-a: Verifier matches `subj` and based on the `type` attaches the new `triple` to the same `subj` 

Option - b: is `Neater` but implies modifying every triple where we need an update, to set the action

I would request we focus on the `intent` of the action and then agree whether to go with option `a` or option `b`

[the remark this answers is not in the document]
nedmsmith: >I would request we focus on the intent of the action and then agree whether to go with option a or option b

Maybe this part of the discussion should happen in issue #171 since the conclusion would affect the PR definition.

nedmsmith:
I created issue #194 that makes suggestions for how to approach updates. The x-triples section should probably be moved to a new top level section dealing with update processing. Some of the update processing goals are not unique to x-triples, hence can be described in a way that applies broadly to updates regardless of which update vehicle is used.

cddl/update-record.cddl
henkbirkholz: because that is triple speak? If not, that's a good question.

nedmsmith:
Given it is a non-goal for CoRIM to bias a particular update mechanism / process, we should disposition this PR the same as other approaches by excluding it from the "base" spec. It could be defined behind a profile, or as an extension to the base by a companion I-D.

PR #174, opened 2023-12-01
MEC Endorsements

henkbirkholz:
The MEC-E part of #168 including a new record type for the conditions.

### DIFF
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -1278,6 +1278,8 @@ Notes:
 * In order to give the expected result, the condition must describe the expected context completely.
 * The scope of a single MEC triple encompasses an arbitrary amount of environments across all layers in an Attester.
 
+There are scope-related questions that need to be answered.  ([^tracked-at] https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/176)
+
 ## Extensibility {#sec-extensibility}
 
 The base CORIM schema is described using CDDL {{-cddl}} that can be extended


cddl/triples-map.cddl
```diff
@@ -17,5 +17,9 @@ triples-map = non-empty<{
     [ + conditional-endorsement-series-triple-record ]
   ? &(conditional-endorsement-triples: 9) =>
     [ + conditional-endorsement-triple-record ]
+  ? &(mec-endorsement-triples: 10) =>
+    [ + multi-env-conditional-endorsement-triple-record ]
+  ? &(condition-triples: 11) =
```
thomas-fossati, line 22:
```suggestion
  ? &(condition-triples: 11) =>
```

cddl/mec-endorsement-triple-record.cddl
```diff
@@ -0,0 +1,4 @@
+multi-env-conditional-endorsement-triple-record = [
+  conds: [ + condition-triple-record ]
+  actions: [ + stateful-environment-record ]
```
thomas-fossati, line 3: IIRC we agreed this would be the other way round, i.e.: the condition expressed in terms of a bunch of stateful-environments, the "actions" as EMTs.

cddl/condition-triple-record.cddl
```diff
@@ -0,0 +1,4 @@
+condition-triple-record = [
+  environment-map
+  measurement-map
+]
```
thomas-fossati, line 4: this needs to be explicitly added to the CDDL fragments' list to be visible

draft-ietf-rats-corim.md
```diff
@@ -1186,6 +1194,39 @@ Evidence.
...
+~~~ cddl
+{::include cddl/mec-endorsement-triple-record.cddl}
+~~~
+
+A `multi-env-conditional-endorsement-triple-record` has the following parameters:
+
+* `conds`: all target environments, along with a specific state, that need to match in order for the endorsement(s) to apply
+* `actions`: TODO 
+* `env`: the environment to which the endorsed value (conditionally) applies
+* `ends`: the endorsed value(s) associated with `env`
```
thomas-fossati, line 1214: these two should be dropped

henkbirkholz: addressed in https://github.com/ietf-rats-wg/draft-ietf-rats-corim/pull/174/commits/f553dd364fc1dea0e1663f183555c45e92c96103

cddl/mec-endorsement-triple-record.cddl
henkbirkholz: Maybe. Let's fix that nit in the next call.

cddl/condition-triple-record.cddl
henkbirkholz: so I thought.... actually fixed in https://github.com/ietf-rats-wg/draft-ietf-rats-corim/pull/174/commits/c261262fcadb5c76e62b648a0d030cfeea80d19a

draft-ietf-rats-corim.md
```diff
@@ -1186,6 +1186,32 @@ Evidence.
 {::include cddl/domain-membership-triple-record.cddl}
 ~~~
 
+
+
+#### Multi-Environment Conditional (MEC) Endorsements Triple {#sec-comid-triple-mec-endorsements}
+
+The semantics of the Multi-Environment Conditional (MEC) Endorsements Triple is as follows:
+
+> "IF accepted state matches the `cond` value, THEN `env` is associated with the endorsed value(s) `ends`."
```
andrew-draper:
```suggestion
> "IF accepted state matches all `cond` values, THEN every `endorsements` value is added to the accepted state"
```

thomas-fossati: good catch.  fixed in 6d01bbc

draft-ietf-rats-corim.md
```diff
@@ -1186,6 +1186,32 @@ Evidence.
...
+* `conds`: all target environments, along with a specific state, that need to match `state-triples` entries in the ACS for the endorsement(s) to apply
+* `endorsements`: endorsements that are added to the ACS `state-triples` if all `conds` match.
+
+The order in which MEC Endorsement triples are evaluated is important: different sorting may produce different end-results in the computed ACS.
+
+Therefore, the set of applicable MEC Endorsement triple MUST be topologically sorted based on the criterion that a MEC Endorsement triple is evaluated before another if its Target Environment and Endorsement pair is found in any of the stateful environments of the second triple.
+
+Notes:
+
+* In order to give the expected result, the condition must describe the expected context completely.
```
andrew-draper: This note makes me think about a lot of different questions:
* Is there a case where other conditions might partly describe the expected context?
* What problems occur if the CoRIM author chooses to only partly describe the context?
* Why is this triple different from other triples, for example conditional-endorsement-triple-record?

yogeshbdeshpande: > This note makes me think about a lot of different questions:
> 
> * Is there a case where other conditions might partly describe the expected context?
> * What problems occur if the CoRIM author chooses to only partly describe the context?
> * Why is this triple different from other triples, for example conditional-endorsement-triple-record?
Agree 100% on your last point, we need to make edits to optimise and only put forward one triple, instead of these two which are effectively singular and plural instances of doing the same thing!

thomas-fossati: > This note makes me think about a lot of different questions:
> 
> * Is there a case where other conditions might partly describe the expected context?

hmm, what kind of implicit matching are you thinking of?

> * What problems occur if the CoRIM author chooses to only partly describe the context?

You'd get false positives.

> * Why is this triple different from other triples, for example, conditional-endorsement-triple-record?

This is a superset of conditional-endorsement-triple-record. As such, it makes the other redundant, at a small increase in the serialisation cost.

The "series" one is a bit of a different beast: it does some sort of short-circuited OR, so in terms of condition-matching rules it's substantially different.

draft-ietf-rats-corim.md
```diff
@@ -1249,6 +1254,30 @@ applies to all measurements in the triple, including those in `measurement-value
 {::include cddl/conditional-endorsement-triple-record.cddl}
 ~~~
 
+#### Multi-Environment Conditional (MEC) Endorsement Triple {#sec-comid-triple-mec-endorsement}
+
+The semantics of the Multi-Environment Conditional (MEC) Endorsement Triple is as follows:
+
+> "IF accepted state matches all `conds` values, THEN every `endorsements` value is added to the accepted state"
```
yogeshbdeshpande, line 1261:
```suggestion
> "IF accepted state matches all `conds` values, THEN every  entry in the `endorsements` is added to the accepted state"
```

draft-ietf-rats-corim.md
```diff
@@ -1186,6 +1194,39 @@ Evidence.
...
+All the entries in `cond` MUST match.
+
+The order in which MEC Endorsement triples are evaluated is important: different sorting may produce different end-results in the computed ACS.
+
+Therefore, the set of applicable MEC Endorsement triple MUST be topologically sorted based on the criterion that a MEC Endorsement triple is evaluated before another if its Target Environment and Endorsement pair is found in any of the stateful environments of the second triple.
+
+Notes:
+
+* In order to give the expected result, the condition must describe the expected context completely.
+* The scope of a single MEC triple encompasses an arbitrary amount of environments across all layers in an Attester.
```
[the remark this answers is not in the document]
yogeshbdeshpande: > The implication of a scope that covers all grouping / appraisal contexts is the EMT expressions must have some aspect that is globally unique (at least within the expected scope). EMT scope should be described as part of the EMT construction and not as a footnote to a particular triple construction.

See Issue https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/176

PR #181, opened 2023-12-13
First initial Proposal for adding multiple measurement maps to a Reference Triple

yogeshbdeshpande:
[WIP] Multiple Measurements in a Reference Triple

### DIFF
--- a/cddl/reference-triple-record.cddl
+++ b/cddl/reference-triple-record.cddl
@@ -1,4 +1,4 @@
 reference-triple-record = [
   environment-map
-  measurement-map
+  [ +measurement-map ]
 ]
--- a/cddl/stateful-environment-record.cddl
+++ b/cddl/stateful-environment-record.cddl
@@ -1,5 +1,5 @@
 ; an environment with a set of measurements that must match evidence
 stateful-environment-record = [
   environment-map,
-  measurement-map
+  [ +measurement-map ]
 ]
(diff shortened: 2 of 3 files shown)


yogeshbdeshpande:
I will update examples for the PR to compile properly, once we have evaluated the `impact`

henkbirkholz:
Are we revisiting this topic? If my memory does not fail me, didn't we go for "only one object in the triple" choosing simplicity over conciseness in this case?

yogeshbdeshpande:
> Are we revisiting this topic? If my memory does not fail me, didn't we go for "only one object in the triple" choosing simplicity over conciseness in this case?

`Yes` we are revisiting this topic. I believe this will be beneficial, as many use cases exists today where this change will help the users of the specification.

yogeshbdeshpande:
Closing this PR as the alternative proposal is progressed via discussions on 
https://github.com/ietf-rats-wg/draft-ietf-rats-corim/pull/189

deeglaze:
We're revisiting this as of 2024-08-14 so we can account for single evidence objects that measure multiple things, each with data type representations that fit with the standard codepoints in measurement-values-map.

The mkey use for "subcomponents" I think needs environment-map representation. There are different requirements here. Any inclusion of mkey should come with measurement-map multiplicity. If there's any multiplicity, then every element must have an mkey value, and they must be distinct.

yogeshbdeshpande:
Addressed by PR#257

PR #187, opened 2023-12-18
changed cryptokeys code point to 13

nedmsmith:
code point 12 is already used by concise-evidence spec for spdm indirect.


thomas-fossati (approved):
in hindsight it'd have been better to offset the CoEv registration range

nedmsmith:
> in hindsight it'd have been better to offset the CoEv registration range

Agree.

PR #189, opened 2023-12-20
add integrity registers

### LINKED ISSUE #185: PCR bank measurement type
Some environments (e.g., Arm CCA realms, TPMs) have a bunch of measurements bundled in PCR banks rather than single digests.

Instead of having each technology provide their own custom PCR-bank measurement type, we should define it as a basic, reusable CoMID type.

For example - modelled on the `digest` type:

```cddl
; design A

pcrs = [
  alg:  (int / text)
  slots: [ + bytes / nil ]
]
```

The `nil` alternative is for when the PCR slot is not populated or match is to be skipped.

CoRIM profiles will have to define the exact shape of the bank, i.e., the number of elements in the `slots` array.

thomas-fossati:
Fix #185

### DIFF
--- a/cddl/integrity-registers.cddl
+++ b/cddl/integrity-registers.cddl
@@ -1,5 +1,5 @@
-integrity-register-id = uint / text
+integrity-register-id-type-choice = uint / text
 
 integrity-registers = {
-  + integrity-register-id => [ + digest ]
+  + integrity-register-id-type-choice => [ + digest ]
 }
--- a/cddl/integrity-registers.cddl
+++ b/cddl/integrity-registers.cddl
@@ -1,0 +1,5 @@
+integrity-register-id = uint / text
+
+integrity-registers = {
+  + integrity-register-id => [ + digest ]
+}
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -1120,7 +1120,7 @@ Identifiers are either unsigned integers or text strings and their type matters,
 {::include cddl/integrity-registers.cddl}
 ~~~
 
-All the measured objects in an Integrity Registers map are explicitly named and the order in which they are found on the map is irrelevant.
+All the measured objects in an Integrity Registers map are explicitly named and the order in which they appear in the map is irrelevant.
 Any digests associated with a measured object represent an acceptable state for the object.
 Therefore, if multiple digests are provided, the acceptable state is their cross-product.
 For example, given the following Integrity Registers:
@@ -1152,7 +1152,7 @@ and
 
 are acceptable states.
 
-Integrity Registers can be used to model the PCRs in a TPM or vTPM, in which case the identifier is the register index, as well as other kinds of vendor-specific measured objects.
+Integrity Registers can be used to model the PCRs in a TPM or vTPM, in which case the identifier is the register index, or other kinds of vendor-specific measured objects.
 
 
 ##### Domain Types {#sec-comid-domain-type}
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -1112,13 +1112,48 @@ checks, and trust anchor checks.
 
 ##### Integrity Registers {#sec-comid-integrity-registers}
 
-An Integrity register is a group of one or more named measurements.
-Each measurement has an unique identifier (either an unsigned integer or a string) and one or more associated digests.
-This type can be used to model the PCRs in a TPM or vTPM, in which case the identifier is the register index, as well as other vendor-specific "Integrity Registers".
+An Integrity Registers map groups together one or more measured "objects".
+Each measured object has a unique identifier and one or more associated digests.
+Identifiers are either unsigned integers or text strings and their type matters, e.g., unsigned integer 5 is distinct from the text string "5".
 
 ~~~ cddl
 {::include cddl/integrity-registers.cddl}
 ~~~
+
+All the measured objects in an Integrity Registers map are explicitly named and the order in which they are found on the map is irrelevant.
+Any digests associated with a measured object represent an acceptable state for the object.
+Therefore, if multiple digests are provided, the acceptable state is their cross-product.
+For example, given the following Integrity Registers:
+
+~~~cbor-diag
+{
+  0: [ [ 0, h'00' ] ],
+  1: [ [ 0, h'11' ], [ 1, h'12' ] ]
+}
+~~~
+
+then both
+
+~~~ cbor-diag
+{
+  0: [ 0, h'00' ],
+  1: [ 0, h'11' ]
+}
+~~~
+
+and
+
+~~~cbor-diag
+{
+  0: [ 0, h'00' ],
+  1: [ 1, h'12' ]
+}
+~~~
+
+are acceptable states.
+
+Integrity Registers can be used to model the PCRs in a TPM or vTPM, in which case the identifier is the register index, as well as other kinds of vendor-specific measured objects.
+
 
 ##### Domain Types {#sec-comid-domain-type}
 
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -1774,7 +1774,7 @@ If the Reference value for `measurement-values-map` key 1 is a UINT tagged with
 Accepted Claims Set less than the value in the Reference Value then the
 Reference Value does not match.
 
-##### Comparison for digests entries
+##### Comparison for digests entries {#sec-cmp-digests}
 
 The value stored under `measurement-values-map` key 2,
 or a value tagged with
@@ -1828,6 +1828,13 @@ then the next entry from the Reference Values array is likewise
 compared with the next entry of the Accepted Claims Set array.
 If all entries of the Reference Values array match a corresponding entry in the Accepted Claims Set array, then the `cryptokeys` Reference Value matches.
 Otherwise, `cryptokeys` does not match.
+
+##### Comparison for Integrity Registers {#sec-cmp-integrity-registers}
+
+For each Integrity Register entry in Evidence, the Verifier will use the associated identifier to look up the matching Integrity Register entry in the Reference Value.
+If no entry is found, the Reference Value does not match.
+Instead, if an entry is found, the digest comparison proceeds as defined in {{sec-cmp-digests}}.
+Note that it is not required for all the entries in the Reference Value to be used during matching: Evidence could consist of a "quote" (in TPM parlance) of just a subset of the register space.
 
 ##### Handling of new tags
 
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -952,6 +952,9 @@ The following describes each member of the `measurement-values-map`.
   Appraisal verifies that, for each value in `cryptokeys`, there is a matching Reference Value entry.
 Matching is described in {{sec-cryptokeys-matching}}.
 
+* `integrity-registers` (index 14): A register with one or more banks containing measurements associated with the environment.  Described in {{sec-comid-integrity-registers}}.
+
+
 ###### Version {#sec-comid-version}
 
 A `version-map` contains details about the versioning of a measured
@@ -1105,6 +1108,16 @@ checks, and trust anchor checks.
 
 ~~~ cddl
 {::include cddl/crypto-key-type-choice.cddl}
+~~~
+
+##### Integrity Registers {#sec-comid-integrity-registers}
+
+An Integrity register is a register with one or more banks containing measurements.
+Each measurement has an unique identifier (either an unsigned integer or a string) and one or more digests.
+This type can be used to model the PCRs in a TPM or vTPM, in which case the identifier is the register index, as well as other vendor-specific "Integrity Registers".
+
+~~~ cddl
+{::include cddl/integrity-registers.cddl}
 ~~~
 
 ##### Domain Types {#sec-comid-domain-type}
@@ -1797,7 +1810,7 @@ Reference Value does not match.
 
 > I think this comparison method only works if the entry is at key 4 (because
 there needs to be a mask at key 5). Should we have a Reference Value of this
-which stores [expect-raw-value raw-value-mask] in an array?
+which stores `[expect-raw-value raw-value-mask]` in an array?
 
 [^issue]: Content missing. Tracked at https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/71
 
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -1113,7 +1113,7 @@ checks, and trust anchor checks.
 ##### Integrity Registers {#sec-comid-integrity-registers}
 
 An Integrity register is a register with one or more banks containing measurements.
-Each measurement has an unique identifier (either an unsigned integer or a string) and one or more digests.
+Each measurement has an unique identifier (either an unsigned integer or a string) and one or more associated digests.
 This type can be used to model the PCRs in a TPM or vTPM, in which case the identifier is the register index, as well as other vendor-specific "Integrity Registers".
 
 ~~~ cddl
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -952,7 +952,7 @@ The following describes each member of the `measurement-values-map`.
   Appraisal verifies that, for each value in `cryptokeys`, there is a matching Reference Value entry.
 Matching is described in {{sec-cryptokeys-matching}}.
 
-* `integrity-registers` (index 14): A register with one or more banks containing measurements associated with the environment.  Described in {{sec-comid-integrity-registers}}.
+* `integrity-registers` (index 14): A group of one or more named measurements associated with the environment.  Described in {{sec-comid-integrity-registers}}.
 
 
 ###### Version {#sec-comid-version}
@@ -1112,7 +1112,7 @@ checks, and trust anchor checks.
 
 ##### Integrity Registers {#sec-comid-integrity-registers}
 
-An Integrity register is a register with one or more banks containing measurements.
+An Integrity register is a group of one or more named measurements.
 Each measurement has an unique identifier (either an unsigned integer or a string) and one or more associated digests.
 This type can be used to model the PCRs in a TPM or vTPM, in which case the identifier is the register index, as well as other vendor-specific "Integrity Registers".
 
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -1866,10 +1866,10 @@ Otherwise, `cryptokeys` does not match.
 
 ##### Comparison for Integrity Registers {#sec-cmp-integrity-registers}
 
-For each Integrity Register entry in Evidence, the Verifier will use the associated identifier to look up the matching Integrity Register entry in the Reference Value.
+For each Integrity Register entry in the Reference Value, the Verifier will use the associated identifier to look up the matching Integrity Register entry in Evidence.
 If no entry is found, the Reference Value does not match.
 Instead, if an entry is found, the digest comparison proceeds as defined in {{sec-cmp-digests}}.
-Note that it is not required for all the entries in the Reference Value to be used during matching: Evidence could consist of a "quote" (in TPM parlance) of just a subset of the register space.
+Note that it is not required for all the entries in Evidence to be used during matching: the Reference Value could consist of a "quote" (in TPM parlance) of just a subset of the register space.
 
 ##### Handling of new tags
 
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,5 @@
 source 'https://rubygems.org'
 
 gem 'json_pure'
-gem 'cddl', '>=0.9.1'
+gem 'cddl', ['>=0.9.1', '!=0.10.5' ]
 gem 'cbor-diag', '=0.8.1'
--- a/cddl/corim-frags.mk
+++ b/cddl/corim-frags.mk
@@ -46,6 +46,7 @@ COMID_FRAGS += ueid.cddl
 COMID_FRAGS += uuid.cddl
 COMID_FRAGS += version-map.cddl
 COMID_FRAGS += digest.cddl
+COMID_FRAGS += integrity-registers.cddl
 COMID_FRAGS += concise-swid-tag.cddl
 
 COMID_EXAMPLES := $(wildcard examples/comid-*.diag)
--- a/cddl/examples/comid-integrity-registers.diag
+++ b/cddl/examples/comid-integrity-registers.diag
@@ -33,7 +33,7 @@
                 / hash-value / h'deadbeef'
               ]
             ],
-            5: [
+            "my-ir": [
               [
                 / hash-alg-id / 1, / sha256 /
                 / hash-value / h'50aa341af9cb20a879440e58dd6581c14fa14bccafb75f488259262d6ea3a4d9'
--- a/cddl/examples/comid-integrity-registers.diag
+++ b/cddl/examples/comid-integrity-registers.diag
@@ -1,0 +1,51 @@
+/ concise-mid-tag / {
+  / comid.tag-identity / 1 : {
+    / comid.tag-id / 0 : h'3f06af63a93c11e4979700505690773f'
+  },
+  / comid.entity / 2 : [ {
+    / comid.entity-name / 0 : "ACME Inc.",
+    / comid.reg-id / 1 : 32("https://acme.example"),
+    / comid.role / 2 : [ 0 ] / tag-creator /
+  } ],
+  / comid.triples / 4 : {
+    / comid.reference-triples / 0 : [ [
+      / environment-map / {
+        / comid.class / 0 : {
+          / comid.class-id / 0 :
+            / tagged-uuid-type / 37(
+              h'67b28b6c34cc40a19117ab5b05911e37'
+            ),
+          / comid.vendor / 1 : "ACME Inc.",
+          / comid.model / 2 : "ACME RoadRunner",
+          / comid.layer / 3 : 1
+        }
+      },
+      / measurement-map / {
+        / comid.mval / 1 : {
+          / comid.integrity-registers / 14 : {
+            0: [
+              [
+                / hash-alg-id / 1, / sha256 /
+                / hash-value / h'44aa336af4cb14a879432e53dd6571c7fa9bccafb75f488259262d6ea3a4d91b'
+              ],
+              [
+                / hash-alg-id / "my-alg-id",
+                / hash-value / h'deadbeef'
+              ]
+            ],
+            5: [
+              [
+                / hash-alg-id / 1, / sha256 /
+                / hash-value / h'50aa341af9cb20a879440e58dd6581c14fa14bccafb75f488259262d6ea3a4d9'
+              ],
+              [
+                / hash-alg-id / "my-alg-id",
+                / hash-value / h'fefefafa'
+              ]
+            ]
+          }
+        }
+      }
+    ] ]
+  }
+}
--- a/cddl/measurement-values-map.cddl
+++ b/cddl/measurement-values-map.cddl
@@ -14,5 +14,6 @@ measurement-values-map = non-empty<{
   ? &(uuid: 10) => uuid-type
   ? &(name: 11) => text
   ? &(cryptokeys: 13) => [ + $crypto-key-type-choice ]
+  ? &(integrity-registers: 14) => integrity-registers
   * $$measurement-values-map-extension
 }>


andrew-draper:
Please add a section immediately below "Comparison for cryptokeys entries" which describes how a stateful environment (nee Reference Value) is matched against the ACS.

nedmsmith (approved):

henkbirkholz (approved):

nedmsmith (approved):

yogeshbdeshpande:
I have few concerns on the `way` we express the CDDL in the specification for `Integrity registers`

I will state that so that we can consider this during our meeting tomorrow (10/Jan)

I would prefer we keep the encoding `simple` as below:
-----
$register-id-type-choice /= uint
$register-id-type-choice /= text

integrity-register = {
&(id:0) => register-id-type-choice
&(digests:1) => [+digest]
}

  ? &(version: 0) => version-map
  ? &(svn: 1) => svn-type-choice
  ? &(digests: 2) => [ + digest ]
  ? &(flags: 3) => flags-map
  ? (
      &(raw-value: 4) => $raw-value-type-choice,
      ? &(raw-value-mask: 5) => raw-value-mask-type
    )
  ? &(mac-addr: 6) => mac-addr-type-choice
  ? &(ip-addr: 7) =>  ip-addr-type-choice
  ? &(serial-number: 8) => text
  ? &(ueid: 9) => ueid-type
  ? &(uuid: 10) => uuid-type
  ? &(name: 11) => text
  ? &(cryptokeys: 13) => [ + $crypto-key-type-choice ]
  ? &(integrity-registers: 14) => [+integrity-register] 
  * $$measurement-values-map-extension
}>

------
What are the benefits:
1. The expression allows flexibility in type expansion
2. More clear and simple semantics
3. This is more readable 

Disadvantages:
1. May not be as optimised as the proposed solution in the PR

For the current PR proposal:
The flexibility of encoding a register-id as a compound type.
Even if we express that it will be very hard to understand to the reader!
Hard to comprehend (at least to me for sure!)

I would like to seek feedback on how other authors feel about this and then we decide as a team

thomas-fossati:
> I would like to seek feedback on how other authors feel about this and then we decide as a team

hmm, we already had that discussion in #185.

nedmsmith:
>I would prefer we keep the encoding simple as below:
I don't get what about the PR isn't aligned with what is cited as the "simple" encoding. Isn't this restating what is in the PR?

yogeshbdeshpande:
> > I would prefer we keep the encoding simple as below:
> 
> I don't get what about the PR isn't aligned with what is cited as the "simple" encoding. Isn't this restating what is in the PR?

No it is slightly different, if I understand correctly (@thomas-fossati can correct if this is mis-interpreted).  Please check the examples.
In the proposed PR, the code point is encoded as the Register ID. 

While I am suggesting an explicit allocation of register-id.

thomas-fossati:
> No it is slightly different, if I understand correctly (@thomas-fossati can correct if this is mis-interpreted). Please check the examples. In the proposed PR, the code point is encoded as the Register ID.

There is no semantic difference.

What your proposal does is re-encoding a map as an array of (k, v) tuples.

Echoing Ned, I don't understand how that would result in a clearer, simpler and more readable representation.

henkbirkholz:
> > I would like to seek feedback on how other authors feel about this and then we decide as a team
> 
> hmm, we already had that discussion in #185.

@yogeshbdeshpande, so you are saying that nesting two redundant quantities (1st layer [+integrity-register], and then 2nd layer [+digest]) is more simple and readable? I am having a hard time a.) ignoring the established consensus based on #185 and b.) to see the actual improvement here? Maybe I am missing something obviouse, but I like the changes in this PR and am perplexed by the new suggestion (also using the suggestion feature in the files changed tab would have saved me some times spotting the differences in the CDDL :wink:)

yogeshbdeshpande:
> > > I would like to seek feedback on how other authors feel about this and then we decide as a team
> > 
> > 
> > hmm, we already had that discussion in #185.
> 
> @yogeshbdeshpande, so you are saying that nesting two redundant quantities (1st layer [+integrity-register], and then 2nd layer [+digest]) is more simple and readable? I am having a hard time a.) ignoring the established consensus based on #185 and b.) to see the actual improvement here? Maybe I am missing something obvious, but I like the changes in this PR and am perplexed by the new suggestion (also using the suggestion feature in the files changed tab would have saved me some times spotting the differences in the CDDL 😉)

I do not see this as a two layer thing, why you say that?

It is an array representation with an Id and a Value. My point was that if one need to associate with Composite ID, instead of an int/text, then how this can be manifested in the existing PR ?

henkbirkholz:
Okay. Let me rephrase my question.

What problem do you solve with your suggestion?

henkbirkholz:
> The benefit is that the register-id can be a composite identity without losing clarity!

> My point was that if one need to associate with Composite ID

What issue do think are you fixing?

thomas-fossati:
> It is an array representation with an Id and a Value. My point was that if one need to associate with Composite ID, instead of an int/text, then how this can be manifested in the existing PR ?

OK, we complicate the serialisation to solve a problem that does not exist :-)

thomas-fossati:
> I am not clear, how in the current proposal if we want to associate an ID (example a composite ID like a Class ID), can we do that neatly, is the question?

yes.

yogeshbdeshpande:
> > It is an array representation with an Id and a Value. My point was that if one need to associate with Composite ID, instead of an int/text, then how this can be manifested in the existing PR ?
> 
> OK, we complicate the serialisation to solve a problem that does not exist :-)

Why is the serialisation complex? I do not agree! 

As we are modelling Digests, it is a perfectly valid point to keep in mind future extensibility.  In-fact if we do not think properly, in future it may not be very easily usable, for all the possible use cases!

yogeshbdeshpande:
> > I am not clear, how in the current proposal if we want to associate an ID (example a composite ID like a Class ID), can we do that neatly, is the question?
> 
> yes.

ok, can you please cite some examples, how the encoding will look like in a composite ID ?

thomas-fossati:
> > > I am not clear, how in the current proposal if we want to associate an ID (example a composite ID like a Class ID), can we do that neatly, is the question?
> > 
> > 
> > yes.
> 
> ok, can you please cite some examples, how the encoding will look like in a composite ID ?

In CBOR you can do things like:

```
map-with-map-keys = {
 { 1 => 2 } => [ "pippo", "pluto" ]
}
```

thomas-fossati:
> > > > I am not clear, how in the current proposal if we want to associate an ID (example a composite ID like a Class ID), can we do that neatly, is the question?
> > > 
> > > 
> > > yes.
> > 
> > 
> > ok, can you please cite some examples, how the encoding will look like in a composite ID ?
> 
> In CBOR you can do things like:
> 
> ```
> map-with-map-keys = {
>  { 1 => 2 } => [ "pippo", "pluto" ]
> }
> ```

nedmsmith:
> OK, we complicate the serialization to solve a problem that does not exist :-)

I think I agree with Thomas. If there is a need to associate the digest with a composite ID (e.g., `class-map`) then use of the triple / stateful-environment already does that. In a real-world use case of a TPM, the TPM module is the environment and the array of `integrity-register` is the PCRs. TPM uses a scalar (0-15) to name the PCRs (not a composite identifier). Another possibility is TPM-NV is used for integrity in this case some set of NV indexes will be carved from the NV addresses as integrity registers. These might be numeric or given pet names by the spec that assigns the NV indexes. In which case, the pet name (tstr) or the index id (uint) is sufficient to identify the register.

henkbirkholz:
> I do not see this as a two layer thing, why you say that?

And to address that question: because this seems to introduce yet another quantity of one-or-more. There is already a one-or-more on the digest level. What do you gain from the added array? (other than bringing a mix of register-ids into a measurement-values-map instead of using separate measurement-values-map instances, which to me seems to go against the simple triple layout)

I am also at a loss why we need more "flexibility in type expansion", but that is a separate confusion. Preemptively, I am suggesting that we try to clear anyone's confusion here before trying to fix confusion in calls, okay?

yogeshbdeshpande:
> > I do not see this as a two layer thing, why you say that?
> 
> And to address that question: because this seems to introduce yet another quantity of one-or-more. There is already a one-or-more on the digest level. What do you gain from the added array? (other than bringing a mix of register-ids into a measurement-values-map instead of using separate measurement-values-map instances, which to me seems to go against the simple triple layout)
> 
> I am also at a loss why we need more "flexibility in type expansion", but that is a separate confusion. Preemptively, I am suggesting that we try to clear anyone's confusion here before trying to fix confusion in calls, okay?

I think the PR also proposes a two level array, it is just that it not clearly evident in the CDDL?
We have an array of ingtegrity-registers using the +symbol, I have not suggested anything different

yogeshbdeshpande:
Thank you @thomas-fossati @henkbirkholz and @nedmsmith been patient with me 

In addition to above I see another issue with the changes proposed...

The encoding of the CDDL in the base specification relies on the Profile specific 
settings. Specifically, on line 4, of integrity-registers.cddl an encoder will assign a tag/code point based on
the specific register-id settings. The Receiver cannot natively decide the
code point for digest, unless it is aware of the profile specific information, i.e. which profile to apply?
and to expect what code point?

This is against the principles of Base Schema where always deterministic type setting is expected
for each type based on the well-known code point, given in the specification?

Also I am not sure that the order is guaranteed for each encoding, so not sure how this would work, at the receiver ?

Can anyone help me clarify this aspect?

nedmsmith:
> I find the existing CDDL `terse`

Agree. The CDDL could make better use of optional tagging, e.g., [ + mytag: mycontent ]

nedmsmith:
> I am not sure that the order is guaranteed for each encoding

The intent is that order should not matter. Verifier considers the codepoint value for each entry and finds the matching codepoint in the reference value; if they are the same, then compare the digest array values. It may be ambiguous if the digest array (e.g.: `[ + digest ]`) can be compared byte-by-byte or if the parser walks each array entry. Maybe it's an implementation choice.

nedmsmith:
> each type based on the well-known code point

The goal is universal interoperability of Verifiers for appraisal (only) processing. Since the goal is to represent digests that belong to systems that are not defined by this spec, Verifiers are not expected to do additional processing that, for example, knows about which components are not supposed to be extended into PCR4. As long as the `integrity-register-id` is unique to the `integrity-registers` set, the verifier shouldn't need additional context to do the compare operation.

thomas-fossati:
> The encoding of the CDDL in the base specification relies on the Profile specific settings. Specifically, on line 4, of integrity-registers.cddl an encoder will assign a tag/code point based on the specific register-id settings. The Receiver cannot natively decide the code point for digest, unless it is aware of the profile specific information, i.e. which profile to apply? and to expect what code point?
> 
> This is against the principles of Base Schema where always deterministic type setting is expected for each type based on the well-known code point, given in the specification?

codepoint 14 means "integrity-registers", I don't see anything _"[a]gainst the principles of Base Schema always deterministic type setting is expected for each type based on the well-known code point, given in the specification"_.

Note that this is not different from other objects in our info model (e.g., the recently added "tagged bytes".)

> Also I am not sure that the order is guaranteed for each encoding, so not sure how this would work, at the receiver ?

Integrity registers are explicitly named.  Order, i.e., implicit naming, is irrelevant.

thomas-fossati:
> I think the PR also proposes a two level array,

No, it doesn't.

> it is just that it not clearly evident in the CDDL?

Nope.

The core of your proposal is to replace a representation that uses *one* native type (i.e., a map):

```
a = {
  + $key-type => $val-type
}
```

with one that reimplements the _same_ semantics using a two-level construction that serialises the associative array as a list of tuples:

```
a = [
  + [ k: $key-type, v: $val-type ]
]
```

> We have an array/multiplicity of integrity-registers using the +symbol, I have not suggested anything different, but more clear expression.

Comparing the two constructions above I struggle to understand how the latter would be clearer than the former.

> I find the existing CDDL `terse` as far as understanding is concerned

It's terse exactly because it uses a native (i.e., natural) type for representing the concept.

yogeshbdeshpande:
> > The encoding of the CDDL in the base specification relies on the Profile specific settings. Specifically, on line 4, of integrity-registers.cddl an encoder will assign a tag/code point based on the specific register-id settings. The Receiver cannot natively decide the code point for digest, unless it is aware of the profile specific information, i.e. which profile to apply? and to expect what code point?
> > This is against the principles of Base Schema where always deterministic type setting is expected for each type based on the well-known code point, given in the specification?
> 
> codepoint 14 means "integrity-registers", I don't see anything _"[a]gainst the principles of Base Schema always deterministic type setting is expected for each type based on the well-known code point, given in the specification"_.
> 
> Note that this is not different from other objects in our info model (e.g., the recently added "tagged bytes".)
> 
> > Also I am not sure that the order is guaranteed for each encoding, so not sure how this would work, at the receiver ?
> 
> Integrity registers are explicitly named. Order, i.e., implicit naming, is irrelevant.

I am not referring to Code Point 14, but to the code point of the digests. With the proposed changes in the PR, within the base schema, the encoding for [+digest] is dependent on the specific `profile` setting the [+digest].  With the tagged bytes, we always use a fixed tag (to encode the provided bytes from the user). I am not sure, how can one assume at the receiver side, whether the code point is encoded as an integer or a string (text) unless one has the profile information ?

thomas-fossati:
> [h]ow can one assume at the receiver side, whether the code point is encoded as an integer or a string (text) unless one has the profile information ?

Text string and unsigned integer are CBOR types.

yogeshbdeshpande:
> > I think the PR also proposes a two level array,
> 
> No, it doesn't.
> 
> > it is just that it not clearly evident in the CDDL?
> 
> Nope.
> 
> The core of your proposal is to replace a representation that uses _one_ native type (i.e., a map):
> 
> ```
> a = {
>   + $key-type => $val-type
> }
> ```
> 
> with one that reimplements the _same_ semantics using a two-level construction that serialises the associative array as a list of tuples:
> 
> ```
> a = [
>   + [ k: $key-type, v: $val-type ]
> ]
> ```
> 
> > We have an array/multiplicity of integrity-registers using the +symbol, I have not suggested anything different, but more clear expression.
> 
> Comparing the two constructions above I struggle to understand how the latter would be clearer than the former.
> 
> > I find the existing CDDL `terse` as far as understanding is concerned
> 
> It's terse exactly because it uses a native (i.e., natural) type for representing the concept, which (to me) is a great plus.

I think the correct representation for what I am proposing is like below:

 ```
 a = [
   + { k: $key-type, v: $val-type }
 ]
 ```

yogeshbdeshpande:
If one sees the same elements in CBOR Diag, it is not hugely different:

```
  / measurement-map / {
        / comid.mval / 1 : {
          / comid.integrity-registers / 14 : [
            {
			   /id / 0, 26
			   / digests/, 1: [
               [
                / hash-alg-id / 1, / sha256 /
                / hash-value / h'44aa336af4cb14a879432e53dd6571c7fa9bccafb75f488259262d6ea3a4d91b'
              ],
              [
                / hash-alg-id / "my-alg-id",
                / hash-value / h'deadbeef'
              ]
			]
            },
            {
			   /id / 0, 2
			   / digests/, 1: [
               [
                / hash-alg-id / 1, / sha256 /
                / hash-value / h'50aa341af9cb20a879440e58dd6581c14fa14bccafb75f488259262d6ea3a4d9'
              ],
              [
                / hash-alg-id / "my-alg-id",
                / hash-value / h'fefefafa'
              ]
			]
            }
          ]
        }
      }
```

draft-ietf-rats-corim.md
```diff
@@ -952,6 +952,9 @@ The following describes each member of the `measurement-values-map`.
   Appraisal verifies that, for each value in `cryptokeys`, there is a matching Reference Value entry.
 Matching is described in {{sec-cryptokeys-matching}}.
 
+* `integrity-registers` (index 14): A register with one or more banks containing measurements associated with the environment.  Described in {{sec-comid-integrity-registers}}.
```
yogeshbdeshpande, line 955:
```suggestion
* `integrity-registers` (index 14):  Multiple registers with one or more banks containing measurements associated with the environment.  Described in {{sec-comid-integrity-registers}}.
```

thomas-fossati:
> If one sees the same elements in CBOR Diag, it is not hugely different:
> 
> ```
> [...]
> ```

You have not managed to convince us that your last-minute change proposal is an improvement over what this PR (and the associated issue) put forward.

I think that continuing this discussion to the last reply is not a productive use of our time and energy.

yogeshbdeshpande:
> It's terse exactly because it uses a native (i.e., natural) type for representing the concept, which (to me) is a great plus.

The `terse` CDDL is making me nervous and I would prefer to go with more clarity rather than optimising the CDDL.

nedmsmith:
> The `terse` CDDL

The diag file is helpful in visualizing the list of digests. Although, it would help to include a tstr codepoint example as in "myIR" : [ ... ]

thomas-fossati:
> The diag file is helpful in visualizing the list of digests. Although, it would help to include a tstr codepoint example as in "myIR" : [ ... ]

Good point.  Done in [43c73dc](https://github.com/ietf-rats-wg/draft-ietf-rats-corim/pull/189/commits/43c73dc84a80cb117e089a2570885543975b2edd)

> This thread is turning into style differences which usually isn't productive.

Massive 👍

yogeshbdeshpande:
> > > The diag file is helpful in visualizing the list of digests. Although, it would help to include a tstr codepoint example as in "myIR" : [ ... ]
> > 
> > 
> > Good point. Done in [43c73dc](https://github.com/ietf-rats-wg/draft-ietf-rats-corim/pull/189/commits/43c73dc84a80cb117e089a2570885543975b2edd)
> > > This thread is turning into style differences which usually isn't productive.
> > 
> > 
> > Massive 👍
> 
> This is super confusing, I guess, in one of the payload (CoMID ref-measurement-value) we can only have one type of Register ID? How can we mix `uint` and `text` in the same diag?

Sure in that case, I would produce two diags one  exemplifying the usage of  `text` and the other `uint`.

I would not prefer to produce wrong examples, if mixing is not applicable

thomas-fossati:
> I would not prefer to produce wrong examples, if mixing is not applicable

Wrong example?  Where do we state that the keys have to be homogeneous?

yogeshbdeshpande:
> > I would not prefer to produce wrong examples, if mixing is not applicable
> 
> Wrong example? Where do we state that the keys have to be homogeneous?

I am yet to grapple with the fact both type of encoding in the same measurement-value-map is a possibility for a same `profile`

cddl/integrity-registers.cddl
```diff
@@ -0,0 +1,5 @@
+integrity-register-id = uint / text
```
yogeshbdeshpande, line 1: Should not this be 
integrity-register-id-type-choice as per the convention:
https://ietf-rats-wg.github.io/draft-ietf-rats-corim/draft-ietf-rats-corim.html#name-cddl-typographical-conventi

thomas-fossati: done in [425d5f4](https://github.com/ietf-rats-wg/draft-ietf-rats-corim/pull/189/commits/425d5f4189b32cb0876c315abb2912468aac041e)

draft-ietf-rats-corim.md
```diff
@@ -1107,6 +1110,16 @@ checks, and trust anchor checks.
 {::include cddl/crypto-key-type-choice.cddl}
 ~~~
 
+##### Integrity Registers {#sec-comid-integrity-registers}
+
+An Integrity register is a group of one or more named measurements.
```
yogeshbdeshpande, line 1115:
```suggestion
Integrity registers is a group of one or more named measurements.
```

draft-ietf-rats-corim.md
```diff
@@ -1107,6 +1110,16 @@ checks, and trust anchor checks.
...
+Each measurement has an unique identifier (either an unsigned integer or a string) and one or more associated digests.
+This type can be used to model the PCRs in a TPM or vTPM, in which case the identifier is the register index, as well as other vendor-specific "Integrity Registers".
```
yogeshbdeshpande, line 1117:
```suggestion
Integrity registers can be used to model the PCRs in a TPM or vTPM ( in which case the identifier is the register index), or vendor-specific "Integrity Registers".
```

yogeshbdeshpande, line 1117: It is not clear from line 117 in the change whether we are using `This type` to refer to the identifier or Integrity register specifically. Hence I tried to word smith it to add better clarity, however feel free to improve it further, if need be!

draft-ietf-rats-corim.md
```diff
@@ -1816,6 +1829,13 @@ compared with the next entry of the Accepted Claims Set array.
 If all entries of the Reference Values array match a corresponding entry in the Accepted Claims Set array, then the `cryptokeys` Reference Value matches.
 Otherwise, `cryptokeys` does not match.
 
+##### Comparison for Integrity Registers {#sec-cmp-integrity-registers}
+
+For each Integrity Register entry in Evidence, the Verifier will use the associated identifier to look up the matching Integrity Register entry in the Reference Value.
```
yogeshbdeshpande, line 1834:
```suggestion
For each Integrity Register entry in Evidence, the Verifier will use the associated identifier to look up the matching Integrity Register entry in the Reference Value. It is expected that the `type` of identifier in the Evidence is same as the one set in the Reference Values, as dictated by the profile.  Otherwise the comparison is undefined behaviour.
```

yogeshbdeshpande (commented):
Some more clarifications

draft-ietf-rats-corim.md
nedmsmith: Somewhere in this section it might make sense to describe verifier behavior. I believe it is somewhat like ASN.1 SETS (vs. SEQUENCE) in that an entry "PCR5" at position 0 in the list should match a reference value even if the reference placed "PCR5" at position n>0. 

Maybe it goes without saying that verifier will not do type translation. That is if evidence has an entry `'5' : [...]` and reference values has an entry `5 : [...]` there is not ascii-to-numeric / numeric-to-ascii translation. What do you think?

draft-ietf-rats-corim.md
thomas-fossati: reworded in [0daa574](https://github.com/ietf-rats-wg/draft-ietf-rats-corim/pull/189/commits/0daa57453d485e1ad42dd2d338868d0b3d389fcd)

draft-ietf-rats-corim.md
thomas-fossati: reworded in [0daa574](https://github.com/ietf-rats-wg/draft-ietf-rats-corim/pull/189/commits/0daa57453d485e1ad42dd2d338868d0b3d389fcd)

thomas-fossati: reworded in [0daa574](https://github.com/ietf-rats-wg/draft-ietf-rats-corim/pull/189/commits/0daa57453d485e1ad42dd2d338868d0b3d389fcd)

draft-ietf-rats-corim.md
```diff
@@ -1107,6 +1110,16 @@ checks, and trust anchor checks.
 {::include cddl/crypto-key-type-choice.cddl}
 ~~~
 
+##### Integrity Registers {#sec-comid-integrity-registers}
+
+An Integrity register is a register with one or more banks containing measurements.
+Each measurement has an unique identifier (either an unsigned integer or a string) and one or more associated digests.
```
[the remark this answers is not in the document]
thomas-fossati: reworded in [0daa574](https://github.com/ietf-rats-wg/draft-ietf-rats-corim/pull/189/commits/0daa57453d485e1ad42dd2d338868d0b3d389fcd)

nedmsmith (approved):
lgtm

yogeshbdeshpande (approved):
LGTM!

thomas-fossati:
hi @andrew-draper, you had comments at the very beginning, which I think I have addressed.  Could you have a look and confirm you are OK with the current wording?

draft-ietf-rats-corim.md
```diff
@@ -1816,6 +1864,13 @@ compared with the next entry of the Accepted Claims Set array.
 If all entries of the Reference Values array match a corresponding entry in the Accepted Claims Set array, then the `cryptokeys` Reference Value matches.
 Otherwise, `cryptokeys` does not match.
 
+##### Comparison for Integrity Registers {#sec-cmp-integrity-registers}
+
+For each Integrity Register entry in Evidence, the Verifier will use the associated identifier to look up the matching Integrity Register entry in the Reference Value.
+If no entry is found, the Reference Value does not match.
+Instead, if an entry is found, the digest comparison proceeds as defined in {{sec-cmp-digests}}.
+Note that it is not required for all the entries in the Reference Value to be used during matching: Evidence could consist of a "quote" (in TPM parlance) of just a subset of the register space.
```
andrew-draper, line 1872:
```suggestion
Note that it is not required for all the entries in Evidence to be used during matching: the Reference Value could consist of a "quote" (in TPM parlance) of just a subset of the register space.
```

draft-ietf-rats-corim.md
```diff
@@ -1816,6 +1864,13 @@ compared with the next entry of the Accepted Claims Set array.
 If all entries of the Reference Values array match a corresponding entry in the Accepted Claims Set array, then the `cryptokeys` Reference Value matches.
 Otherwise, `cryptokeys` does not match.
 
+##### Comparison for Integrity Registers {#sec-cmp-integrity-registers}
```
andrew-draper, line 1867: For other types of measurement, the reference value chooses which values to compare. Can this comparison work the same way?
Diffs below suggesting what the text would look like if this suggestion is adopted.

thomas-fossati: > For other types of measurement, the reference value chooses which values to compare. Can this comparison work the same way?

I guess the crucial point is whether we want the integrity registers to support "quote" semantics a la TPM, i.e., we allow evidence to be a user-defined subset of the complete state of the attester's IRs.  If so, evidence dictates the direction of matching.  If not, i.e., if we can assume that the quote is always _complete_, then we can flip the direction.

nedmsmith: The question is whether (a) the IR array should be considered an atomic measurement or (b) should each array member be atomic? If (b), then an IR as evidence could contain a superset of members contained in an RV array. The verifier should apply subset matching. If (a), then the matching semantic should be set equivalence. I don't feel strongly either way. If (a), then a second IR measurement can be used in evidence. However, if (b), the verifier would separate the IR into multiple IRs, and keep track of authority differences.

thomas-fossati: Going back to our F2F discussion: *if* it can be assumed that a relying party cannot request a quote from the attester *independently* of the verifier - i.e., the RP cannot decide autonomously the geometry of the quote - then I am OK with @andrew-draper proposed text.

draft-ietf-rats-corim.md
```diff
@@ -1816,6 +1864,13 @@ compared with the next entry of the Accepted Claims Set array.
...
+
+For each Integrity Register entry in Evidence, the Verifier will use the associated identifier to look up the matching Integrity Register entry in the Reference Value.
```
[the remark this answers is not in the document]
deeglaze: Given the extensible semantics of the word "matching", I would recommend referencing {{sec-cmp-digests}} here instead of 2 sentences from now for the definition of "matching [...] entry".

For each Integrity Register entry `id => digests_{ref}` in the Reference Value, the Verifier will claim a "match" with Evidence if the Evidence's integrity register entry `id => digests_{ev}` compares positively with the method defined in {{sec-cmp-digests}}. If `id` is not present in the Evidence, then the Reference Value does not "match".

draft-ietf-rats-corim.md
yogeshbdeshpande: @nedmsmith : We would await a new version of suggested `text` from your side and then we can proceed here!

draft-ietf-rats-corim.md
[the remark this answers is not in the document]
nedmsmith:
```suggestion
When asserted as Evidence, the Verifier will accept each Integrity Register entry into the ACS.
When asserted as Reference Values, all of the Integrity Register entries of the matching EMT MUST have a corresponding Integrity Register entry in the ACS Evidence.
Additionally, the Verifier will use the associated Integrity Register label (i.e.,`integrity-register-id-type-choice`) to match the label in a corresponding Integrity Register entry in the ACS.
The Verifier compares the labels which MUST exactly match.
If the labels match, then the Integrity Register digest arrays are compared (see {{sec-comid-integrity-registers}} and {{sec-comparison-for-digests-entries}}).
```
Note: {#sec-comparison-for-digets-entries} needs to be added.

nedmsmith (commented):
See suggested texts

draft-ietf-rats-corim.md
yogeshbdeshpande: I am ok, with the suggested text from Ned, as it makes the matching constraints more crisp and consistent.

draft-ietf-rats-corim.md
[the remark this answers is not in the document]
yogeshbdeshpande: @nedmsmith : This text is not clear. We should keep it at a high level describing just the Evidence and Reference Value requirements. Bringing EMT and ACS into mix has made it more complicated, given that we are towards simplification, so this is too much detail without adding much value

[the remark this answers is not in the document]
yogeshbdeshpande: I am happy with @andrew-draper suggested `text` here. It is also `sort` of aligned with RATS specification: As mentioned in * Section  8.3 of RFC 9334
`
8.3.  Reference Values

 Reference Values used in appraisal procedures come from a Reference
 Value Provider and are then used by the Verifier to compare to
 Evidence.  Reference Values with matching Evidence produce acceptable
 Claims.  Additionally, an appraisal policy may play a role in
 determining the acceptance of Claims.
`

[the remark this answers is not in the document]
nedmsmith: > We should keep it at a high level

We already document the high level rules in earlier sections. It doesn't add value to repeat it here. The reader is potentially uncertain how a verifier will process the array of measurements as each measurement could be processed independently. Alternatively, the collection of measurements could be processed as a singleton measurement. The current text provides no insight or guidance that resolves the reader's confusion.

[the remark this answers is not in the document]
yogeshbdeshpande: I quite liked the crisp nature of text @deeglaze added, so tried to modify her text slightly to specify more clarity on 
matching semantics and attempted purely to see how this expresses. See https://github.com/ietf-rats-wg/draft-ietf-rats-corim/pull/198/files

draft-ietf-rats-corim.md
yogeshbdeshpande, line 1867:
```suggestion
##### Comparison for Integrity Registers {#sec-cmp-integrity-registers}
For each Integrity Register entry in the Reference Value, identified as: id => digests_{ref}, the Verifier will claim a "match" with Evidence only if the Evidence has an Integrity Register entry, identified as: id => digests_{ev} such that the `id` in Evidence and Reference Value both have the same type and value and the digests matches using the method defined in {{sec-cmp-digests}}.

If no entry is found, the Reference Value does not match.

Note that it is not required for all the Evidence entries in an Integrity Register array to be used during matching: Evidence could consist of a Target Environment's complete register space, while the Integrity Register array in Reference Values could consist of a subset of the Target Environment's register space.
```

draft-ietf-rats-corim.md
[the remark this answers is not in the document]
yogeshbdeshpande: > > We should keep it at a high level
> 
> We already document the high level rules in earlier sections. It doesn't add value to repeat it here. The reader is potentially uncertain how a verifier will process the array of measurements as each measurement could be processed independently. Alternatively, the collection of measurements could be processed as a singleton measurement. The current text provides no insight or guidance that resolves the reader's confusion.

@nedmsmith : I do not think, we have so far introduced any section as to how `Contents` within Integrity Registers (within Evidence and Reference Values) are to be compared. All I was stating was that: We should not describe the loading of Evidence and  RefValue into ACS specific actions here  and just describe the direct matching semantics of the Registers.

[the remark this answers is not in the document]
nedmsmith: > We should not describe the loading of Evidence and RefValue into ACS

Line 1869 of my proposed edit could be removed based on this objective. Lines 1870-1873 are aimed at describing the direct matching semantics. Possibly there is better wording, but it seems to say what we want it to say.

draft-ietf-rats-corim.md
nedmsmith: > id => digests_{ev}

This nomenclature is confusing as it doesn't refer to the actual CDDL. Substitute `id` with `integrity-register-id-type-choice` and ` => digests_{ev}` with ` => [ + digest ]` and it would be more clear. Alternatively, include the aliased terms in the CDDL definition. E.g.:
```
integrity-registers = {
  + id: integrity-register-id-type-choice => digests: [ + digest ]
}
```
Then it would be OK to use `id` and `digests`.

yogeshbdeshpande: > > the Verifier will claim a "match" with Evidence only if the Evidence has an Integrity Register entry
> 
> This wording is confusing because "Claim" has special meaning but it is used in the verb form. Other parts of the document use "match" in the verb form already. Maybe it's better to stay with this convention.
> 
> Nevertheless, the matching semantic is misrepresented by the sentence. The Verifier doesn't match evidence per se, as it accepts it simply because the (authenticated) Attester asserted it.
> 
> The Ref Val is accepted only if it is a subset of the Evidence.
> 
> The question that needs to be answered is whether or not the Ref Val subset requirements are applied to the IR array too. I think the answer should be yes.
> 
> The text I supplied seems to work in terms of describing applying the subset matching semantics to the IR array.

I agree we should use the term `match` only not mis-use the word claim.

nedmsmith: +1

henkbirkholz: TL;DR the RP cannot decide autonomously the geometry of the quote and Andy's text is fine

I've been raising the question to the TCG IWG. There was a clear answer: the RP does not meddle with the geometry of the quote. The only usage scenario where that implicitly could be the case is where Attester owner and RP owner are collapsed roles and the local configuration of an Attester is required by certain RPs. In these cases the RP owner (which is also the Attester Owner) would also take on the role of Endorser and produce a CoRIM for the Verifier to take care of "the RP's needs". So, in the end the requirement would be again be addressed by the de-facto Endorser and not the RP.

thomas-fossati: > The question that needs to be answered is whether or not the Ref Val subset requirements are applied to the IR array too. I think the answer should be yes.

Given @henkbirkholz https://github.com/ietf-rats-wg/draft-ietf-rats-corim/pull/189#discussion_r1475724199 I am OK too.

> The text I supplied seems to work in terms of describing applying the subset matching semantics to the IR array.

LGTM

thomas-fossati: I suggest we accept Andy's proposed text and merge this.  As suggested by Yogesh, we can track the refinement put forward by Ned in a separate PR.

yogeshbdeshpande: This works with me +1

yogeshbdeshpande: @nedmsmith : We have raised a new issue #199 to address your comments. We can discuss this in the incoming meeting. In the mean time, we intend to apply suggestions from @andrew-draper after clarification from @nedmsmith and acceptance from @thomas-fossati. I hope this is ok, with you.

Please reply with +1 if you agree so that we can un-block it today.

nedmsmith: +1

nedmsmith (approved):

PR #198, opened 2024-02-01
Corim integrity register new text

yogeshbdeshpande:


nedmsmith:
I'm confused why there are competing PRs. Can we decide to use #189 or #198 but not both?

yogeshbdeshpande:
> I'm confused why there are competing PRs. Can we decide to use #189 or #198 but not both?

No intention to confuse anybody. This was just a draft for you to check how the text aligns. I will delete #198 immediately.

Better I add a suggestion in the main PR. But I wanted to avoid too many commennts, hence thought of aligning this way

yogeshbdeshpande:
Will be addressed by #189 ONLY hence deleting this!

Issue #197, opened 2024-01-31
CoRIM simplification structure


deeglaze:
We discussed in today's meeting that next time we will go over a PR or slides about simplification to CoRIM as a standard to a Core and some other bits.

Here's what I think ought to be addressed:

1. CoRIM is a closed CDDL without extension points.
2. All extensibility of representation and interpretation is relegated to the CoRIM profile.
3. Profiles for standard attestation technologies may choose to go through IETF standardization to allocate CBOR tags through IANA, but this is not required for profiles.
4. The CoRIM working group may define reference value types with "matching" semantics that can be reused by other profile definitions. This should help extend the composability of the initially standardized notions of reference values. The policy should be, but still at the WG's discretion, that reference values are not given privileged CBOR representation through tag allocation without being applicable to at least 2 standard profiles. The applicability of the generalization must be concrete to give clear motivation behind the representation and semantics.
5. The ACS discussion should not use CDDL because it confuses the conceptual with the serializable. It should use an ABNF to describe an algebra of claims and sets of claims. All reference value and endorsement types must have a defined operation to reflect into the algebra. Claims can be partially ordered in order to define simplification through subsumption. A set of claims must therefore contain pairwise incomparable claims. By defining this algebra, the standard may give a clear definition to word "match" for composition and reuse, but it does not limit what profiles are definable–it is merely scaffolding. The algebraic notion of Claim is extensible by any profile, but each type of Claim is defined and owned by the standard/profile that defines it.
6. A CoRIM either fails to "match" an accepted claim set, or it augments individual claims in the ACS with "authorized-by" tags.
7. Evidence Appraisal should be defined explicitly as an operation to translate an octet string into a claim set.

For all of this, we've boiled the representation down to profile * profile-defined octet string. The algebra is a communication tool for specifying the semantics for composable standardized profile components.

All of the schema that is specified in the current document can be interpreted as, "if your profile's representation structurally matches profiles defined in RFCs, then it must semantically mean the same". You can keep TPM, PSA, and DICE following the same schema if you'd like as a way to provide a baseline for expected expressiveness.

[...]

PR #201, opened 2024-02-05
Theory of operation

nedmsmith:
Added section 5.2 describing the verifier theory of operation


draft-ietf-rats-corim.md
```diff
@@ -1434,34 +1434,137 @@ they are not required to use the same internal data structures.
...
+
+If an augmentation tuple's condition is not met, then the tuple is placed on an input queue to be re-tried after one or more other inputs have been applied.
+
+Every augmentation tuple has a condition appearing to the right of the ACS entry (see {{acs}}) that points to a search scope that is a range of ACS entries that are searched by a given condition.
+
+The conceptual message type determines ACS search scope.
+Evidence condition search scope is the header ACS entry.
+Reference Values condition search scope is any Evidence entry.
+
+However, technically Evidience Claimset can contain more Claims than are contained in the Reference Values Claimset.
```
deeglaze, line 1551: sp

nedmsmith:
>if we assume that all claims are conjunctive, then the problem becomes HORNSAT

The conditional series triple is possibly the most complex, which is a 2-step condition, but both are conjunctive. The idea of an <update> that is a function could become problematic if the function somehow introduces negation or disjunction. But the anticipated use with identity-key and attest-key triples doesn't do this. At most, the function result is a conjunction that adds new claims to ACS. Most likely it produces a <nil> result.

nedmsmith (commented):
LGTM

draft-ietf-rats-corim.md
```diff
@@ -1434,34 +1434,137 @@ they are not required to use the same internal data structures.
...
 this document uses abstract representation of Verifier internals.
 
 * Claim: A piece of information, in the form of a key-value pair.
-* Environment Measurement Tuple (EMT): A structure containing a set of environment
-Claims that describe a Target Environment and a set of measurement Claims that
-describe attributes of the Target Environment.
-* reference state: Claims that describe various alternative states of a Target Environment.
+* Claimset: A structure containing a set of Claims that describe a Target Environment and its associated measurement Claims.
+* Conceptual Message Type: An indication of the type of conceptual message that contains Claimsets.
+* Reference state: Claims that describe various alternative states of a Target Environment.
```
deeglaze, line 1446: I'm a little stuck on the theory of operation for how we'll be able to practically sign our ACPI table data. Given that MMIO addresses are a part of it, and there's a good amount of variability across VM shapes

From what I understand, since I'm not an expert in ACPI, we build asl files offline to a particular shape of machine. We don't want to rewrite asl files and run a compiler at VM creation time. We treat these asl binaries not as all the tables, but as table templates. At VM creation time, we use VM instance creation request data to rewrite very specific areas of the built templates as a way to configure that specific VM.

The EV_POST_CODE event that is measured into the TPM or CCEL is a digest of the tables after this template-substitution operation. We thus don't have a reference value that "matches" the digests in the measurement-values-map key 2 or tagged digest value comparison algorithm.

What we _can_ do is require that the lead attester send the acpidump or specially-trimmed dump to find just the substituted positions for the verifier to plug into the template and compute the digest that one would expect for the ACPI EV_POST_CODE event.

If TCG were to have a profile for how to appraise TPM 2.0 quotes with their event logs, I don't imagine they'd have something this specific. I can't revise the official profile with a different profile for this supplemental ACPI table evidence. It seems like I'd be able to have "authorized-by" tag this extra ACPI table evidence, but not the TPM evidence.

deeglaze: I've discussed this more with the engineer working on ACPI tables, and we may be able to pre-generate and store the digests but not the tables since the combinatorial explosion of UART count, CPU count, and PCI root bus count is prohibitive space-wise when copying whole tables. This does get into the many allowable digests problem that Andy has been trying to solve though.

nedmsmith: The "tables of measurements" approach seems to resonate with what Chris Fenner has described in other forums. The basic approach for working with table of measurements is the `environment-map` refers to the table of measurements and the `measurement-values-map` contains a `digests` or `integrity-register` with the digest of the table of measurements. 

The table of measurements seems to be something that is outside the spec however.  Though, it seems possible to use `integrity-register` as a way to represent the table of measurements in a reference value - at least at a high level it seems possible.

nedmsmith: Another possibility to consider is the CCEL digest (which is dynamically generated on the Attester - if I'm not mistaken) could be reported as an `instance-id` in the `environment-map` of the evidence. This allows reference values that relate to measurments that are in a reference manifest to be matched without getting hung up on the CCEL digest not having a reference value. (I'm not close enough to the use case to know if the CCEL digest has security properties however).

deeglaze: I'm unfamiliar with Chris Fenner's hashed table of measurements approach. Do you have a link?

I'm still unclear how profiles for interpreting evidence into an ACS are meant to match up with profiles for reference values. It seems like there is no such thing as a profile for reference values, and that you have to use the same profile as the evidence, which leads to this problem. "A profile allows the base CoRIM schema to be customised to fit a specific Attester", not "A profile allows the base CoRIM CDDL to be customised to fit a specific Attester, Reference Value Provider, or Endorser".

If I'm understanding you correctly, an attester should only use class-id and optionally group in an environment to not conflict with an RVP's use of `instance` to delineate which of the allowed sets of measurements may match a particular family of target environments. That sounds like you still need the Attester profile to account for all the ways that an RVP wants to allow for variability in its CoRIMs, correct?

Specific PCR bank subsets correspond to boot layers kinda, so PCR 0 is the launch measurement, 1 and 7 (RTMR 0) for firmware, PCRs 2 and 6 in RTMR 1 for UEFI driver and host platform-specific measurements, PCRs 8-15 in RTMR 2 for static OS. PCRs 3,4, and 5 are chopped liver, even though GRUB2 is measured into PCR 4 and GRUB2. In practice 8, 9 are used by GRUB2 to measure the kernel and kernel cmdline, so 10-15 are more for Linux IMA. If the kernel build comes from Canonical and the kernel modules come from NVIDIA, and other sources, then we have a very hard time combining expected reference values into signed RVs due to the way NVIDIA has chosen to architect their attestation verification service, specifically for driver reference values. You can't easily have an indexed event log signed by a single principle, even if the boot is sliced and diced to different environments for different sets of integrity registers.

Ideally one would like the event log composition to be analyzed by the Verifier against a policy of which specific events are allowed, so that the combined replay into the PCR bank in the quote is acceptable. Again we have multiple approvers for pieces of a single measurement, just like in Issue#202. If the final value of an integrity register has to be signed by a single principle, then we're going to get into the habit of the Verifier itself signing these values as a composite measurement approver. Is that the expected outcome?

nedmsmith: >It seems like there is no such thing as a profile for reference values, and that you have to use the same profile as the evidence, which leads to this problem. "A profile allows the base CoRIM schema to be customised to fit a specific Attester", not "A profile allows the base CoRIM CDDL to be customised to fit a specific Attester, Reference Value Provider, or Endorser".

CoRIM defines a profile, see `corim-map`. Likewise, evidence can have a profile too (see EAT `profile-label` and `spdm-toc-map` in https://trustedcomputinggroup.org/wp-content/uploads/TCG-DICE-Concise-Evidence-Binding-for-SPDM-Version-1.0-Revision-54_pub.pdf. 

There could be a single profile for all of them with sections that are specific to a conceptual message or one profile could be a master profile that describes which other profiles are used for say evidence. If the CoRIM profile is selected as the master profile, then the statement "...CoRIM CDDL to be customized to fit a specific Attester, Reference Value Provider, or Endorser" could be reasonable.

nedmsmith: >If the final value of an integrity register has to be signed by a single principle, then we're going to get into the habit of the Verifier itself signing these values as a composite measurement approver. Is that the expected outcome?

Given the use case uses multiple verifiers (say one at each layer of the boot sequence), then each layer can be evaluated according to the `integrity-registers` measurement for that layer. 
If each layer verifier uses a different key to assert the results, and given the results are asserted as a digest back into an `integrity-registers` structure. Then each layer is a principal that signs a partial attestation result (for each layer). A downstream / remote verifier can verify the intermediate layer results and arrive at an overall result (according to what the RP / use case expects).

nedmsmith: That's essentially what I'm suggesting. 
Given this longish thread, are there wording changes you would like to see in the above proposed text?

deeglaze: Remove the term until it's used somewhere. There are no other uses of the term "reference state" and the definition that it means "various alternative states of a Target Environment" doesn't match up with what I understand what a CoRIM can represent. Are all these varieties only expressible within separate tags? The term is hard to interpret on its own.

thomas-fossati:
superseded by #197

Issue #154, opened 2023-09-27
Assign CBOR tag for digest processing rules


andrew-draper:

nedmsmith:
The issue should include the use case. It seems the PR #203 points to this issue for context, but none is provided.

PR #203, opened 2024-02-12
Assign tag for digests

andrew-draper:
Issue #154 notes that there is no tag assigned for a digests measurement when it appears at a local codepoint (for example at measurement-values-map codepoint -42).

This diff allocates the next available codepoint to inform the verifier that it should use the digests type.

### DIFF
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -299,6 +299,8 @@ should only be used when the `digest` type conveys reference value
 measurements that are matched verbatim with Evidence that uses the same
 convention - e.g., {{Section 4.4.1.5 of -psa-token}}).
 
+The `digests-type-choice` measurement is either an untagged array of `digest` or the same array tagged with the CBOR tag `#6.562`.
+
 ~~~ cddl
 {::include cddl/digest.cddl}
 ~~~
@@ -1813,9 +1815,7 @@ Reference Value does not match.
 
 ##### Comparison for digests entries {#sec-cmp-digests}
 
-The value stored under `measurement-values-map` key 2,
-or a value tagged with
-#6.TBD is a digest entry.
+The value stored under `measurement-values-map` key 2, or a value tagged with #6.562 is a digest entry.
 It contains one or more digests, each measuring the
 same object. A Reference Value may contain multiple digests, each with a
 different algorithm acceptable to the Reference Value provider. If the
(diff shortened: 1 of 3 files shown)


yogeshbdeshpande (approved):
LGTM!

draft-ietf-rats-corim.md
```diff
@@ -299,6 +299,8 @@ should only be used when the `digest` type conveys reference value
 measurements that are matched verbatim with Evidence that uses the same
 convention - e.g., {{Section 4.4.1.5 of -psa-token}}).
 
+The `digests-type-choice` measurement is either an untagged array of `digest` or the same array tagged with the CBOR tag `#6.562`.
```
nedmsmith, line 302:
```suggestion
The `digests-type-choice` measurement is either an untagged array of `digest` or the same array tagged with the CBOR tag `#6.562`.
If a usage of `digests` differs from the semantics described for `#6.562`, then a different CBOR tag value MUST be used that defines the different semantics such that verifiers can agree to correct processing behavior.
```

thomas-fossati (changes requested):
I am afraid I don't understand the goal of this PR.
The issue too is quite cryptic.

nedmsmith:
> I am afraid I don't understand the goal of this PR. The issue too is quite cryptic.

There are 3 ways to disambiguate semantics of a right-hand-side (RHS) expression. Profiles, code points (LHS), CBOR tags (RHS). Profiles, code points, and CBOR tags all have vendor-specific alternatives, that if used, require vendor-specific updates to Verifiers. Given the goal of a universal verifier is to avoid vendor-specific updates, standard profiles (e.g., DICE / base CoMID), positive code points, and standard CBOR tags `#6.<0..32767>` must be used.

A while back we argued ourselves out of `[ + measurement-values-map ]` multiplicity. However, if there is a need for multiple measurements of the same type e.g., `digests`. There are 4 ways to do that. 
(1) multiple triple entries where the `environment-map` is repeated for each different `measurement-map`.
(2) use multiple but different code points.
(3) define a new measurement type that includes multiplicity on the RHS i.e., `[ + RHS ]`, e.g., `integrity-registers`. 
(4) add back multiplicity around `measurement-values-map`.

Option (1) seems to create bloat by repeating the `environment-map` multiple times. But verifiers might not preserve ordering, hence could get the entries reversed. So, not a realistic option.
Option (4) still requires positional logic to disambiguate individual instances. Hence, some other form of index value is needed that is recognized by universal verifiers. So, not a realistic option.
Option (3) requires an update to a standard for each new instance of the measurement. Flexibility could be an issue.
Option (2) was discussed a few times and there seemed to be consensus for its use. But non-vendor specific CBOR tags are needed for the RHS to ensure a universal verifier can process multiples. Nevertheless, (unless I'm mistaken) the CDDL that defines the negative code points has to be defined, possibly via a profile? 
But more generally, it could be defined by reserving a few LHS code points that expect `#6.<0..32767>(any)` or maybe `#6.<500..599>(any)` as the RHS value?  E.g.:
```
measurement-values-map = {
    ? < -25..-1 > => #6.<500..599>(any)
}
```

nedmsmith:
The CBOR code point defines a standardized structure such as ‘digests’.

Sent from my iPhone

On Feb 19, 2024, at 4:55 AM, Thomas Fossati ***@***.***> wrote:

﻿

measurement-values-map = {
    ? < -25..-1 > => #6.<500..599>(any)
}


Even so, you'd need a profile to tell the verifier what -25 and 599 mean.

nedmsmith:
Section 1.3.8 contains a description of a singleton digest but also says `A measurement can be obtained using different hash algorithms. A digests-type can be used to collect multiple digest values obtained by applying different hash algorithms on the same input.` Arguably, this statement belongs in section 3.1.4.1.5.2 measurement-values-map where the semantics of code point 2 are described. Because the multiple digests describe the same object, code point 2 is an algorithm agility capability where some implementations may use one hash algorithm while another may use something different. 

If there is another use case for multiple digests, then we should document it and assign a code point for it. 

Additionally, it might make sense for the second usage to require algorithm agility properties, in which case the structure might need to be an array of arrays? (e.g. `integrity-registers`).

Otherwise, this PR becomes a request for an extensibility point on code point 2 which seems nuanced and subtle when taken in the context of all the other possible ways to extend.

thomas-fossati:
@andrew-draper to file an issue discussing extension points and data types, that can be discussed before a new PR is created.

PR #206, opened 2024-02-14
Clarify "standard CoRIM type"

deeglaze:
There's a base schema and there's a standard type, so we ought to define the separate concepts.

### DIFF
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -453,7 +453,6 @@ that MUST have a different identifier.
 {::include cddl/profile-type-choice.cddl}
 ~~~
 
-A "standard CoRIM type" references definitions in {{sec-corim-cddl}} and any further definitions in relevant IANA registries.
 
 ### Entities {#sec-corim-entity}
 
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -453,6 +453,8 @@ that MUST have a different identifier.
 {::include cddl/profile-type-choice.cddl}
 ~~~
 
+A "standard CoRIM type" references definitions in {{sec-corim-cddl}} and any further definitions in relevant IANA registries.
+
 ### Entities {#sec-corim-entity}
 
 The CoRIM Entity is an instantiation of the Entity generic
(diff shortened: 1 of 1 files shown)


draft-ietf-rats-corim.md
```diff
@@ -453,6 +453,8 @@ that MUST have a different identifier.
 {::include cddl/profile-type-choice.cddl}
 ~~~
 
+A "standard CoRIM type" references definitions in {{sec-corim-cddl}} and any further definitions in relevant IANA registries.
```
nedmsmith, line 456:
```suggestion
A "standard CoRIM type" references definitions in {{sec-corim-cddl}} and any further definitions in relevant standards or IANA registries.
```

nedmsmith (commented):
Added relevant standards.

draft-ietf-rats-corim.md
deeglaze: Is this assumed to be IETF standard, or is there a list of expected standards bodies that extensions could come from?

nedmsmith: There could be other I-D from IETF or even other standards organizations could extend. There are some control surfaces in terms of IANA registries for code points and CBOR tagging, but IANA doesn't constrain itself to IETF-only standards. 

I'm not entirely clear on why this kind of wording in needed in the first place however. Is the "standard CoRIM type" terminology used elsewhere in the document?

deeglaze: It's used in the document without definition, so unless you want to replace the current use to something clearer, we should add a definition. There's a "standard CoRIM extension" later referencing CoTS, but that seems like a different notion of extension than the map codepoint extensions.

deeglaze: I would expect that any number in IANA registries would link to the standard document that introduced it and gives it meaning, which is why I just used the IANA registry. That's the one authority that can assign numbers that are used in these extension points, no?

andrew-draper: I think "standard CoRIM type" is referring to the codepoints within `measurement-values-map` but this isn't very clear from the text.

I wonder whether it would be better to say something along the lines of "a profile SHALL not redefine anything in this document, but it can add to it", which is what I think the current text is trying to say.

draft-ietf-rats-corim.md
```diff
@@ -1352,7 +1353,8 @@ CDDL map extension points have the form `($$NAME-extension)` where "NAME" is the
 and '$$' signifies map extensibility. Typically, map extension requires a convention
 for code point naming that avoids code-point reuse.
 Well-known code points may be in a registry, such as CoSWID {{-coswid-reg}}.
-Additionally, a range of code points may be reserved for vendor-specific use such as negative integers.
+Any negative integer may be assigned meaning locally to the profile extending the map type.
```
yogeshbdeshpande, line 1356:
```suggestion
```

yogeshbdeshpande (approved):
LGTM

PR #215, opened 2024-03-07
Clarify `[+digest]` semantic

### LINKED ISSUE #214: Clarify the semantic of `[ + digest ]`
Currently, `digests` and `integrity-registers` allow one or more digest values:
```
  ; ...
  ? &(digests: 2) => [ + digest ]
  ; ...
}>

integrity-registers = {
  + integrity-register-id-type-choice => [ + digest ]
}
```

In both cases, the intended semantics is that each digest in the array represents one possible value for the measured object (i.e., the array encodes "OR" semantics).

Unfortunately, this is never clearly stated ([§3.1.4.1.7](https://www.ietf.org/archive/id/draft-ietf-rats-corim-04.html#section-3.1.4.1.7) and [§3.1.4.1.5.2](https://www.ietf.org/archive/id/draft-ietf-rats-corim-04.html#section-3.1.4.1.5.2) are the relevant sections) and a reader could instead interpret it to represent a multi-measurements object (i.e., using "AND" semantics).

thomas-fossati:
Fix #214

### DIFF
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -1137,6 +1137,7 @@ The digests use `digests-type` semantic ({{sec-common-hash-entry}}).
 ~~~
 
 All the measured objects in an Integrity Registers map are explicitly named and the order in which they appear in the map is irrelevant.
+Any digests associated with a measured object represent an acceptable state for the object.
 Therefore, if multiple digests are provided, the acceptable state is their cross-product.
 For example, given the following Integrity Registers:
 
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -299,8 +299,8 @@ Whenever possible, using the `int` encoding is RECOMMENDED.
 ~~~
 
 An environment can be measured using different hash algorithms.
-A `digests-type` can be used to collect multiple digest values obtained applying different hash algorithms on the same input.
-All entries in the `digests-type` MUST have different `alg` values.
+A `digests-type` can be used to collect multiple digest values obtained by applying different hash algorithms on the same input.
+Each entry in the `digests-type` MUST have a unique `alg` value.
 
 ### Tagged Bytes Type {#sec-common-tagged-bytes}
 
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -298,7 +298,7 @@ Whenever possible, using the `int` encoding is RECOMMENDED.
 {::include cddl/digest.cddl}
 ~~~
 
-An environment can be measured using different hash algorithms.
+A measurement can be obtained using different hash algorithms.
 A `digests-type` can be used to collect multiple digest values obtained by applying different hash algorithms on the same input.
 Each entry in the `digests-type` MUST have a unique `alg` value.
 
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -921,8 +921,9 @@ The following describes each member of the `measurement-values-map`.
   {{sec-comid-svn}}.
 
 * `digests` (index 2): Contains the digest(s) of the measured environment
-  together with the respective hash algorithm used in the process.  See
-  {{sec-common-hash-entry}}.
+  together with the respective hash algorithm used in the process.
+  Each digest in the array represents one acceptable state of the measured object.
+  See {{sec-common-hash-entry}}.
 
 * `flags` (index 3): Describes security relevant operational modes. For
   example, whether the environment is in a debug mode, recovery mode, not fully
@@ -1125,6 +1126,7 @@ checks, and trust anchor checks.
 
 An Integrity Registers map groups together one or more measured "objects".
 Each measured object has a unique identifier and one or more associated digests.
+Each digest represents one acceptable state of the measured object.
 Identifiers are either unsigned integers or text strings and their type matters, e.g., unsigned integer 5 is distinct from the text string "5".
 
 ~~~ cddl
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -1130,7 +1130,7 @@ checks, and trust anchor checks.
 An Integrity Registers map groups together one or more measured "objects".
 Each measured object has a unique identifier and one or more associated digests.
 Identifiers are either unsigned integers or text strings and their type matters, e.g., unsigned integer 5 is distinct from the text string "5".
-The digests use `digests-type` semantic ({{sec-common-hash-entry}}).
+The digests use `digests-type` semantics ({{sec-common-hash-entry}}).
 
 ~~~ cddl
 {::include cddl/integrity-registers.cddl}
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -298,6 +298,10 @@ Whenever possible, using the `int` encoding is RECOMMENDED.
 {::include cddl/digest.cddl}
 ~~~
 
+An environment can be measured using different hash algorithms.
+A `digests-type` can be used to collect multiple digest values obtained applying different hash algorithms on the same input.
+All entries in the `digests-type` MUST have different `alg` values.
+
 ### Tagged Bytes Type {#sec-common-tagged-bytes}
 
 An opaque, variable-length byte string.
@@ -922,8 +926,7 @@ The following describes each member of the `measurement-values-map`.
 
 * `digests` (index 2): Contains the digest(s) of the measured environment
   together with the respective hash algorithm used in the process.
-  Each digest in the array represents one acceptable state of the measured object.
-  See {{sec-common-hash-entry}}.
+  It uses the `digests-type` described in {{sec-common-hash-entry}}.
 
 * `flags` (index 3): Describes security relevant operational modes. For
   example, whether the environment is in a debug mode, recovery mode, not fully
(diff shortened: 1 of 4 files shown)


andrew-draper:
There is text in 5.5.4.5.2 which is incompatible with this text.
To prevent downgrade attacks, the requirement is that all entries in the digest which are present in both reference value and ACS entry and whose hash algorithm is understood by both parties must match.

thomas-fossati:
> There is text in 5.5.4.5.2 which is incompatible with this text. To prevent downgrade attacks, the requirement is that all entries in the digest which are present in both reference value and ACS entry and whose hash algorithm is understood by both parties must match.

Sorry, I don't understand what this text means.

andrew-draper (changes requested):
I agree that this text needs to be clarified, but I think what is written here doesn't match 5.5.4.5.2. That section assumes that all entries in a `stateful-environment` digest array are measurements of the same object using different digest algorithms.
The text proposed here leaves open the possibility that there are two entries with the same algorithm and different values, which will not match with the 5.5.4.5.2 algorithm.
The 5.5.4.5.2 algorithm is written the way it is to prevent downgrade attacks.
We need to decide what happens if a digests array contains multiple different values with the same algorithm. This text suggests a match if one of them matches. The 5.5.4.5.2 text will not match.

thomas-fossati:
> > There is text in 5.5.4.5.2 which is incompatible with this text. To prevent downgrade attacks, the requirement is that all entries in the digest which are present in both reference value and ACS entry and whose hash algorithm is understood by both parties must match.
> 
> Sorry, I don't understand what this text means.

But besides that, is the term _"acceptable state"_ that is incompatible?  If we used _"reference state"_ would that work for you?

thomas-fossati:
> The text proposed here leaves open the possibility that there are two entries with the same algorithm and different values, which will not match with the 5.5.4.5.2 algorithm. 

I am not sure how the same object can hash to the same value using two different algorithms, unless the hash algorithms are broken.  Is this the case you want to protect against?

thomas-fossati:
> The text proposed here leaves open the possibility that there are two entries with the same algorithm and different values, which will not match with the 5.5.4.5.2 algorithm.

I am not sure how the _same_ measured object (e.g., a specific instance of a FW blob) can be hashed to different values using the same hash function.

henkbirkholz (approved):
Clarifying change. LGTM

henkbirkholz:
> I agree that this text needs to be clarified, but I think what is written here doesn't match 5.5.4.5.2. That section assumes that all entries in a `stateful-environment` digest array are measurements of the same object using different digest algorithms. The text proposed here leaves open the possibility that there are two entries with the same algorithm and different values, which will not match with the 5.5.4.5.2 algorithm. The 5.5.4.5.2 algorithm is written the way it is to prevent downgrade attacks. We need to decide what happens if a digests array contains multiple different values with the same algorithm. This text suggests a match if one of them matches. The 5.5.4.5.2 text will not match.

Oh, I see. I agree with Andrew that the change could allow for different states expressed via the same algorithm (and not only same state with different algorithm). Was that the intent?

draft-ietf-rats-corim.md
```diff
@@ -1128,13 +1132,13 @@ checks, and trust anchor checks.
...
 Identifiers are either unsigned integers or text strings and their type matters, e.g., unsigned integer 5 is distinct from the text string "5".
+The digests use `digests-type` semantic ({{sec-common-hash-entry}}).
 
 ~~~ cddl
 {::include cddl/integrity-registers.cddl}
 ~~~
 
 All the measured objects in an Integrity Registers map are explicitly named and the order in which they appear in the map is irrelevant.
-Any digests associated with a measured object represent an acceptable state for the object.
 Therefore, if multiple digests are provided, the acceptable state is their cross-product.
```
thomas-fossati, line 1142:
```suggestion
Any digests associated with a measured object represent an acceptable state for the object.
Therefore, if multiple digests are provided, the acceptable state is their cross-product.
```

nedmsmith:
The updated definition seems to parallel the semantics of a TPM PCR. If modeling a full TPM, there would need to be 24 instances of `integrity-register`. How should this be accomplished? Should there be a different `environment-map` for each PCR. Maybe use the `index` value to disambiguate PCR number?

thomas-fossati:
> The updated definition seems to parallel the semantics of a TPM PCR. If modeling a full TPM, there would need to be 24 instances of `integrity-register`. How should this be accomplished? Should there be a different `environment-map` for each PCR. Maybe use the `index` value to disambiguate PCR number?

That was not my intention.

I guess the problem is in the use of "environment" in _"[A]n **environment** can be measured using different hash algorithms."_ ?

What if we said: _"A measurement can be obtained using different hash algorithms."_ instead?

nedmsmith:
>That was not my intention.

But it is a reasonable intention of someone. I'm just wondering how someone should complete the exercise. 

If the goal is to define `digests` and `integrity-register` similarly, then they seem to be solving the same problem. Only one has a label and the other one doesn't.

nedmsmith:
>I guess the problem is in the use of "environment" 

Technically, using `index` in environment map makes it a different environment. So, the current wording would still be correct.

thomas-fossati:
> > > That was not my intention.
> > 
> > 
> > But it is a reasonable intention of someone. I'm just wondering how someone should complete the exercise.
> > If the goal is to define `digests` and `integrity-register` similarly, then they seem to be solving the same problem. Only one has a label and the other one doesn't.
> 
> I hate to say this ... but one could use `mkey` of type `uint` 🥲

That said, once you have integrity registers why complicate your life?

draft-ietf-rats-corim.md
```diff
@@ -304,6 +304,10 @@ convention - e.g., {{Section 4.4.1.5 of -psa-token}}).
 {::include cddl/digest.cddl}
 ~~~
 
+An environment can be measured using different hash algorithms.
```
thomas-fossati, line 307:
```suggestion
A measurement can be obtained using different hash algorithms.
```

nedmsmith:
> That said, once you have integrity registers why complicate your life?

Is this how you would model a TPM PCR bank using integrity registers?
```
/ integrity registers / 14 : {
	/ per 0 / 0 : [ [“sha1”, h’0001’], [“sha256”, h’0002’]],
	/ per 1 / 1 : [ [“sha1”, h’0001’], [“sha256”, h’0002’]],
	/etc/
	/ per 24 / 24 : [ [“sha1”, h’0001’], [“sha256”, h’0002’]],
}
```

nedmsmith:
> yes

If modeling a complex object that consists for several sub-components?
```
/ integrity registers / 14 : {
	/ my sub-component 0 / "Frodo" : [ [“sha1”, h’0001’]],
	/ my sub-component 1 / "Bilbo" : [ [“sha1”, h’0001’]],
	/etc/
	/ my sub-component 24 / "Smog" : [ [“sha1”, h’0001’]]
}
```

thomas-fossati:
> > yes
> 
> If modeling a complex object that consists for several sub-components?
> 
> ```
> / integrity registers / 14 : {
> 	/ my sub-component 0 / "Frodo" : [ [“sha1”, h’0001’]],
> 	/ my sub-component 1 / "Bilbo" : [ [“sha1”, h’0001’]],
> 	/etc/
> 	/ my sub-component 24 / "Smog" : [ [“sha1”, h’0001’]]
> }
> ```

Yes, if your measurements' metadata is limited to a "name".

nedmsmith (approved):
LGTM

yogeshbdeshpande (approved):
The `text` and usages of `digest` is much clear now!

LGTM!

draft-ietf-rats-corim.md
```diff
@@ -1128,6 +1132,7 @@ checks, and trust anchor checks.
 An Integrity Registers map groups together one or more measured "objects".
 Each measured object has a unique identifier and one or more associated digests.
 Identifiers are either unsigned integers or text strings and their type matters, e.g., unsigned integer 5 is distinct from the text string "5".
+The digests use `digests-type` semantic ({{sec-common-hash-entry}}).
```
deeglaze, line 1135: nit: "semantics" since I never see "semantic" describing program behavior.

draft-ietf-rats-corim.md
```diff
@@ -923,8 +927,8 @@ The following describes each member of the `measurement-values-map`.
   {{sec-comid-svn}}.
 
 * `digests` (index 2): Contains the digest(s) of the measured environment
-  together with the respective hash algorithm used in the process.  See
-  {{sec-common-hash-entry}}.
+  together with the respective hash algorithm used in the process.
```
deeglaze, line 930: Squash single sentence into 1 line.

andrew-draper (approved):

andrew-draper:
Thanks for making the changes to remove the downgrade attack. I am ok with the replacement text.

draft-ietf-rats-corim.md
yogeshbdeshpande: This will be done in future PR.

draft-ietf-rats-corim.md
yogeshbdeshpande, line 1135:
```suggestion
The digests use `digests-type` semantics ({{sec-common-hash-entry}}).
```

draft-ietf-rats-corim.md
thomas-fossati: I'll let Henk who volunteered to make a pass over the whole document to "onelinearise" it

henkbirkholz (approved):

PR #217, opened 2024-03-08
change measurement-value-map to measurement-values-map

andrew-draper:


draft-ietf-rats-corim.md
```diff
@@ -1599,11 +1599,11 @@ The ACS must maintain the authority information for each EMT. There can be
 multiple entries in `state-triples` which have the same `environment-map`
 and a different `authorized-by` field (see {{sec-authorized-by}}).
 
-If the merged measurement-value-map contains duplicate codepoints and the
+If the merged `measurement-values-map` contains duplicate codepoints and the
```
nedmsmith, line 1602: I'm wondering if this case can never happen given it follows cddl. CDDL shouldn't let you have `state-triples` where the `measurement-values-map` contains duplicate code points. I wonder if we really need to describe "merging" since the case above describes redundant copies of an EMT.  If so, you can just delete redundant duplicate EMTs. No need to describe deeper "merging" semantics.

draft-ietf-rats-corim.md
```diff
@@ -1599,11 +1599,11 @@ The ACS must maintain the authority information for each EMT. There can be
...
 measurement values are equivalent, then duplicate claims SHOULD be omitted.
 Equivalence typically means values MUST be binary identical.
 
-If the merged measurement-value-map contains duplicate codepoints and the
+If the merged `measurement-values-map` contains duplicate codepoints and the
```
nedmsmith, line 1606: ditto previous comment.

draft-ietf-rats-corim.md
```diff
@@ -1759,7 +1759,7 @@ A CoRIM profile may define additional tags and their matching algorithms.
 
 If the Verifier does not recognize the stateful environment CBOR tag value then the stateful environment does not match.
 
-If the stateful environment is not tagged and the measurement-value-map key is a value with handling described in  the sub-sections below, then the algorithm appropriate to that key is used to match the entries.
+If the stateful environment is not tagged and the `measurement-values-map` key is a value with handling described in  the sub-sections below, then the algorithm appropriate to that key is used to match the entries.
```
nedmsmith, line 1762:
```suggestion
If the stateful environment is not tagged and the `measurement-values-map` key is a value with handling described in the sub-sections below, then the algorithm appropriate to that key is used to match the entries.
```

nedmsmith (commented):
multiple suggestions

yogeshbdeshpande (approved):
LGTM!

thomas-fossati (approved):
thanks!

Issue #222, opened 2024-04-03
Delegate "validity" to signing key certificates


deeglaze:
We should remove `validity-map` entirely. The validity of a signature is the purview of the PKI only. Validity without a signature is irrelevant.

thomas-fossati:
Absent a way to revoke a certain CoRIM, we must have a mechanism to timebomb it.

That mechanism is the validity-map.

(Even if an explicit revocation mechanism existed, the timebomb is also a useful defense.)

thomas-fossati:
Clarifying question: Is the proposal to require the supply chain to mint a one-shot, short-term certificate for every CoRIM?

deeglaze:
Not necessarily every CoRIM, but perhaps with every CoRIM bundle created for a release, yes. This is common practice for code signing.

deeglaze:
Alternatively, whatever notion of validity a profile wants can be added later. There really should be a core and extended form of CoRIM for all these bells and whistles.

thomas-fossati:
> Alternatively, whatever notion of validity a profile wants can be added later. There really should be a core and extended form of CoRIM for all these bells and whistles.

I think that the lifetime of the information asserted in a CoRIM is a basic feature and should be provided (at least in some crude form) by the base spec.

deeglaze:
Okay, can we say this issue is to move validity-map and uses to a Lifetime management section to discuss its use? It seems there is no removing anything at this point, but we can at least give interpretations as meant by the default (DICE) profile.

Issue #223, opened 2024-04-03
Organize the presentation of the CDDL into skeleton and standard extension tables


deeglaze:
This issue is getting at the heart of my call for simplification to really whittle down the presentation to separate form from function. It's very easy to get lost in the details otherwise.

This will separate out minimum necessary to express reference values from minimum for shared value of the authors of this spec. Endorsements are an extension that are not necessary to verify evidence from an attester, specifically because they are defined as a claim without evidence.

We've discussed but have not implemented the convention that every $$ extension should have its own IANA table to assign meaning to extensible maps. We should specify all extensible maps just as their repeated extension points and provide descriptions of the initial table definitions in a later section.

Extensible type choices are left indeterminate or as a very opinionated minimum selection since they could be parsed as anything. We can provide in a later section some common recommended choices that profile writers could be reasonably confident that their profile could be easily adopted by verifier implementations.

deeglaze:
There are a number of missing IANA registry subsections for extensible map sockets.

```
$$corim-signer-map-extension
$$triples-map-extension
$$measurement-values-map-extension
$$flags-map-extension
$$corim-entity-map-extension
$$comid-entity-map-extension
```

This raises the question of whether the corim and comid entity map extensions should refer to the same registry or not.

yogeshbdeshpande:
Closing as merged the PR #391

PR #224, opened 2024-04-07
Remove validity-map from core and remove meta-map

deeglaze:
The validity of a document should be determined by the signing key's expiration. Any other form of validity should be an extension to the base.

Since the meta-map only had the signer-map in it, I've collapsed that indirection. If the overall corim validity-map is still desired, we can add that as a standard extension as per issue #223

Closes Issue #222

### DIFF
--- a/cddl/corim-map.cddl
+++ b/cddl/corim-map.cddl
@@ -3,7 +3,6 @@ corim-map = {
   &(tags: 1) => [ + $concise-tag-type-choice ]
   ? &(dependent-rims: 2) => [ + corim-locator-map ]
   ? &(profile: 3) => $profile-type-choice
-  ? &(rim-validity: 4) => validity-map
   ? &(entities: 5) => [ + corim-entity-map ]
   * $$corim-map-extension
 }
--- a/cddl/corim-meta-map.cddl
+++ b/cddl/corim-meta-map.cddl
@@ -1,4 +1,0 @@
-corim-meta-map = {
-  &(signer: 0) => corim-signer-map
-  ? &(signature-validity: 1) => validity-map
-}
--- a/asides/theory-of-validity.md
+++ b/asides/theory-of-validity.md
@@ -1,0 +1,129 @@
+# CoRIM Validity Operational Model
+
+The claims in a CoRIM document all have different interpretations and impact on security posture.
+This document describes how component vendors and environment compositors may structure their issuance of CoRIMs to assign appropriate lifetimes to different sorts of claims.
+The main position is that the validity of claims should be bound to the validity of their signatures.
+Public key infrastructure (PKI) already provides mechanisms for certificate lifetimes and otherwise revocation of certificates.
+The CoRIM specification does not seek to reinvent the notion of validity of the CoRIM itself.
+
+## Fundamentals
+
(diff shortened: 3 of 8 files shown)


andrew-draper:
I agree with the direction this PR is taking.

cddl/corim-map.cddl
```diff
@@ -3,7 +3,6 @@ corim-map = {
   &(tags: 1) => [ + $concise-tag-type-choice ]
   ? &(dependent-rims: 2) => [ + corim-locator-map ]
   ? &(profile: 3) => $profile-type-choice
-  ? &(rim-validity: 4) => validity-map
-  ? &(entities: 5) => [ + corim-entity-map ]
+  ? &(entities: 4) => [ + corim-entity-map ]
```
andrew-draper, line 6: Please don't renumber existing map keys as that makes existing files invalid.

deeglaze: Done.

deeglaze:
I've added a document about signing and validity. @yogeshbdeshpande @thomas-fossati @andrew-draper @nedmsmith

deeglaze:
The intended audience is y'all, since there was hesitation about removing the validity-map. I'm not sure how supplemental explainers are meant to be associated with the reference material that is an RFC.

nedmsmith:
> I'm not sure how supplemental explainers are meant to be associated with the reference material that is an RFC.

We have a "musings" wiki https://github.com/ietf-rats-wg/draft-ietf-rats-corim/wiki that has supplemental or explanatory content. It's also fine to add it to the issue itself. The GitHub repo doesn't go away AFAIK. Hence, anyone who is interested can discover justifications for various decisions made.

deeglaze:
I can move the document there now, but I put it in the PR for line-by-line comments in case folks want to discuss.

nedmsmith:
> I put it in the PR for line-by-line comments in case folks want to discuss.

It's unfortunate that GitHub dosn't support line-by-line discussion threads on wiki pages and so forth. However, I find it confusing to put ancillary discussion threads in the build image.

cddl/corim-meta-map.cddl
```diff
@@ -1,4 +0,0 @@
-corim-meta-map = {
-  &(signer: 0) => corim-signer-map
-  ? &(signature-validity: 1) => validity-map
```
nedmsmith, line 3: I believe use of validity here is only as a hint to parsers who would have to parse the payload if a security relevant decision is being made as corim-meta-map is part of the COSE unprotected header. But given there may be implementations already that use corim-signer-map, maybe it is best to leave it as is? Note, that the freshness mechanism in corim (as is or as an extension to the corim-map) implies the hint in the unprotected header should parallel what is in the payload. Hence, corim-meta-map should be similarly extensible.  Given there may be existing use, leaving it as is and relying on profiles to define its use is less disruptive. I don't think collapsing corim-meta-map into corim-signer-map for simplification is that significant (esp. if the goal of the hint is to parallel what is included in the payload).

deeglaze:
My understanding is that removing the validity map at this point is a non-option. I'll close this and open an issue to create a lifecycle management section that describes lifecycle management a bit more holistically in such a way that it's clear that validity-map is an option but not prescriptive for lifecycle management.

PR #226, opened 2024-04-23
Change Measurement to Property where appropriate

deeglaze:
A property is either a reference value or endorsed value, so the code points used and nomenclature in the document are overly constrained in points.

Closes Issue#225


nedmsmith:
It seems this PR is doing a global replace of the word measurement with claim while leaving the semantics in place. Is that correct?

nedmsmith:
Id did a quick review of 9334. It seems measurement and claim can be interchangeable. For example:
 "Measurements can describe a variety
   of attributes of system components, such as hardware, firmware, BIOS,
   software, etc., and how they are hardened."
and
"Endorsement:  A secure statement that an Endorser vouches for the
      integrity of an Attester's various capabilities, such as Claims
      collection and Evidence signing."
I don't think use of measurement instead of claim is technically incorrect.

deeglaze:
The semantics stay in place, correct.

RFC9334 states that a claim appears to be synonymous with evidence, which is different from how claims are treated in the theory of verification and the notion of Claim Sets. It then points at Claim from RFC7519, which defines claim as "A piece of information asserted about a subject."

Endorsements and reference measurements are claims of security attributes and authenticity for the claim values they embody.

It's confusing to use "measurement" to describe unmeasurable qualities, such as "certified EAL4".

nedmsmith (approved):
LGTM

nedmsmith:
Note that the current definition of "Claim: A piece of information, in the form of a key-value pair." is very broad and open ended. Pretty much any map or record expressed in CDDL fits this definition. The use of Claim in ECT is intended to be more focused on the properties that are associated with environments. I prefer using something different such as Environment-Properties-Tuple (EPT). 

It is OK to talk about claims in the abstract such as when describing CMs as consisting of "claims". 

The history behind the term being used in RATS is that it intends to follow the definition originated by JWT. It is a non-goal that ECT imply there's a JWT embedded in an ECT.

henkbirkholz:
The `C`laim in `ECT` came up several times. The generic use of Claims (as defined in JWT) and the increasingly specific use in RFC 9334 and then here seem to create more confusion than necessary.

We are considering changing Claim to a less overloaded synonym, such as property or characteristic. In the end it is just naming and I'd typically refrain from doing name changes at this stage. Then again, the topic starts to eat time and maybe it is better to go with a synonym for now?

draft-ietf-rats-corim.md
```diff
@@ -843,68 +843,70 @@ The types defined for a group identified are UUID and variable-length opaque byt
 {::include cddl/group-id-type-choice.cddl}
 ~~~
 
-##### Measurements
+##### Claims
+
+A Claim may either be a measurement that can be directly compared with attestation evidence, or it can be an endorsement of a quality that is not directly comparable with evidence.
```
thomas-fossati, line 848: Trying to recover a suggestion that got lost in the discussion.
```suggestion
A Claim represents a single piece of actual or reference state.
A Claim may either be a measurement that can be directly compared with attestation evidence, or it can be an endorsement of a quality that is not directly comparable with evidence.
```
There was a further suggestion to s/measurement/reference value/ in L848

draft-ietf-rats-corim.md
```diff
@@ -845,28 +845,37 @@ The types defined for a group identified are UUID and variable-length opaque byt
...
 
+A Claim may be called an assertion, property, measurement, or many other classifications based on its role in Evidence appraisal.
+Attester-authenticated Claims are typically called measurements.
+Measurements can be of a variety of components including software, firmware, configuration files, read-only memory, fuses, IO ring configuration, partial reconfiguration regions, etc.
+Measurements comprise raw values, digests, or status information.
+
+There may be multiple Claims made of the same Target Environment.
+A single Claim may describe a combination of multiple components.
+
+The scope of a Claim depends on the specificity of the environment in the triple it appears in.
```
thomas-fossati, line 867:
```suggestion
The triple’s environment defines the scope of a Claim.
```

thomas-fossati (approved):
LGTM

I've left a couple of purely editorial comments.

draft-ietf-rats-corim.md
```diff
@@ -769,68 +769,79 @@ The types defined for a group identified are UUID and variable-length opaque byt
...
 ~~~
 
-##### Measurements
+##### Claims
 
-Measurements can be of a variety of things including software, firmware,
-configuration files, read-only memory, fuses, IO ring configuration, partial
-reconfiguration regions, etc. Measurements comprise raw values, digests, or
-status information.
+A Claim is a key, value pair that in CoRIM represents a single piece of actual state or a reference state.
```
thomas-fossati, line 774: ["key-value pair"](https://en.wikipedia.org/wiki/Name–value_pair)

nedmsmith:
This PR doesn't fully address the terminology differences between measurement and claim as the `mkey` and `mval` members borrow `m` from measurement. 
Additionally, the 9334 definition of claim is at a higher level of abstraction and applies equally to `enviornment-map` as it does `measurement-map` or some other CDDL expression. 
My preference is to leave "claim" as an abstract term and come up with different vocabulary to refer to the values not in environment-map. 
Maybe use properties `properties-map` instead of `claims-map`? 
This would suggest `mval` would change to `pval` and `mkey` to `pkey`. 
I guess I'm indifferent whether `claim-values-map` changes to `property-values-map` if the bounding structure is called `properties-map`.

This implies ECT becomes EPT.

draft-ietf-rats-corim.md
nedmsmith: If we're following the endorsements draft terminology it would read something like: "A Claim is a key-value pair that in CoRIM represents a single piece of actual or possible state.

deeglaze:
I've updated the text to used "property" instead.

draft-ietf-rats-corim.md
```diff
@@ -1523,44 +1535,44 @@ To facilitate the description of normative Verifier behavior, this document uses
 The following terms are used:
 
 {: vspace="0"}
-Claim:
+Property:
 : A piece of information, in the form of a key-value pair.
```
henkbirkholz, line 1539: Here the Property and the Claim definition that RFC9334 inherits collide a bit. Maybe something like:

```suggestion
: A piece of information, in the form of a Claim as specified in {{-rats-arch}} where the name of the name/value pair is always an integer with an annotated text string name.
```

draft-ietf-rats-corim.md
```diff
@@ -768,68 +769,79 @@ The types defined for a group identified are UUID and variable-length opaque byt
 {::include cddl/group-id-type-choice.cddl}
 ~~~
 
-##### Measurements
+##### Properties
+
+A Property is a key-value pair that in CoRIM represents a single piece of actual state or a possible state.
```
henkbirkholz, line 774: Is now unfortunately defined in a list of terms later in the document. Maybe:

```suggestion
A Property in CoRIM represents a single piece of actual state or a possible state (see {{sec-verifier-abstraction}} below).
```

draft-ietf-rats-corim.md
```diff
@@ -768,68 +769,79 @@ The types defined for a group identified are UUID and variable-length opaque byt
...
+
+A Property in CoRIM always appears within a triple.
+The triple expresses the intended meaning of the Property.
+In all cases, a Property is made by an authority, such as an Attester, Endorser, or Reference Value Provider.
+
+A Property from an Attester is called Evidence.
```
henkbirkholz, line 780:
```suggestion
A Property from an Attester is conveyed via Evidence.
```

draft-ietf-rats-corim.md
```diff
@@ -768,68 +769,79 @@ The types defined for a group identified are UUID and variable-length opaque byt
...
+A Property in a `reference-triple-record` is called a Reference Value.
+A Property in another triple is called an Endorsement.
```
henkbirkholz, line 782: Does the list in [5.1.4 Triples](https://ietf-rats-wg.github.io/draft-ietf-rats-corim/draft-ietf-rats-corim.html#section-5.1.4) fit that statement?

draft-ietf-rats-corim.md
```diff
@@ -484,11 +485,11 @@ The following describes each child item of this group.
 
 # Concise Module Identifier (CoMID) {#sec-comid}
 
-A CoMID tag contains information about hardware, firmware, or module composition.
+A CoMID tag contains information about subjects such as hardware, firmware, or module composition.
 
 Each CoMID has a unique ID that is used to unambigously identify CoMID instances when cross referencing CoMID tags, for example in typed link relations, or in a CoBOM tag.
 
-A CoMID defines several types of Claims, using "triples" semantics.
+A CoMID represents expectations for Evidence appraisal with "triples".
```
henkbirkholz, line 492: Is the use of `expectations` trying to combine the meanings of reference values and endorsed values?

draft-ietf-rats-corim.md
```diff
@@ -768,68 +769,79 @@ The types defined for a group identified are UUID and variable-length opaque byt
...
+A Property from an Attester is called Evidence.
+A Property in a `reference-triple-record` is called a Reference Value.
+A Property in another triple is called an Endorsement.
+A Reference Value triple is expressible as a conditional Endorsement that if the Attester made some Properties, then the RVP also makes those Properties.
 
-Measurements can be of a variety of things including software, firmware,
-configuration files, read-only memory, fuses, IO ring configuration, partial
-reconfiguration regions, etc. Measurements comprise raw values, digests, or
-status information.
+A Property may be called an assertion, property, measurement, or many other classifications based on its role in Evidence appraisal.
```
henkbirkholz, line 785: I assume that property and claim have to switch places here.

```suggestion
A Property may be called an assertion, claim, measurement, or many other classifications based on its role in Evidence appraisal.
```
And then, why do we say that here in the first place? Should that go to the definition of Property?

draft-ietf-rats-corim.md
```diff
@@ -768,68 +769,79 @@ The types defined for a group identified are UUID and variable-length opaque byt
...
 
-###### Measurement Values {#sec-comid-mval}
+###### Property Values {#sec-comid-pval}
 
-A `measurement-values-map` contains measurements associated with a certain
+A `property-values-map` contains properties associated with a certain
 environment. Depending on the context (triple) in which they are found,
-elements in a `measurement-values-map` can represent class or instance
-measurements. Note that some of the elements have instance scope only.
+elements in a `property-values-map` can represent class or instance
```
henkbirkholz, line 833:
```suggestion
members of a `property-values-map` can represent class or instance
```

draft-ietf-rats-corim.md
```diff
@@ -768,68 +769,79 @@ The types defined for a group identified are UUID and variable-length opaque byt
...
+properties. Note that some of the elements have instance scope only.
```
henkbirkholz, line 834:
```suggestion
properties. Note that some of the member types have instance scope only.
```

draft-ietf-rats-corim.md
thomas-fossati:
```
: Compare: Claim {{-rats-arch}}
```

yogeshbdeshpande:
Lets wait for @nedmsmith to come back and we can settle with something which is liked by majority!

andrew-draper (changes requested):
I think some parts of this change are good, but I don't see the value in changing "Claim" to "Property" throughout. Let's wait for Ned to come back and discuss again.

deeglaze:
In the August 21 meeting, we all shared our thoughts that Claim is more consistent with RFC 9334 and preferable to Property. @nedmsmith if you could please explain more your desire for "property" that would help us make a decision.

deeglaze:
> We will retain 'Claim', waiting for all editors to agree on that.

Thanks, just let me know when and I'll revert the claim->property commit of this PR.

deeglaze:
Do we give up and just change ACS to AMS and ECT to EMT to keep internal consistency?

yogeshbdeshpande:
Unanimously Agreed that property is not sutable, but claims is more suitable, hence closing this PR!

Issue #176, opened 2023-12-06
Need to define the namespace scoping for Environment


yogeshbdeshpande:
An Attester can be from multiple vendors though, the Elements of Class Identifier are Optional Elements like
Vendor, Model, Class-ID etc.

This leaves the scope for a user to only add a Model, which could collide with other Vendor Models.

Need to define the usage of Class Identifier that should be unique across Multiple Vendors and Multiple Attesters belonging to the same Vendor. Specification needs some clarification about usage on Class Identifier, been how to mint it as a Unique Identifier!

andrew-draper:
My assumption is that the attester and/or the evidence collection protocol defines the `environment-map`. During ACS initialization the evidence entries with this environment map are added to the ACS.
CoMID triples intending to interact with this evidence should use the same `environment-map` as appears in evidence.
CoMID triples can add Endorsements with a new `environment-map` and can select a unique value for the `vendor` field if they want a new namespace.
We have added a free-format `instance` field which can be used to identify different device instances, is any more work needed to close this issue?

nedmsmith:
Environment includes both `environment-map` and `stateful-environment-record` as various triples may use both to identify Target Environments (TE). 

The existing prose for `environment-map` doesn't explicitly state namespace scope expectations. In the case of `instance-id` the types of instance values suggest global namespace scope (as a `$crypto-key-type-choice` implies cryptographic identifiers have global uniqueness. `class-id` forms include `uuid` and `oid` which also imply global scope. However, it also supports `int` and `bytes` variants that don't necessarily imply global scope. Additionally, use of `class-id` is optional leaving various combinations of 'vendor', 'model', 'layer', 'index' that have no implied scoping. 

If Environments are required to have global scope, then it is impossible (absent malicious intent) for environment-A to be confused with environment-B due to name collision. If it is impractical to expect environments will have global scope, the rules for local scoping need to be defined. 

At one point, there was wording in a PR that suggested namespace scope could be localized by placing an environment inside a grouping construct such as a 'domain' or 'bundle'. This may be a reasonable approach to support locally scoped namespaces, but is there a use case that motivates it?

[...]

PR #227, opened 2024-04-24
environment-map global scope

nedmsmith:
Added text requiring environment-map to be globally unique


nedmsmith:
This PR addresses issue #176

deeglaze:
"Globally unique" doesn't really match my understanding given that an environment can be elaborated to have an instance field to make it even-more-unique for matching instances against reference values for classes.

deeglaze:
In the meeting you said just that the class-id needs to be globally unique, which makes more sense to me.

draft-ietf-rats-corim.md
```diff
@@ -777,6 +777,9 @@ context (triple) in which the environment is used.
 An environment is named after a class, instance or group identifier (or a
 combination thereof).
 
+An environment MUST be globally unique.
+The combination of values within `environment-map` may be combined so that the combination is globally unique.
```
andrew-draper, line 781:
```suggestion
The combination of values within `environment-map` must combine to form a globally unique identifier.
```

yogeshbdeshpande (approved):

draft-ietf-rats-corim.md
```diff
@@ -777,6 +777,9 @@ context (triple) in which the environment is used.
 An environment is named after a class, instance or group identifier (or a
 combination thereof).
 
+An environment MUST be globally unique.
```
andrew-draper, line 780: For example, an environment with just `class-id` may denote a specific organization's SKU (product name and specifications), the `instance` field is unique within the `class-id` namespace.

andrew-draper, line 780:
```suggestion
An environment MUST be globally unique.
For example, an environment with just `class-id` may denote a specific organization's SKU (product name and specifications), the `instance` field is unique within the `class-id` namespace.
```

nedmsmith:
Might I suggest that additional clarification in the form of examples be added by a different PR? It seems there is consensus on the simple change this PR originally proposed.

nedmsmith (commented):
I resolved the threads related to exemplary text since we can add this later.

andrew-draper (approved):
I think the proposed changes attached to this revision make the spec better.
I suggest we approve these now, and to put additional changes into another PR.

PR #245, opened 2024-06-24
Clarify untagged value comparison requirements

deeglaze:
As stated, it appears that keys of measurement-values-map in this document are given more leeway for specifying comparisons than profiles. I interpret the text as saying any untagged value for a key that is not 0 through 14 must be compared with binary equality. There is no MUST in the current language, so that does not seem to be the intention. A profile should be allowed to specify a comparison method as irregular as raw-value and raw-value-mask without the need for registering a CBOR tag number with the IANA.

If a profile's comparison method will be common to multiple profiles that are intended for general verifier consumption, then we just recommend that the comparison gets registered with a CBOR tag to avoid having to duplicate language and potentially very slightly deviate from the original meaning in different profiles.

### DIFF
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -2068,10 +2068,10 @@ Additions to the ACS MUST be atomic.
 
 [^issue] https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/71
 
-This specification defines the comparison algorithm for the codepoints and CBOR tagged values described in sub-sections below.
-A CoRIM profile may define additional tags and their matching algorithms.
-Specifications that extend CoMID MUST also define comparison algorithms for the extended values.
-Any new codepoints requiring non-default comparison MUST add a CBOR tag to the extension that describs the desired behaviour.
+This specification defines the comparison algorithm for the codepoints described in sub-sections below.
+A CoRIM profile may define additional negative codepoints and their matching algorithms.
+Specifications that extend CoMID MUST also define comparison algorithms for their added codepoints, and augment comparison algorithms if existing codepoints' value CDDL expands.
+If a codepoint's comparison algorithm is not stated or is referred to as "default", then the Verifier MUST compare the binary equality of the CBOR encodings of the values.
 
 ### Environment Comparison {#sec-compare-env}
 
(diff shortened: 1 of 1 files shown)


deeglaze:
@nedmsmith @yogeshbdeshpande updated following the meeting discussion.

deeglaze:
I'll add the evidence transformation changes in another commit soon.

draft-ietf-rats-corim.md
```diff
@@ -2097,19 +2097,15 @@ The comparison algorithm performed depends on the value of the codepoint being c
...
 
-If the `measurement-values-map` value has an associated CBOR tag.
-The comparison algorithm should comprehend the structure identified by the CBOR tag.
+If the `measurement-values-map` value has an associated CBOR tag, the comparison algorithm should comprehend the structure identified by the CBOR tag.
 
-If the stateful environment `measurement-values-map` value is tagged with a CBOR tag {{-cbor}} then the Verifier MUST use the comparison algorithm associated with that tag.
+If the Verifier does not recognize a CBOR tag value then the value MUST NOT match.
 
-If the value is not tagged then the Verifier MUST use the comparison algorithm associated with the `measurement-values-map` codepoint for the entry.
+Note: CBOR tags are useful for discriminating values amongst alternates, but the comparison of tagged values is still determined by the codepoint they appear under. It is recommended but not required to compare values with a specific CBOR tag the same way across codepoints. For this reason, it is recommended to specify a default comparison algorithm with the CBOR tag's registration.
```
nedmsmith, line 2107:
```suggestion
The CBOR tag's registration specification SHOULD specify a comparison algorithm if not defined in this or other specification.

Note: The comparison algorithm of CBOR tagged values is determined by the key (i.e., codepoint) they appear under.
It is recommended, but not required, that comparison algorithms involving the same CBOR tag process the tagged information similarly across the different codepoints.
```

draft-ietf-rats-corim.md
```diff
@@ -2097,19 +2097,15 @@ The comparison algorithm performed depends on the value of the codepoint being c
...
 
-If the Verifier does not recognize the stateful environment CBOR tag value then the stateful environment does not match.
+Profile writers SHOULD use CBOR tags for widely applicable comparison methods to ease Verifier implementation compliance across profiles.
 
-If the stateful environment is not tagged and the `measurement-values-map` key is a value with handling described in the sub-sections below, then the algorithm appropriate to that key is used to match the entries.
-
-If the stateful environment is not tagged, and the `measurement-values-map` key is not a value described below, then the entries are compared using binary comparison of their CBOR encoded values.
-If the values are not binary identical then the stateful environment does not match.
+The following subsections define non-default comparison algorithms for some `measurement-values-map` key codepoints.
```
nedmsmith, line 2111:
```suggestion
The following subsections define the comparison algorithms for the `measurement-values-map` keys defined by this specification.
```

nedmsmith (commented):
See suggested changes.

yogeshbdeshpande:
@deeglaze I agree with Ned's comments. So if you apply them, I am happy to approve this..

draft-ietf-rats-corim.md
yogeshbdeshpande: I would correct the `suggested` edit on New Line  2107 to:

The CBOR tag's registration specification SHOULD specify a comparison algorithm either in this specification or in an equivalent profile document of this specification

nedmsmith: >this specification or in an equivalent profile

I was leaving the door open for other I-Ds or even other standards bodies could define extensions to measurement-values-map or triples-map that result in the need for a comparison algorithm definition. The IANA experts are in charge of overseeing that the extensions are aligned with base corim and other standard extensions. A profile can't define a "standard" extension / comparison algorithm (based on an IANA registry), but it can define non-standard ones using CBOR tags' vendor specific range.

nedmsmith (commented):
See additional comments / suggestions.

draft-ietf-rats-corim.md
deeglaze: I don't know what vendor specific range you're talking about. https://www.iana.org/assignments/cbor-tags/cbor-tags.xhtml has 3 classes of ranges, and none of them suggest that a tag number can be reinterpreted in a vendor-specific way.

nedmsmith: > I don't know what vendor specific range you're talking about. https://www.iana.org/assignments/cbor-tags/cbor-tags.xhtml has 3 classes of ranges, and none of them suggest that a tag number can be reinterpreted in a vendor-specific way.

The semantics for allocating from the first come first served range is the definition of what is behind the tag doesn't have to be in a standard specification. The requester can just say what it is. This process makes it effectively a vendor specific / defined CBOR tag.

yogeshbdeshpande (approved):
LGTM!

yogeshbdeshpande (approved):
LGTM!

henkbirkholz (approved):

nedmsmith (approved):

PR #246, opened 2024-06-27
ascii art

thomas-fossati:
some ascii art I have made that may be useful as-is or as a starting point


nedmsmith:
It is reasonable for the Verifier to assert new claims (which could be treated like endorsements) but the diagrams don't suggest this possibility exists. The use case for this is if the Verifier applies additional integrity checking operations (which is implied by the identity-key/attest-key triples and trust dependency triples. These are applied after the other endorsements, so it makes sense to show another ECS'''' IMHO. Possibly, our revised RATS Arch diagram should include a flow from the RP to the Verifier labeled "Interesting Claims". Trying to do this through some sort of backchannel between RP Owner and Verifier Owner seems dubious.

Also, the RATS Arch omitted a flow from the RP to the Verifier that allows the Verifier to know which claims are "interesting". This is different from Appraisal Policy because of the entity that originates it. Of course, we can say that this flow is out of scope (but exists), but it affects the staging. If the diagram models policy evaluation as a separate (but logical) phase, and the only input is appraisal policy then there isn't enough information available for an RP to specify a condition of which types of claims are interesting.

nedmsmith (approved):
LGTM

Issue #179, opened 2023-12-06
Cleanup Section 5.4.3. Adding CoMID Endorsed Values to the Accepted Claims Set


nedmsmith:
Section 5.4.3 "Adding CoMID Endorsed Values to the Accepted Claims Set" hasn't been authored yet. It seems relevant to work on this text since there is on-going work to define and refine endorsement triples.

Suggested text:
"An Endorser may publish Endorsements, which add new claims to the Accepted Claims Set (ACS) if specified conditions of the ACS are met. This section describes the process performed by the Verifier to determine if the ACS status satisfies the Endorsement condition and where to add the Endorsement claims.

The verifier compares the condition specified by endorsement triples to the ACS. Endorsement conditions may describe environments, or combinations of environment-measurement pairings, specific measurement values, and the authority that asserted the claims.

If the condition matches, the Endorsement claims, including the Endorser's authority, are added to the ACS. The new Endorsed claims are added to the part of the ACS as described by Endorsement triples (see {{sec-comid-triple-endval}}, {{sec-comid-triple-cond-series}}, {{sec-comid-triple-cond-end}}, {{sec-comid-triple-mec-endorsement}})."

deeglaze:
I don't understand the wontfix, since I don't understand endorsed triples from the text as written.

> An Endorsed Values triple declares additional claims that are valid when a Target Environment has been verified against reference measurements.

There is no separation of the measurement-values-map contents into evidence property or endorsement property, so I'm not sure what the actual condition is that you're checking. If "A says X only if B says X about environment E", then are we supposed to track that X came from reference triples instead of endorsement triples, so that to process an endorsement "A says Y only if C says Y about environment E", we check that E had a reference triple already accepted by A (A is the CoRIM issuer)? And the properties in question are irrelevant? Are all reference triples required preconditions or is does just one need to be accepted?

I don't see similar conditions on reference values for the conditional endorsements triples, so it seems only this triple has a triple-type processing order requirement. Is the order across CoMID processing, within fixed point, or only locally to the single CoMID in a single step?

nedmsmith:
Addressed through refactoring.

Issue #70, opened 2023-03-17
Matching a CoMID against a single attester or multiple attesters


andrew-draper:
## Verifier matching measurements from a single set of attestation environments

This issue uses the term Attestation Transport to describe the transport over which evidence was passed to the Verifier. A SPDM connection is an example of an Attestation Transport, it carries measurements from the SPDM Attestation Environment and from related DICE Attestation Environments within the same entity.

The simple use case for a Verifier is where it has received a set of measurements through a single Attestation Transport and it needs to match these against a CoMID tag containing reference values.

IMHO we should make the simple case easy to encode. I suggest that when matching a Reference Value Triple containing only `environment-map` and `measurement-values-map`, the Verifier must ensure that match them against evidence received from the same Attestation Transport. In some cases there may be multiple Attestation Environment groups which match against the same Reference Value Triple, if this happens then each group would receive an identical Endorsement.

For example, a CoMID tag encoding the semantics "If this device has a measurement identified by OID 1.2.3.4 and SHA-384 hash 0x5678 then the device is an Acme Ethernet Card version 1.0"

## Verifier matching measurements from multiple attesters

Higher level Verifiers, or rather the CoRIMs controlling higher level Verifiers, will want to match against more complicated sets of evidence. For example, a CoRIM might want to encode the semantics _Match against a type A PCIe card in slot 0, with firmware hash B, and against a second identical PCIe card in slot 1_.

Do we need a way for a CoRIM to indicate that it must match against a set of evidence from one source (eg one Attestation Transport) and a potentially different set of evidence from a different source?

## An example encoding of reference-values when matching against multiple Attestation Transports

I suggest that we use the `mkey` field to identify the attestor to be chosen.

For example, the following encoding encodes the two PCIe cards case above:

```
- Reference Value Triples
  - (first card)
    - Environment Map A
    - Measurement Map
      - mkey = K1
      - mvalue = B
  - (second card)
    - Environment Map A
    - Measurement Map
      - mkey = K2
      - mvalue = B

```
This would be associated with an Endorsement triple or similar providing more information about the combined system.

The exact values of K1 and K2 are not important, what is important is that all measurements with the same `mkey` value must come over the same Attestation Transport. Using the integer encoding of `mkey` normally produces the smallest encoding.

We might need a profile to enable this behavior.

I think this can be extended to multiple levels of compound devices, but wanted to get a sense of whether this is going in the right direction first.

## Question

[...]

PR #248, opened 2024-07-03
Editorial changes to measurement discussion

deeglaze:
These are have been separated from the renaming in PR#226


draft-ietf-rats-corim.md
```diff
@@ -166,11 +166,19 @@ This specification describes the CoRIM format and documents how a Verifier shoul
...
 In addition to the external CoRIM documents, the Verifier is expected to create an internal representation for each input and map each external representation to an internal one.
 By using the internal representation, the Verifier processes inputs as if they are part of a conversation, keeping track of who said what.
-The origin of the inputs is tracked as *authority*.
-The authority for the Claims in a CoRIM is the CoRIM issuer.
 
 Effectively, Attesters, Reference Value Providers, Endorsers, Verifier Owners, Relying Parties, and even the Verifier potentially all contribute to the conversation.
-Each producer of corresponding RATS Conceptual Messages can assert Claims about an Attester's actual or allowed state.
+These individual principals of the conversation may have their identity rooted back to some *authority* farther up the certificate chain than the signing key itself, such as a digital certificate authority intermediate or root key.
+The origin of contributions/inputs to the conversation are tracked at the granularity of authority, so CoRIM authors can predicate their reference measurements or endorsements based on existing properties rooted from the same authority.
```
henkbirkholz, line 172: This might require en example to be less abstract?

draft-ietf-rats-corim.md
```diff
@@ -166,11 +166,19 @@ This specification describes the CoRIM format and documents how a Verifier shoul
...
+
+CoRIM inputs to the conversation include measurements within a specific Environment.
```
henkbirkholz, line 174: you are introducing "the conversation" here. Is there some similar wording that already exists in the I-D to use instead. Otherwise, we have to clarify a bit of the scope, involved actors, and subject of the conversation, I think

draft-ietf-rats-corim.md
```diff
@@ -166,11 +166,19 @@ This specification describes the CoRIM format and documents how a Verifier shoul
...
+A measurement never appears outside the scope of an associated Environment, so the Environment-Claim Tuple (ECT) is a data structure used to encapsulate all scope context for a given set of measurements.
```
henkbirkholz, line 175: Maybe: 
```suggestion
A specific set of measuremenst always originates from a single identifiable and associated Environment. Hence. an Environment-Claim Tuple (ECT) is a data structure used to encapsulate that scope context for a given set of measurements.
```

draft-ietf-rats-corim.md
```diff
@@ -166,11 +166,19 @@ This specification describes the CoRIM format and documents how a Verifier shoul
...
+The authority for an ECT from a CoRIM is the CoRIM issuer.
+The collective content of an ECT may be referred to as a property, claim, or assertion when the content need not be presented in the ECT data structure.
```
henkbirkholz, line 177:
```suggestion
When the content of an a ECT data structure is presented to roles external to the Verifier, it MAY be referred to via different terms, such as property, claim, or assertion. Whenever an ECT is consumed by a role external to the Verifer, the consuming party MUST always attribute provenance and the ECT's intended use.
```

draft-ietf-rats-corim.md
```diff
@@ -166,11 +166,19 @@ This specification describes the CoRIM format and documents how a Verifier shoul
...
+
+Important: The Verifier's objective is to produce a list of properties that describe the Attester's presumed actual state.
+
+Each producer of corresponding RATS Conceptual Messages can assert claims about an Attester's actual or allowed state.
```
henkbirkholz, line 181: Are you also implying that Evidence can assert claims about allowed state?

draft-ietf-rats-corim.md
```diff
@@ -771,26 +779,24 @@ The types defined for a group identified are UUID and variable-length opaque byt
 
 ##### Measurements
 
-Measurements can be of a variety of things including software, firmware,
-configuration files, read-only memory, fuses, IO ring configuration, partial
-reconfiguration regions, etc. Measurements comprise raw values, digests, or
-status information.
+A measurement can be directly compared with attestation evidence, or it can be a quality that is not directly comparable with evidence.
```
henkbirkholz, line 782: I cannot really follow "[Evidence] can be a quality that is not directly comparable with evidence". Could you maybe provide an example so it becomes more clear?

henkbirkholz: In general, I think that Evidence is composed of "Measurements" and there might be a bigger disconnect here that I cannot follow.

draft-ietf-rats-corim.md
```diff
@@ -771,26 +779,24 @@ The types defined for a group identified are UUID and variable-length opaque byt
...
+The scope of a mroperty is defined by the `environment-map` and optional `mkey` it appears with.
```
henkbirkholz, line 783:
```suggestion
The scope of a property is defined by the `environment-map` and optional `mkey` it appears with.
```

draft-ietf-rats-corim.md
```diff
@@ -771,26 +779,24 @@ The types defined for a group identified are UUID and variable-length opaque byt
...
 
-An environment has one or more measurable elements. Each element can have a
-dedicated measurement or multiple elements could be combined into a single
-measurement. Measurements can have class, instance or group scope.  This is
-typically determined by the triple's environment.
+Measurements can be of a variety of things including software, firmware, configuration files, read-only memory, fuses, IO ring configuration, partial reconfiguration regions, etc. Measurements comprise raw values, digests, or status information.
```
henkbirkholz, line 785:
```suggestion
Measurements can be of a variety of things including measurement values about software, firmware, configuration files, read-only memory, fuses, IO ring configuration, partial reconfiguration regions, etc. Measurements comprise raw values, digests, or status information.
```

draft-ietf-rats-corim.md
```diff
@@ -771,26 +779,24 @@ The types defined for a group identified are UUID and variable-length opaque byt
...
 
-Class measurements apply generally to all the Attesters in the given class.
-Instance measurements apply to a specific Attester instance.  Environments
-identified by a class identifier have measurements that are common to the
-class. Environments identified by an instance identifier have measurements that
-are specific to that instance.
+Each element can have a dedicated measurement/endorsement value or multiple elements could be combined into a single measurement.
```
henkbirkholz, line 787: This might benefit from another example.

draft-ietf-rats-corim.md
```diff
@@ -771,26 +779,24 @@ The types defined for a group identified are UUID and variable-length opaque byt
...
+An `environment-map` contains at least one of: class, instance, or group.
+The more elements in the `environment-map`, the narrower the scope, since more elements must be in common to match.
+Measurements paired with an `environment-map` that contains only a class may be referred to as class mroperties, and similarly with instance and group measurements for respective elements of the `environment-map`.
```
henkbirkholz, line 790:
```suggestion
Measurements paired with an `environment-map` that contains only a class may be referred to as class properties, and similarly with instance and group measurements for respective elements of the `environment-map`.
```

draft-ietf-rats-corim.md
```diff
@@ -771,26 +779,24 @@ The types defined for a group identified are UUID and variable-length opaque byt
...
 
-The supply chain entity that is responsible for providing the the measurements (i.e. Reference Values or Endorsed Values)
-is by default the CoRIM signer. If a different entity is authorized to provide measurement values,
-the `authorized-by` statement can be supplied in the `measurement-map`.
+A measurement from an Attester is called Evidence.
+A measurement in a `reference-triple-record` is called a Reference Value.
+A measurement in another triple is called an Endorsement.
```
henkbirkholz, line 794: Another triple?

draft-ietf-rats-corim.md
deeglaze: I'm not introducing it here. It's introduced in text immediately prior about all parties that contribute to "the conversation" with authorities tracking "who said what".

draft-ietf-rats-corim.md
deeglaze: This suggestion is saying much more than I'm saying, since I don't insist that each specific set of measurements has a single environment. Indeed, sets of measurements don't necessarily have a globally unique identifier for distinguishing them across individual measurements. I can have the same card on two different sockets and have the same measurement set, but a different environment due to my assignment of card to group.

Perhaps "specific set" is too ambiguous. Let's go with "a specific occurrence of a set", since the occurrence refers to where it is in the CoRIM.

draft-ietf-rats-corim.md
deeglaze: What is "intended use"? I'm not sure how to comply with that. My code that touches it could be my intention because I'm a hot shot with "self-documenting code". Is that enough?

I thought ECT is a representation of an utterance in the conversation, with all scope-closing elements provided.

draft-ietf-rats-corim.md
deeglaze: Conceptual messages aren't all evidence. I thought there was already a cm-ind allocated for policy?

draft-ietf-rats-corim.md
deeglaze: I don't think that an assigned name or serial number is always provided by an attesting environment's measurement mechanism. That information is added metadata provided the reference values are as expected.

pseudocode:

```
reference-values-triple [E, M[integrity-registers: X]]
conditional-endorsement-triple [
  conds: [E, [integrity-registers: X authorities: [me]]]
  [E, M'[name: "Product: Enterprise Edition", svn: 5]]
]
```

I'm not really sure how to express associated metadata in a less-clunky way, unless this conditionality is what is the expected semantics of the non-conditional endorsement-triple. It's unclear from the text https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/179#issuecomment-2184116580

draft-ietf-rats-corim.md
deeglaze: Every non-reference-value-triple that contains measurement-map is some flavor of endorsement, so that's the most concise way I could state that. I'm still unclear if there's a conditional relationship between reference triples and the rest of the triples in the same CoMID.

draft-ietf-rats-corim.md
deeglaze: Added. Resolve if it's okay.

draft-ietf-rats-corim.md
deeglaze: Added. Resolve if it's okay.

draft-ietf-rats-corim.md
yogeshbdeshpande: @Dionna This is aintroductory text, making some high level statements. 
- This is way too a technical details in the introductory sections
- Specific items like authority for entities like Relying Party are not in scope of this document.

draft-ietf-rats-corim.md
nedmsmith: >I'm still unclear if there's a conditional relationship between reference triples and the rest of the triples 

If they're related, the relationship is defined by the 'condition' portion of the RV triple. In the `reference-triples` case, both the `environment-map` and `measurement-map` are the condition. If there is a match, the "corroborated" values (the actual state representation) is added to the ACS under the RVP authority. If RVs are already expressed as actual state representations, then you can say "corroboration" is passthrough and the RVPs are copied into the ACS (differing only by authority from what was matched).

There is an ongoing discussion whether RVs should be used to "corroborate" ACS values that originated from Endorsers vs. Attesters. Both are representations of **actual state**. So, its technically possible. The use case for allowing it is when there are claims that can't reasonably be asserted by an AE that a vendor of that Attester wishes to supply reference values. A contrived example is if a CC certifying lab reports that an Attester meets EAL4+, but the Attester vendor wants to restrict the upper bound (for some reason), then the attester vendor could create a RV for a range, say EAL2+ to EAL4. The RV condition applies to claims that only an endorser would make. But the Verifier can easily process them just like RVs written to match Evidence.

The concern is a constrained verifier might not have resources necessary to handle a large set of inputs and one way to reduce them is to disallow processing of RVs aimed at matching Endorsements.

nedmsmith: > Another triple?

It would be clearer to mention a specific endorsement triple such as `endorsed-triples`

nedmsmith (commented):
see various comments and suggestions

draft-ietf-rats-corim.md
```diff
@@ -771,26 +779,24 @@ The types defined for a group identified are UUID and variable-length opaque byt
...
-configuration files, read-only memory, fuses, IO ring configuration, partial
-reconfiguration regions, etc. Measurements comprise raw values, digests, or
-status information.
+A measurement can be directly compared with attestation evidence, or it can be a quality that is not directly comparable with evidence.
+The scope of a mroperty is defined by the `environment-map` and optional `mkey` it appears with.
 
-An environment has one or more measurable elements. Each element can have a
-dedicated measurement or multiple elements could be combined into a single
-measurement. Measurements can have class, instance or group scope.  This is
-typically determined by the triple's environment.
```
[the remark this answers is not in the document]
deeglaze: I find lowercase "environment" difficult to map to a single concept. Here's my understanding. An `environment-map` does not contain measurements. An Attesting Environment has measurable elements. A Target Environment has reference values for elements.

I reword or reformat the rest of these sentences below. I don't like "typically" here since measurements always are within the scope on an environment-map.

draft-ietf-rats-corim.md
```diff
@@ -809,7 +815,8 @@ The following describes each member of the `measurement-map`:
 
 ###### Measurement Keys {#sec-comid-mkey}
 
-The types defined for a measurement identifier are OID, UUID or uint.
+A measurement-associated (sub-)environment is at the granularity of a single element, such as the second enumerated network interface card.
```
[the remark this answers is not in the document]
deeglaze: I would like to, but given the existing use of "measured-element-type-choice" to refer to those logically composed objects, I'd like to know what word we want to use that's not the "element" from the CDDL spec. Do we want to say measured-object?

draft-ietf-rats-corim.md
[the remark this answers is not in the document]
deeglaze: Yes, fixed.

draft-ietf-rats-corim.md
```diff
@@ -771,26 +779,24 @@ The types defined for a group identified are UUID and variable-length opaque byt
...
+Measurements can be of a variety of things including software, firmware, configuration files, read-only memory, fuses, IO ring configuration, partial reconfiguration regions, etc. Measurements comprise raw values, digests, or status information.
 
-Class measurements apply generally to all the Attesters in the given class.
-Instance measurements apply to a specific Attester instance.  Environments
-identified by a class identifier have measurements that are common to the
-class. Environments identified by an instance identifier have measurements that
-are specific to that instance.
+Each element can have a dedicated measurement/endorsement value or multiple elements could be combined into a single measurement.
+An `environment-map` contains at least one of: class, instance, or group.
+The more elements in the `environment-map`, the narrower the scope, since more elements must be in common to match.
```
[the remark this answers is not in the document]
deeglaze: Moved.

draft-ietf-rats-corim.md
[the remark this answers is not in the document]
nedmsmith: An environment (aka Target Environment) is the thing that is being measured by an Attesting Environment. The name of the environment is expressed as key-value pairs (which is what environment-map and stateful-environment-record do). An AE collects measurements about the TE and those measurements are expressed as key-value pairs. The measurement-value-map structure does this. 

We can change "environment" to use TE if lower case environment is confusing.
I'm OK with removing "typically".

draft-ietf-rats-corim.md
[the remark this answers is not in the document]
nedmsmith: I suppose everyone has different intuition around what words to use to describe compositions of things. Using RFC9334 as a guide, it uses environment (TE/AE) and "sub-Attester" which is another way of saying some sort of construction of TEs and AEs, but it didn't use "sub-TE" or "sub-AE" and some would find this awkward. 

The current wording reflects the idea that a TE can be broken into finer granularity sub-parts and the English dictionary definition of "element" is a word that isn't too overloaded.
If a lay person described a computing environment as something that consisted of several sub-elements it wouldn't be too far from reality while still being abstract enough to fit a wide variety of implementations.

draft-ietf-rats-corim.md
```diff
@@ -809,7 +815,8 @@ The following describes each member of the `measurement-map`:
 
 ###### Measurement Keys {#sec-comid-mkey}
 
-The types defined for a measurement identifier are OID, UUID or uint.
+A measurement-associated (sub-)environment is at the granularity of a single element, such as the second enumerated network interface card.
+The types of this measured element identifier are OID, UUID or uint.
```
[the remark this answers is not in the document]
nedmsmith: Since there is a pending discussion around whether or not we want to key mkey. We should hold off on making changes to mkey related content until that discussion is resolved.

draft-ietf-rats-corim.md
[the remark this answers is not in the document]
deeglaze: I say (sub-)enviroment consistently with the existing text for `mkey`. Notion of it being a sub-environment is that it is further limiting the scope of the subject named by the `environment-map`, but is not itself an element of `environment-map`.

I changed from "element" to "object-id" since I read that you wanted element to consistently mean the same thing, i.e., the meaning from CDDL. Are you suggesting that "element" is not too overloaded as an English term, so we _don't_ have to use "element" to consistently refer to the CDDL concept? I'm confused what you want.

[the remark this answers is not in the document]
nedmsmith: >I'm confused what you want.

I'm suggesting that before making changes to CDDL and spec text as it pertains to `mkey`, we should resolve the meta conversation of whether mkey is going to be removed. 
Issue #70 was tracking a use case that possibly justifies its use. The other use cases (that I'm aware of) are profile specific. 

There was a thread related to `mkey` as part of the internal representation conversation that resulted in the following CDDL being included:
```
ECT = {
  ? e: environment-map
  ? c: claims-map / [ + local-claim ]
  ? a: [ + $crypto-key-type-choice ]
  ? ns: text
  ? cm: cm-type
  ? p: $profile-type-choice
}
local-claim = {
  le: local-environment
  c: claims-map
}
local-environment =  bstr / tstr
```
The strongest reason to remove `mkey` is its another form of an ECT structure (that is awkwardly nested in the Claim part of a "parent?" ECT). Since we clarified that the identifier scope for environment values in `environment-map` is global. If an `mkey` has global scope, another ECT could be used instead.  

So, the salient benefit that `mkey` brings is some syntactic sugar when `mkey` identifiers are locally scoped (relative to the parent ECT) to make RIMs less verbose. Instead of having to define another OID or UUID to disambiguate multiple measurements of the same kind (i.e., multiple entries with the same measurement-values-map code point). A local scope identifier, such as a number or string, is used.

But we need to have this conversation in the context of a use case (e.g., issue #70). So, lets prioritize #70 above #248.

draft-ietf-rats-corim.md
```diff
@@ -706,6 +706,9 @@ The following describes each member of the `environment-map`:
 * `group` (index 2): identifier for a group of instances, e.g., if an
   anonymization scheme is used.
 
+An `environment-map` contains at least one of `class`, `instance`, or `group`.
+The more elements in the `environment-map`, the narrower the scope, since more elements must be in common to match.
```
yogeshbdeshpande, line 710:
```suggestion
```

thomas-fossati (approved):

yogeshbdeshpande:
Not adequate to progress this hence closing as per @deeglaze suggestion.

PR #265, opened 2024-09-04
Proposal to add multiplicity of Measurement Maps

yogeshbdeshpande:
Modern day attesters have Multiple Measured Elements, within a Target Environment. This PR intends to allow multiple measured elements to be expressed using a single Reference Value Triple.

### DIFF
--- a/cddl/measurement-maps-type-choice.cddl
+++ b/cddl/measurement-maps-type-choice.cddl
@@ -1,0 +1,1 @@
+measurement-maps-type-choice = measurement-map / #6.533([ measurement-map ])
--- a/cddl/corim-frags.mk
+++ b/cddl/corim-frags.mk
@@ -47,6 +47,7 @@ COMID_FRAGS += version-map.cddl
 COMID_FRAGS += digest.cddl
 COMID_FRAGS += integrity-registers.cddl
 COMID_FRAGS += concise-swid-tag.cddl
+COMID_FRAGS += measurement-maps-type-choice.cddl
 
 COMID_EXAMPLES := $(wildcard examples/comid-*.diag)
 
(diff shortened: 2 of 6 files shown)


yogeshbdeshpande:
This is a slight modification to already closed PR #181 which introduces `backward compatibility` with existing CDDL in Rev-01 to Rev-5 of the specification and also the Base TCG Specification

yogeshbdeshpande:
@deeglaze If you feel al-right, feel free to make more commits on the branch!

cddl/measurement-maps-type-choice.cddl
deeglaze, line 1: What is the meaning of this tag? I'm not sure why you need one when arrays and maps have different CBOR major type tags.

deeglaze:
> @deeglaze If you feel al-right, feel free to make more commits on the branch!

I don't have push rights to this branch, so I have suggestions in my own commit on top in https://github.com/ietf-rats-wg/draft-ietf-rats-corim/compare/multiplicity...deeglaze:draft-ietf-rats-corim:multiplicity

nedmsmith (commented):
A straightforward example is added to the tcg-compatible branch https://github.com/ietf-rats-wg/draft-ietf-rats-corim/tree/tcg-compatible

It includes examples that are backward compatible with the TCG specs. It doesn't seem to need a CBOR tag.

deeglaze:
Agreed it doesn't need the tag, but I do believe that conditional endorsements need to be updated to describe multiple mkeys at a time.

nedmsmith:
> Agreed it doesn't need the tag, but I do believe that conditional endorsements need to be updated to describe multiple mkeys at a time.

Conditional endorsements were an IETF invention. Hence, no backwards compatibility concerns. The conditional endorsement triple was already updated to add multiplicity. The conditional series triple has the needed multiplicity properties already (at least I haven't had any requests to change it from the people I work with).

nedmsmith:
Contrast with #268 
Maybe we should close this PR?

PR #276, opened 2024-09-18
Impact of Multiple Measured Elements to Conditional Triples

yogeshbdeshpande:
Extending the Compatibility PR to other triples. Once we have a consensus, this PR will be deleted, it will be either merged into
 PR #268

Issue #244, opened 2024-06-19
authorized-by is a security flaw


deeglaze:
I don't think it's appropriate at all for X to sign a document that Y is the one that actually authorized the claims. There is no signature by Y. This is a clear spoofing problem. Even if the Verifier is supposedly forwarding an authorization that it had previously seen by including it in the CoRIM, I don't think that Y has delegated that authority to the Verifier. The Verifier would need to use a "hearsay" claim that it had already shown that Y authorized the claim.

nedmsmith:
> X to sign a document that Y is the one that actually authorized the claims

This isn't the intended semantics of `authorized-by`. The motivating use case is entity A wants to specify whether entity B is authorized to say X (where X is a set of claims aka an ECT). The logic is "A says B says X".  Note, it is still necessary for B to say X. `authorized-by` is therefore a matching condition (as opposed to an addition).

We think AEs are not likely to need this, so it is excluded from Evidence. 

But it is reasonable that X could be applied to RVs or Endorsements. 

The suggested change: 
"remove authorized-by from measurement-map and add an optional ‘authorities’ entry to `stateful-environment-record`"

would have the same semantics as `authorized-by` only it restricts its use to Endorsements as `stateful-environment-record` is only used by Endorsements. To make a symmetric change to the CDDL `reference-triple-record` should be updated too.
```
reference-triple-record = [
  environment-map
  measurement-map
 ? authority: [ + $crypto-key-type-choice ]
]
```
In both structures, the authority is used as a matching condition that determines if B did in fact assert X. In the case above, authority and environment-map are the matching condition, and environment-map and measurement-map are the addition (which is added under the authority of the RIM signer - A). 

An internal representation is supposed to keep track of the dynamic that A says B says X. 

I don't think the current internal representation using ECTs is sufficient to represent this dynamic. We could modify ECT so that there is an outer authority. I'll use `policy` to explain what I mean:
Given:
```
policy = [
    condition: [ + ECT ]
    addition: [ + ECT ]
]
```
By defining a nested ECT:
```
nested-ECT = [
  ECT
  ? a: [ + $crypto-key-type-choice ]
  ? cm: cm-type
]
```
Policy can be changed to:
```
policy = [
    condition: [ + ECT ]
    addition: [ + ECT / nested-ECT ]
]
```
This would allow expressions such as:
```
/cbor diag example/
[
  [
    e: 'x0',
    c: 'x1',
    a: B,
[...]

PR #277, opened 2024-09-18
Account for multiplicity in conditional endorsements

deeglaze:
A conditional endorsement series is about a target environment plus additional state expectations that lead to further endorsements.

I removed the undefined CBOR tag around multiple measuremnt-maps since array and map have different CBOR major tags and a custom tag is not required.

### DIFF
--- a/cddl/conditional-series-record.cddl
+++ b/cddl/conditional-series-record.cddl
@@ -1,6 +1,6 @@
 conditional-series-record = [
-  ; reference values to be matched against evidence
-  refv: measurement-values-map
+  ; reference values to be matched against evidence in each mkey
+  refv: [ + measurement-map ]
   ; endorsed values that apply in case revf matches
-  endv: measurement-values-map
+  endv: [ + mkeyvalue-pair ]
 ]
--- a/cddl/measurement-maps-type-choice.cddl
+++ b/cddl/measurement-maps-type-choice.cddl
@@ -1,1 +1,1 @@
-measurement-maps-type-choice = measurement-map / #6.533([ measurement-map ])
+measurement-maps-type-choice = measurement-map / [ + measurement-map ]
--- a/cddl/corim-frags.mk
+++ b/cddl/corim-frags.mk
@@ -47,7 +47,6 @@ COMID_FRAGS += version-map.cddl
 COMID_FRAGS += digest.cddl
 COMID_FRAGS += integrity-registers.cddl
 COMID_FRAGS += concise-swid-tag.cddl
-COMID_FRAGS += measurement-maps-type-choice.cddl
 
 COMID_EXAMPLES := $(wildcard examples/comid-*.diag)
 
--- a/cddl/endorsed-triple-record.cddl
+++ b/cddl/endorsed-triple-record.cddl
@@ -1,4 +1,4 @@
 endorsed-triple-record = [
   environment-map
-  measurement-maps-type-choice
+  [ + measurement-map ]
 ]
--- a/cddl/reference-triple-record.cddl
+++ b/cddl/reference-triple-record.cddl
@@ -1,4 +1,4 @@
 reference-triple-record = [
   environment-map
-  measurement-maps-type-choice
+  [ + measurement-map ]
 ]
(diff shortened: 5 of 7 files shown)


cddl/conditional-series-record.cddl
```diff
@@ -1,6 +1,6 @@
 conditional-series-record = [
-  ; reference values to be matched against evidence
-  refv: measurement-values-map
+  ; reference values to be matched against evidence in each mkey
+  refv: [ + measurement-map ]
   ; endorsed values that apply in case revf matches
-  endv: measurement-values-map
+  endv: [ + mkeyvalue-pair ]
```
yogeshbdeshpande, line 5:
```suggestion
  endv: [ + measurement-map]
```

cddl/measurement-maps-type-choice.cddl
```diff
@@ -1 +1 @@
-measurement-maps-type-choice = measurement-map / #6.533([ measurement-map ])
\ No newline at end of file
+measurement-maps-type-choice = measurement-map / [ + measurement-map ]
```
yogeshbdeshpande, line 1: We agreed with Ned's PR that is optionality is not required!

yogeshbdeshpande:
@deeglaze Some more work needs to be done! Also, I already floated PR #276 , please have a look and comment. I am happy to fold this one on PR #276 and proceed with just one copy rather than multiple identical `intent`  PR's

cddl/conditional-series-record.cddl
deeglaze: We don't want to have authorized-by in the endorsed values, since it is not meaningful here.

yogeshbdeshpande: It is not true, Endorsements or Reference Values, all should support Authority!

deeglaze: No, authority in a measurement-map in a CoRIM represents a condition. Authority in a measurement-map in an ECT represents actual authorities that signed the information. For any _added_ measurement, the only authority is the CoRIM issuer. Given that the triple has already given two other positions for stating an authority condition, it doesn't make sense for the endv to have its own conditionality semantics.

nedmsmith:
Is there an issue that this PR links to so that reviewers can know the problem statement that motivates this PR?

If we think there are use cases that can't be supported by the current triple it would be helpful to characterize them.

If the goal is to create a series triple that is more generalized, then we should start with the internal representation.
```
evs = [
  condition: [ + ECT ]
  series: + {
    selection: [ + ECT ]
    addition: [ + ECT ]
  }
]
```
Since the ECT can be mapped to the external representation readily, the CDDL for a revised series triple should be relatively straight forward.

cddl/conditional-series-record.cddl
yogeshbdeshpande: I understand your point, however, this has a deeper implications.

By not providing a source of Authority for Added Endorsements in the triple, has implications.

It means the supplier of Endorsements has to have an implicit authority of CoRIM Signer.

It cannot inherit the Authority from RefValue provider Authority has they may not be the same!

Which implies, that for a Env with ClassID : X, with FW running from Authority coming from FW Supplier PQR (in RefVal), the

Test House is vouching for its Endorsements.

It is fine to have CoRIM as a single source of Authority in this case, in most of the cases, so we can start with that.

I was thinking as Authority is `optional` retaining it will provide, the flexibility to introduce endorsements coming from different authority, For an Environement, like for FW Element -1 : Authority X, FW Element 2: Authority Y

deeglaze: The only integrity protection you can ensure is who signed the information. https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/244 affirms that authorized-by is a condition, and not part of an addition. If it's not part of the condition and is actual addition, we need to have a triple signed by the delegate that the CoRIM issuer is authorized to sign on its behalf. I don't think we should do that though. You generally do that with PKI certificate paths.

Please express your example in terms of the conditional series triple so I may understand.

yogeshbdeshpande:
issue #288 tracks the discussion!

deeglaze:
Closing since there are other PRs covering this.

PR #268, opened 2024-09-11
Revert endorsement and reference triple records to original

nedmsmith:
This PR reverts `endorsement-triple-record` and `reference-triple-record` to original CDDL by adding array brackets around `measurement-map` to allow multiplicity of measurements for a given environment.

The Measurements and Measurement Keys sections are updated to better describe measurement multiplicity and disambiguation of measurement values.

PR #284, opened 2024-09-19
ensure internal representation is valid CDDL

### LINKED ISSUE #282: Internal representation CDDL needs CI integration
Since the internal representation should be valid CDDL, it needs CI integration and code snippets should be imported into the document.

thomas-fossati:
Add automatic validation of internal representation CDDL

Fix #282


cddl/corim-frags.mk
```diff
@@ -76,3 +76,19 @@ CORIM_FRAGS += validity-map.cddl
 CORIM_FRAGS += $(COMID_FRAGS)
 
 CORIM_EXAMPLES := $(wildcard examples/corim-*.diag)
+
+INTREP_FRAGS := intrep-acs.cddl
```
yogeshbdeshpande, line 80: Just a minor nit,

The CDDL names in our repo are long but quite expressive, for example:
`measured-element-type-choice.cddl`

I would request, we rename INTREP everywhere to 

`INTERNAL-REP`

So the CDDL becomes: internal-rep-claims-map.cddl

This avoids any confusion in the naming to mis-interpret it as Integer Represenation!

yogeshbdeshpande (approved):
LGTM!, just left one minor Nit to look at!

cddl/corim-frags.mk
thomas-fossati: > This avoids any confusion in the naming to mis-interpret it as Integer Representation!

No way, that's a feature, not a bug :-) I made it expressly to sound like INTREPid!

yogeshbdeshpande: But it is not INTREPID after-all ??

nedmsmith:
Are you opposed to using the version that is in PR #268 which is better aligned to the existing wording? 
~~~ cddl
ECT = {
  ? environment: environment-map
  ? properties: [ + properties-map ]
  ? authority: [ + $crypto-key-type-choice ]
  ? namespace: text
  ? cmtype: cm-type
}
properties-map = {
  ? eid: element-id
  claims: measurement-values-map
}
element-id =  $measured-element-type-choice
cm-type =  &(
  reference-values: 0
  endorsements: 1
  evidence: 2
  attestation-results: 3
  verifier: 4
  policy: 5
)
~~~

thomas-fossati:
> Are you opposed to using the version that is in PR #268 which is better aligned to the existing wording (sans namespace)?

Content-wise, I am OK with anything.  The intent of this PR was purely mechanical, but if you want me to put in a more meaningful CDDL, I can certainly do that.  Or we can do it in #268 after rebase.  Let me know what's better from your perspective.

nedmsmith:
> > Are you opposed to using the version that is in PR #268 which is better aligned to the existing wording (sans namespace)?
> 
> Content-wise, I am OK with anything. The intent of this PR was purely mechanical, but if you want me to put in a more meaningful CDDL, I can certainly do that. Or we can do it in #268 after rebase. Let me know what's better from your perspective.

Since the CDDL blocks will all be replaced with imported CDDL, I'll focus on prose changes that reflect the environment:element construct. (see revised CDDL example in the previous comment).

thomas-fossati:
> Since the CDDL blocks will all be replaced with imported CDDL, I'll focus on prose changes that reflect the environment:element construct. (see revised CDDL example in the previous comment).

OK, I'll use the new CDDL here then.

thomas-fossati:
> OK, I'll use the new CDDL here then.

Done in [0d82b15](https://github.com/ietf-rats-wg/draft-ietf-rats-corim/pull/284/commits/0d82b15fffd8cef5d09f4747e5e45bed8dd1caf6)

cddl/intrep-ect.cddl
```diff
@@ -0,0 +1,21 @@
+ECT = {
+  ? environment: environment-map
+  ? element-list: [ + element-map ]
+  ? authority: [ + $crypto-key-type-choice ]
+  ? cmtype: cm-type
```
thomas-fossati, line 5:
```suggestion
  cmtype: cm-type
```

cddl/intrep-ev.cddl
```diff
@@ -0,0 +1,12 @@
+ev = [
```
deeglaze, line 1: It seems endorsements, conditional endorsements, and conditional series endorsements are unrolling a recursive definition that would allow for simpler expression.

```
evs = [ + {
? / condition: / 0 => [ + ECT ]
/ addition: / 1 => evs / [ + ECT ]
} ]
```
If a condition applies and the addition isn't a map, add the ECTs. If the addition is a map, re-enter the top of the loop to further specialize the condition. If the condition doesn't apply, go to the next map in the series.

This allows for nested if/elseif.../else expressions. I don't think the nonempty selection requirement allows for a bare "else".

cddl/intrep-ev.cddl
```diff
@@ -0,0 +1,12 @@
...
+  condition: [ + ECT ]
+  addition: [ + ECT ]
+]
+
+evs = [
+  condition: [ + ECT ]
+  series: [ + {
```
deeglaze, line 8: This map syntax for the pair needs key definitions, no? `/ selection: / 0 => [ + ECT ]` for example. Same comment for rv below.

cddl/intrep-ev.cddl
deeglaze: No, [ + ECT ] is a base case. You can unfold as many nestings as you want, but you eventually bottom out in a flat list of additions.

cddl/intrep-ev.cddl
thomas-fossati: It's good like that.  We decided to avoid codepoint aliases here and just use string labels because compression is not needed.

cddl/intrep-ev.cddl
thomas-fossati: @deeglaze clarifying question: your `evs` definition above would also subsume `ev`?

cddl/intrep-ev.cddl
nedmsmith: Keep in mind that a primary goal of defining an internal representation is to facilitate readability of document prose. It isn't to arrive at the most optimal CDDL. 

Recursive structures are harder to describe IMHO.  The purpose for the 'series: ' label is to make it easier to describe the processing differences (which uses OR semantics where the first match terminates processing) from the 'addition:' (which uses AND semantics.)

Combining ev and evs with nesting seems like its going in the wrong direction.

Actual implementations can always apply these sorts of optimizations if they want to.

PR #289, opened 2024-09-23
Rework ECT comparison section to match new internal representation

andrew-draper:
This is the first edit to section 8.9 - it aligns the terminology with the change to ECT internal representation

It also defines the comparison rules for raw-value (codepoint 4 and 5).

### DIFF
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -1932,7 +1932,7 @@ Reference Value Providers (RVP) publish Reference Values triples that are matche
 Reference Values may describe multiple acceptable states for Attesters; hence "matching" determines that Evidence (contained in the ACS) satisfies an appropriate subset of the available Reference Values.
 If the appropriate subset matches, the authority of the RVP is added to the appropriate ACS entries.
 
-The Verifier compares each `reference-triple-record` against ACS entries as described in {{sec-match-one-se}}, where the `reference-triple-record` takes the place of a `stateful-environment-record`.
+The Verifier compares each `reference-triple-record` against ACS entries as described in {{sec-match-condition-ect}}, where the `reference-triple-record` takes the place of a `stateful-environment-record`.
 If all fields of the `reference-triple-record` match the ACS, then the Verifier MUST add the RVP authority to each matching ACS field.
 
 If any `reference-triple-record` in the Reference Value triple does not match the ACS then the entire triple is ignored.
@@ -1952,7 +1952,7 @@ The verifier checks whether Conditional Endorsements are applicable by comparing
 
 #### Processing Conditional Endorsement Triple
 
-For each Conditional Endorsement Triple the Verifier compares each of the `stateful-environment-record` fields from the `cond` field in the triple against the ACS (see {{sec-match-one-se}}).
+For each Conditional Endorsement Triple the Verifier compares each of the `stateful-environment-record` fields from the `cond` field in the triple against the ACS (see {{sec-match-condition-ect}}).
 
 If every stateful environment matches a corresponding ACS entry, then the Verifier MUST add an Endorsement entry to the ACS (see {{sec-add-to-acs}}) for each `endorsed-triple-record` in the `endorsements` field.
 Each Endorsement from the `endorsed-triple-record` includes the authority which signed the Conditional Endorsement Triple.
@@ -1961,36 +1961,10 @@ Each Endorsement from the `endorsed-triple-record` includes the authority which 
 
 For each Conditional Endorsement Series Triple the Verifier iterates over the `conditional-series-record`s within the triple, stopping if it finds a match.
 
-For each iteration, the Verifier creates a temporary `stateful-environment-record` by merging the `stateful-environment-record` in the triple with the `refv` field in the `conditional-series-record`. It compares this temporary record against the ACS (see {{sec-match-one-se}}).
+For each iteration, the Verifier creates a temporary `stateful-environment-record` by merging the `stateful-environment-record` in the triple with the `refv` field in the `conditional-series-record`. It compares this temporary record against the ACS (see {{sec-match-condition-ect}}).
 
 If one of the temporary records matches then the Verifier MUST add the `endv` Endorsement entry to the ACS.
 This Endorsement includes the authority which signed the Conditional Endorsement Series Triple.
-
-#### Processing a stateful environment against the Appraisal Claims Set {#sec-match-one-se}
-
-This section describes how a stateful environment is matched against an ACS entry.
-If any part of the processing indicates that the stateful environment does not match then the remaining steps in this section are skipped for that stateful environment.
-
-The Verifier initializes a temporary "candidate entries" variable with all entries in the ACS where the stateful enviromnment `environment-map` is a subset of the ACS `environment-map`.
-
-A stateful environment `environment-map` is a subset of an ACS entry `environment-map` if each field (for example `class`, `instance` etc.) which is present in the stateful environment `environment-map` is also present in the ACS entry, and the CBOR encoded field values in the stateful environment and ACS entry are binary identical.
-If a field is not present in the stateful environment `environment-map` then the presence of, and value of, the corresponding ACS entry field does not affect whether the `environment-map`s are subsets.
-
-Before performing the binary comparison, a Verifier SHOULD convert `environment-map` fields into a form which meets CBOR Core Deterministic Encoding Requirements {{-cbor}}.
-
-If the stateful environment contains an `authorized-by` field then the Verifier SHALL remove all candidate entries whose `authorized-by` field does not contain one of the keys listed in the stateful environment `authorized-by` field (see {{sec-authorized-by}} for more details).
-
-If there are no candidate entries then the triple containing the stateful environment does not match.
-
-The stateful environment entry is compared against each of the candidate entries.
-
-For each of the candidate entries, the Verifier SHALL iterate over the codepoints which are present in the `measurement-values-map` field within the stateful environment `measurement-map`.
-Each of the codepoints present in the stateful environment is compared against the candidate entry.
-
-If any codepoint present in the stateful environment `measurement-values-map` does not match the same codepoint within the candidate entry `measurement-values-map` then the stateful environment does not match.
-
-If all checks above have been performed successfully then the stateful environment matches.
-If none of the candidate entries match the stateful environment entry then the stateful environment does not match.
 
 ## Examples for optional phases 5, 6, and 7 {#sec-phases567}
 
@@ -2089,31 +2063,66 @@ Logically, new ECT entries are appended to the existing ACS.
 But implementations may optimize ECT order to achieve better performance.
 Additions to the ACS MUST be atomic.
 
-## ECT Comparison {#sec-ect-compare}
+## Comparing a condition ECT against the ACS {#sec-match-condition-ect}
 
 [^issue] https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/71
 
-This specification defines the comparison algorithm for the codepoints described in sub-sections below.
-A CoRIM profile may define additional negative codepoints and their matching algorithms.
-Specifications that extend CoMID MUST also define comparison algorithms for their added codepoints.
-If a codepoint's comparison algorithm is not stated or does not default to the comparison algorithms described in this specification, then the Verifier MUST compare the binary equality of the CBOR encodings of the values.
+The algorithm used to compare a condition ECT against the ACS is stateless, it depends only on the condition ECT being compared and the contents of the ACS at the time of the comparison
+
+The Verifier uses a temporary "candidate entries" array as part of the comparision. Each element in the array is an ECT.
+
+After completing the steps described in this sub-section, the algorithm returns false if the candidate entries array is empty, true if it contains one or more elements.
 
 ### Environment Comparison {#sec-compare-env}
 
+The verifier SHALL iterate over all entries in the ACS and SHALL add each entry where the condition ECT stateful enviromnment `environment-map` is a subset of the ACS entry `environment-map` to the candidate entries array.
+
+The condition ECT `environment-map` is a subset of an ACS entry `environment-map` if each field (for example `class`, `instance` etc.) which is present in the condition ECT `environment-map` is also present in the ACS entry, and the CBOR encoded field values in the stateful environment and ACS entry are binary identical.
+If a field is not present in the condition ECT `environment-map` then the presence of, and value of, the corresponding ACS entry field does not affect whether the `environment-map`s are subsets.
+
+Before performing the binary comparison, a Verifier SHOULD convert `environment-map` fields into a form which meets CBOR Core Deterministic Encoding Requirements {{-cbor}}.
+
+### Mkey comparison {#sec-compare-mkey}
+
+The verifier SHALL iterate over all entries in the candidate entries array and compare the condition ECT `measurement-map/mkey` value to the candidate entry `measurement-map/mkey` value.
+If the two `mkey` values are not binary equal then the candidate entry is removed from the candidate entries array.
+
+Before performing the binary comparison, a Verifier SHOULD convert `mkey` fields into a form which meets CBOR Core Deterministic Encoding Requirements {{-cbor}}.
+
+If the `mkey` field is not present in neither the condition ECT nor the candidate entry then the `mkey` are equal and the candidate entry is not removed.
+If the `mkey` field is not present in only one side of the comparison then the `mkey` are not equal and the candiate entry is removed.
+
+### Authority comparison
+
+The verifier SHALL iterate over all entries in the candidate entries array and compare the condition ECT expected authority (`a`) value to the candidate entry authority value.
+
+If the condition ECT contains an `a` field then the Verifier SHALL remove all candidate entries whose `authorized-by` field does not contain one of the keys listed in the condition ECT `authorized-by` field (see {{sec-authorized-by}} for more details).
+
+When comparing two `$crypto-key-type-choice` fields for equality, the verifier SHALL treat then as equal if their CBOR encoding is binary equal.
+
+The verifier MAY treat two keys as equal if they have different formats but represent the same key.
+For example, the verifier may contain code to compare a key fingerprint against the key which, when hashed, creates that fingerprint.
+
 ### Claims Comparison  {#sec-compare-claims}
 
-#### Comparison of measurement-values-map {#sec-match-one-codepoint}
+The verifier SHALL iterate over all entries in the candidate entries array and compare the condition ECT claims to the candidate entry authority claims.
 
-This section describes the algorithm used to compare the `measurement-values-map` codepoints of an ECT with another ECT.
-The comparison algorithm performed depends on the value of the codepoint being compared.
+The Verifier SHALL iterate over the codepoints which are present in the `measurement-values-map` field within the condition ECT `measurement-map`.
+Each of the codepoints present in the condition ECT is compared against the same codepoint in the candidate entry.
 
-[^issue] https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/203
+If any codepoint present in the condition ECT `measurement-values-map` does not have a corresponding codepoint within the candidate entry `measurement-values-map` then verifier SHALL remove that candidate entry from the candidate entries array.
 
-If the `measurement-values-map` value has an associated CBOR tag, the comparison algorithm should comprehend the structure identified by the CBOR tag.
+If any codepoint present in the condition ECT `measurement-values-map` does not match the same codepoint within the candidate entry `measurement-values-map` then verifier SHALL remove that candidate entry from the candidate entries array.
 
-If the Verifier does not recognize a CBOR tag value then the value MUST NOT match.
+#### Comparison of a single measurement-values-map codepoint{#sec-match-one-codepoint}
 
-Note: CBOR tags are useful for discriminating values amongst alternates, but the comparison of tagged values is still determined by the codepoint they appear under. It is recommended but not required to compare values with a specific CBOR tag the same way across codepoints. For this reason, it is recommended to specify a default comparison algorithm with the CBOR tag's registration.
+If the codepoint key is zero or positive then the Verifier SHALL use a standard comparison algorithm, defined in this document, to compare the claims.
+The codepoint number and, if present, the tag value associated with the condition ECT codepoint SHALL be used to select the algorithm.
+
+If the codepoint key is negative then the Verifier MAY use a comparison algorithm defined by the appropriate CoRIM profile to compare the claims.
+The codepoint number, the profile associated with the condition ECT, and the tag value (if present) shall be used to select the algorithm.
+
+If the verifier is unable to determine the comparison algorithm which applies to a codepoint then it SHALL behave as though the candidate entry does not match the condition ECT.
 
 Profile writers SHOULD use CBOR tags for widely applicable comparison methods to ease Verifier implementation compliance across profiles.
 
@@ -2121,80 +2130,72 @@ The following subsections define the comparison algorithms for the `measurement-
 
 ##### Comparison for svn entries
 
-The value stored under `measurement-values-map` key 1 is an SVN, which must
-have type UINT.
+The value stored under `measurement-values-map` key 1 is an SVN, which must have type UINT.
 
-If the Reference value for `measurement-values-map` key 1 is an untagged UINT or
-a UINT tagged with #6.552 then an equality comparison is performed. If the value
-of the SVN in ACS is not equal to the value in the Reference
-Value then the Reference Value does not match.
+If the condition ECT value for `measurement-values-map` key 1 is an untagged UINT or a UINT tagged with #6.552 then an equality comparison is performed.
+The comparison MUST return true if the value of the SVN in the candidate entry is equal to the value in the condition ECT.
 
-If the Reference value for `measurement-values-map` key 1 is a UINT tagged with
-#6.553 then a minimum comparison is performed. If the value of the SVN in
-ACS less than the value in the Reference Value then the
-Reference Value does not match.
+If the condition ECT value for `measurement-values-map` key 1 is a UINT tagged with #6.553 then a minimum comparison is performed.
+The comparison MUST return true if the value of the SVN in the candidate entry is less than the value in the condition ECT.
 
 ##### Comparison for digests entries {#sec-cmp-digests}
 
-The value stored under `measurement-values-map` key 2,
-or a value tagged with
-#6.TBD is a digest entry.
-It contains one or more digests, each measuring the
-same object. A Reference Value may contain multiple digests, each with a
-different algorithm acceptable to the Reference Value provider. If the
-digest in Evidence contains a single value with an algorithm and value
-matching one of the algorithms and values in the Reference Value then it
-matches.
+The value stored under `measurement-values-map` key 2 is a digest entry.
+It contains one or more digests, each measuring the same object.
+A condition ECT may contain multiple digests, each with a different algorithm acceptable to the condition ECT author.
 
-To prevent downgrade attacks, if there are multiple algorithms which are in
-both the Evidence and Reference Value then the digests calculated using all
-shared algorithms must match.
+In the simple case, a condition ECT digests entry containing one digest matches matches a candidate entry containing a single entry with the same algorithm and value.
 
-If the CBOR encoding of the `digests` entry in the Reference Value or the
-ACS value with the same key is incorrect (for example if fields
-are missing or the wrong type) then the Reference Value does not match.
+To prevent downgrade attacks, if there are multiple algorithms which are in both the condition ECT and candidate entry then the digests calculated using all shared algorithms must match.
 
-The Verifier MUST iterate over the Reference Value `digests` array, locating
-hash algorithm identifiers that are present in the Reference Value and
-in the ACS entry.
+The comparison MUST return false if the CBOR encoding of the `digests` entry in the condition ECT or the ACS value with the same key is incorrect (for example if fields are missing or the wrong type) then the Reference Value does not match.
 
-If the hash algorithm identifier which is present in the Reference Value
-differs from the hash algorithm identifier in the ACS entry then the Reference Value does not match.
+The comparison MUST return false if the condition ECT digests entry does not contain any digests.
 
-If a hash algorithm identifier is present in both the Reference Value and
-the ACS, but the value of the hash is not binary identical
-between the Reference Value and the ACS entry then the
-Reference Value does not match.
+The comparison MUST return false if either digests entry contains multiple values for the same hash algorithm.
+
+The Verifier MUST iterate over the condition ECT `digests` array, locating common hash algorithm identifiers (which are present in the condition ECT and in the candidate entry).
+If the value associated with any common hash algorithm identifier in the condition ECT differs from the value for the same algorithm identifier in the candidate entry then the comparison MUST return false.
 
 ##### Comparison for raw-value entries
 
-> [Andy] *I think this comparison method only works if the entry is at key 4 (because
-there needs to be a mask at key 5). Should we have a Reference Value of this
-which stores `[expect-raw-value raw-value-mask]` in an array?*
+The value stored under `measurement-values-map` key 4 is a raw-value entry, which must have type BSTR.
 
-[^issue] https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/71
+The value stored under the condition ECT `measurement-values-map` key 5 is a raw-value-mask entry, which must have type BSTR.
+
+If the condition ECT does not contain a raw-values-maek then the comparison MUST return true if the condition ECT raw-value is binary equal to the candidate entry raw-value.
+
+If the condition ECT raw-values-mask, condition ECT raw-value and candidate entry raw-value are not the same length then the comparison MUST return false.
+
+If all the lengths are the same then the verifier MUST iterate over the bits in the raw-values-mask which are 1, and compare the corresponding bits in the two raw-value fields.
+
+If, for every bit in the raw-values-mask which is 1, the corresponding bit is the same in both raw-values then the comparison MUST return true.
+
+If any masked in bit differebt between the raw values then the comparison MUST return false.
+
+Note that if a candidate entry contains a value under key 5 then this does not affect the result of the comparison.
 
 ##### Comparison for cryptokeys entries {#sec-cryptokeys-matching}
 
 The value stored under `measurement-values-map` key 12 is an array of `$crypto-key-type-choice` entries. `$crypto-key-type-choice` entries are CBOR tagged values.
 The array contains one or more entries in sequence.
 
-The CBOR tag of the first entry of the Reference Value `cryptokeys` array is compared with
-the CBOR tag of the first entry of the ACS `cryptokeys` value.
-If the CBOR tags match, then the bytes following the CBOR tag from the Reference Value entry
-are compared with the bytes following the CBOR tag from the ACS entry.
+The CBOR tag of the first entry of the condition ECT `cryptokeys` array is compared with
+the CBOR tag of the first entry of the candidate entry `cryptokeys` value.
+If the CBOR tags match, then the bytes following the CBOR tag from the condition ECT entry
+are compared with the bytes following the CBOR tag from the candidate entry.
 If the byte strings match, and there is another array entry,
-then the next entry from the Reference Values array is likewise
+then the next entry from the condition ECTs array is likewise
 compared with the next entry of the ACS array.
-If all entries of the Reference Values array match a corresponding entry in the ACS array, then the `cryptokeys` Reference Value matches.
+If all entries of the condition ECTs array match a corresponding entry in the ACS array, then the `cryptokeys` condition ECT matches.
 Otherwise, `cryptokeys` does not match.
 
 ##### Comparison for Integrity Registers {#sec-cmp-integrity-registers}
 
-For each Integrity Register entry in the Reference Value, the Verifier will use the associated identifier (i.e., `integrity-register-id-type-choice`) to look up the matching Integrity Register entry in Evidence.
-If no entry is found, the Reference Value does not match.
+For each Integrity Register entry in the condition ECT, the Verifier will use the associated identifier (i.e., `integrity-register-id-type-choice`) to look up the matching Integrity Register entry in the candidate entry.
+If no entry is found, the condition MUST return false.
 Instead, if an entry is found, the digest comparison proceeds as defined in {{sec-cmp-digests}} after equivalence has been found according to {{sec-comid-integrity-registers}}.
-Note that it is not required for all the entries in Evidence to be used during matching: the Reference Value could consist of a subset of the device's register space. In TPM parlance, a TPM "quote" may report all PCRs in Evidence, while a Reference Value could describe a subset of PCRs.
+Note that it is not required for all the entries in the candidate entry to be used during matching: the condition ECT could consist of a subset of the device's register space. In TPM parlance, a TPM "quote" may report all PCRs in Evidence, while a condition ECT could describe a subset of PCRs.
 
 ### Authority Comparison  {#sec-compare-auth}
 
@@ -2212,7 +2213,7 @@ The `cm` field comparison tests equality of one or more bits.
 
 ### Profile-directed Comparison {#sec-compare-profile}
 
-A profile may specify handling for new CBOR tagged Reference Values.
+A profile may specify handling for new CBOR tagged condition ECTs.
 The profile must specify how to compare the CBOR tagged Reference Value against the ACS.
 
 Note that the verifier may compare Reference Values in any order, so the comparison should not be stateful.


draft-ietf-rats-corim.md
```diff
@@ -2089,112 +2063,139 @@ Logically, new ECT entries are appended to the existing ACS.
...
+The verifier SHALL iterate over all entries in the ACS and SHALL add each entry where the condition ECT stateful enviromnment `environment-map` is a subset of the ACS entry `environment-map` to the candidate entries array.
+
+The condition ECT `environment-map` is a subset of an ACS entry `environment-map` if each field (for example `class`, `instance` etc.) which is present in the condition ECT `environment-map` is also present in the ACS entry, and the CBOR encoded field values in the stateful environment and ACS entry are binary identical.
+If a field is not present in the condition ECT `environment-map` then the presence of, and value of, the corresponding ACS entry field does not affect whether the `environment-map`s are subsets.
+
+Before performing the binary comparison, a Verifier SHOULD convert `environment-map` fields into a form which meets CBOR Core Deterministic Encoding Requirements {{-cbor}}.
+
+### Mkey comparison {#sec-compare-mkey}
+
+The verifier SHALL iterate over all entries in the candidate entries array and compare the condition ECT `measurement-map/mkey` value to the candidate entry `measurement-map/mkey` value.
```
yogeshbdeshpande, line 2087: Line 2087, should make it very clear which are the list of  `mkey` from the triple and which are list of entries from the ECT.

draft-ietf-rats-corim.md
```diff
@@ -2089,112 +2063,139 @@ Logically, new ECT entries are appended to the existing ACS.
...
+If the two `mkey` values are not binary equal then the candidate entry is removed from the candidate entries array.
+
+Before performing the binary comparison, a Verifier SHOULD convert `mkey` fields into a form which meets CBOR Core Deterministic Encoding Requirements {{-cbor}}.
+
+If the `mkey` field is not present in neither the condition ECT nor the candidate entry then the `mkey` are equal and the candidate entry is not removed.
+If the `mkey` field is not present in only one side of the comparison then the `mkey` are not equal and the candiate entry is removed.
+
+### Authority comparison
+
+The verifier SHALL iterate over all entries in the candidate entries array and compare the condition ECT expected authority (`a`) value to the candidate entry authority value.
```
yogeshbdeshpande, line 2097: Same comment as 2087

draft-ietf-rats-corim.md
```diff
@@ -2089,112 +2063,139 @@ Logically, new ECT entries are appended to the existing ACS.
...
+
+The condition ECT `environment-map` is a subset of an ACS entry `environment-map` if each field (for example `class`, `instance` etc.) which is present in the condition ECT `environment-map` is also present in the ACS entry, and the CBOR encoded field values in the stateful environment and ACS entry are binary identical.
+If a field is not present in the condition ECT `environment-map` then the presence of, and value of, the corresponding ACS entry field does not affect whether the `environment-map`s are subsets.
+
+Before performing the binary comparison, a Verifier SHOULD convert `environment-map` fields into a form which meets CBOR Core Deterministic Encoding Requirements {{-cbor}}.
+
+### Mkey comparison {#sec-compare-mkey}
+
+The verifier SHALL iterate over all entries in the candidate entries array and compare the condition ECT `measurement-map/mkey` value to the candidate entry `measurement-map/mkey` value.
+If the two `mkey` values are not binary equal then the candidate entry is removed from the candidate entries array.
```
deeglaze, line 2088: I don't think so, assuming https://github.com/ietf-rats-wg/draft-ietf-rats-corim/pull/268.

We'd need to iterate over all elements of the ECT's elements and find in the candidate ECT's elements a CBOR encoding-binary-identical mkey before then trying to match the element's mval against the found element's mval. If the candidate ECT does not have such an element, then the candidate ECT does not match.

draft-ietf-rats-corim.md
deeglaze: Right, there is not a single condition authority, but an ANDed list. Each authority must have a matching key in the candidate ECT's authorities.

draft-ietf-rats-corim.md
```diff
@@ -2089,112 +2063,139 @@ Logically, new ECT entries are appended to the existing ACS.
...
+#### Comparison of a single measurement-values-map codepoint{#sec-match-one-codepoint}
 
-If the Verifier does not recognize a CBOR tag value then the value MUST NOT match.
+If the codepoint key is zero or positive then the Verifier SHALL use a standard comparison algorithm, defined in this document, to compare the claims.
+The codepoint number and, if present, the tag value associated with the condition ECT codepoint SHALL be used to select the algorithm.
 
-Note: CBOR tags are useful for discriminating values amongst alternates, but the comparison of tagged values is still determined by the codepoint they appear under. It is recommended but not required to compare values with a specific CBOR tag the same way across codepoints. For this reason, it is recommended to specify a default comparison algorithm with the CBOR tag's registration.
+If the codepoint key is negative then the Verifier MAY use a comparison algorithm defined by the appropriate CoRIM profile to compare the claims.
+The codepoint number, the profile associated with the condition ECT, and the tag value (if present) shall be used to select the algorithm.
```
deeglaze, line 2123:
```suggestion
The codepoint number and the profile associated with the condition ECT may be used to select the algorithm.
```
Again the algorithm is specific to the codepoint and may choose to further dispatch based on value tags.

draft-ietf-rats-corim.md
```diff
@@ -2089,112 +2063,139 @@ Logically, new ECT entries are appended to the existing ACS.
...
-Reference Value does not match.
+The Verifier MUST iterate over the condition ECT `digests` array, locating common hash algorithm identifiers (which are present in the condition ECT and in the candidate entry).
+If the value associated with any common hash algorithm identifier in the condition ECT differs from the value for the same algorithm identifier in the candidate entry then the comparison MUST return false.
 
 ##### Comparison for raw-value entries
 
-> [Andy] *I think this comparison method only works if the entry is at key 4 (because
-there needs to be a mask at key 5). Should we have a Reference Value of this
-which stores `[expect-raw-value raw-value-mask]` in an array?*
+The value stored under `measurement-values-map` key 4 is a raw-value entry, which must have type BSTR.
```
deeglaze, line 2162:
```suggestion
The value stored under `measurement-values-map` key 4 is a raw-value entry, which must have type `bstr`.
```

draft-ietf-rats-corim.md
```diff
@@ -2089,112 +2063,139 @@ Logically, new ECT entries are appended to the existing ACS.
...
 
 Profile writers SHOULD use CBOR tags for widely applicable comparison methods to ease Verifier implementation compliance across profiles.
 
 The following subsections define the comparison algorithms for the `measurement-values-map` keys defined by this specification.
 
 ##### Comparison for svn entries
 
-The value stored under `measurement-values-map` key 1 is an SVN, which must
-have type UINT.
+The value stored under `measurement-values-map` key 1 is an SVN, which must have type UINT.
```
deeglaze, line 2133:
```suggestion
The value stored under `measurement-values-map` key 1 is an SVN, which must have type `uint`.
```

draft-ietf-rats-corim.md
```diff
@@ -2089,112 +2063,139 @@ Logically, new ECT entries are appended to the existing ACS.
...
 
-If the Reference value for `measurement-values-map` key 1 is an untagged UINT or
-a UINT tagged with #6.552 then an equality comparison is performed. If the value
-of the SVN in ACS is not equal to the value in the Reference
-Value then the Reference Value does not match.
+If the condition ECT value for `measurement-values-map` key 1 is an untagged UINT or a UINT tagged with #6.552 then an equality comparison is performed.
```
deeglaze, line 2135:
```suggestion
If the condition ECT value for `measurement-values-map` key 1 is an untagged `uint` or a `uint` tagged with #6.552 then an equality comparison is performed.
```

draft-ietf-rats-corim.md
```diff
@@ -2089,112 +2063,139 @@ Logically, new ECT entries are appended to the existing ACS.
...
+The comparison MUST return true if the value of the SVN in the candidate entry is equal to the value in the condition ECT.
 
-If the Reference value for `measurement-values-map` key 1 is a UINT tagged with
-#6.553 then a minimum comparison is performed. If the value of the SVN in
-ACS less than the value in the Reference Value then the
-Reference Value does not match.
+If the condition ECT value for `measurement-values-map` key 1 is a UINT tagged with #6.553 then a minimum comparison is performed.
```
deeglaze, line 2138:
```suggestion
If the condition ECT value for `measurement-values-map` key 1 is a `uint` tagged with #6.553 then a minimum comparison is performed.
```

draft-ietf-rats-corim.md
```diff
@@ -2089,112 +2063,139 @@ Logically, new ECT entries are appended to the existing ACS.
...
 
 ##### Comparison for raw-value entries
 
-> [Andy] *I think this comparison method only works if the entry is at key 4 (because
-there needs to be a mask at key 5). Should we have a Reference Value of this
-which stores `[expect-raw-value raw-value-mask]` in an array?*
+The value stored under `measurement-values-map` key 4 is a raw-value entry, which must have type BSTR.
 
-[^issue] https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/71
+The value stored under the condition ECT `measurement-values-map` key 5 is a raw-value-mask entry, which must have type BSTR.
```
deeglaze, line 2164:
```suggestion
The value stored under the condition ECT `measurement-values-map` key 5 is a `raw-value-mask` entry, which must have type `bstr`.
```

draft-ietf-rats-corim.md
```diff
@@ -2089,112 +2063,139 @@ Logically, new ECT entries are appended to the existing ACS.
...
+
+If the condition ECT does not contain a raw-values-maek then the comparison MUST return true if the condition ECT raw-value is binary equal to the candidate entry raw-value.
```
deeglaze, line 2166:
```suggestion
If the condition ECT does not contain a `raw-values-mask` then the comparison MUST return true if the condition ECT `raw-value` is binary equal to the candidate entry `raw-value`.
```

deeglaze (changes requested):

draft-ietf-rats-corim.md
```diff
@@ -2089,112 +2063,139 @@ Logically, new ECT entries are appended to the existing ACS.
...
 ### Environment Comparison {#sec-compare-env}
 
+The verifier SHALL iterate over all entries in the ACS and SHALL add each entry where the condition ECT stateful enviromnment `environment-map` is a subset of the ACS entry `environment-map` to the candidate entries array.
+
+The condition ECT `environment-map` is a subset of an ACS entry `environment-map` if each field (for example `class`, `instance` etc.) which is present in the condition ECT `environment-map` is also present in the ACS entry, and the CBOR encoded field values in the stateful environment and ACS entry are binary identical.
+If a field is not present in the condition ECT `environment-map` then the presence of, and value of, the corresponding ACS entry field does not affect whether the `environment-map`s are subsets.
+
+Before performing the binary comparison, a Verifier SHOULD convert `environment-map` fields into a form which meets CBOR Core Deterministic Encoding Requirements {{-cbor}}.
+
+### Mkey comparison {#sec-compare-mkey}
```
thomas-fossati, line 2085:
```suggestion
### Element ID comparison {#sec-compare-mkey}
```

draft-ietf-rats-corim.md
```diff
@@ -2070,112 +2042,139 @@ Logically, new ECT entries are appended to the existing ACS.
...
-If all entries of the Reference Values array match a corresponding entry in the ACS array, then the `cryptokeys` Reference Value matches.
+If all entries of the condition ECTs array match a corresponding entry in the ACS array, then the `cryptokeys` condition ECT matches.
 Otherwise, `cryptokeys` does not match.
 
 ##### Comparison for Integrity Registers {#sec-cmp-integrity-registers}
 
-For each Integrity Register entry in the Reference Value, the Verifier will use the associated identifier (i.e., `integrity-register-id-type-choice`) to look up the matching Integrity Register entry in Evidence.
-If no entry is found, the Reference Value does not match.
+For each Integrity Register entry in the condition ECT, the Verifier will use the associated identifier (i.e., `integrity-register-id-type-choice`) to look up the matching Integrity Register entry in the candidate entry.
+If no entry is found, the condition MUST return false.
```
nedmsmith, line 2175:
```suggestion
If no entry is found, the condition ECT `integrity-registers` value does not match the candidate ACS ECT.
```

nedmsmith (commented):
See multiple comments

draft-ietf-rats-corim.md
andrew-draper: In the internal representation, each ECT has its own authority

draft-ietf-rats-corim.md
```diff
@@ -2070,112 +2042,139 @@ Logically, new ECT entries are appended to the existing ACS.
...
+If the `mkey` field is not present in neither the condition ECT nor the candidate entry then the `mkey` are equal and the candidate entry is not removed.
+If the `mkey` field is not present in only one side of the comparison then the `mkey` are not equal and the candiate entry is removed.
+
+### Authority comparison
+
+The verifier SHALL iterate over all entries in the candidate entries array and compare the condition ECT expected authority (`a`) value to the candidate entry authority value.
+
+If the condition ECT contains an `a` field then the Verifier SHALL remove all candidate entries whose `authorized-by` field does not contain one of the keys listed in the condition ECT `authorized-by` field (see {{sec-authorized-by}} for more details).
+
+When comparing two `$crypto-key-type-choice` fields for equality, the verifier SHALL treat then as equal if their CBOR encoding is binary equal.
```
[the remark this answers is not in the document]
andrew-draper: I will rewrite this to make it clear that it's a set intersection test

draft-ietf-rats-corim.md
```diff
@@ -2089,112 +2063,139 @@ Logically, new ECT entries are appended to the existing ACS.
...
+
+The verifier MAY treat two keys as equal if they have different formats but represent the same key.
```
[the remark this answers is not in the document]
andrew-draper: I added the "may" paragraph to make it clear what the intention of the comid author is. They are asking for authority from an entity, and that entity is currently represented by a key. In future we might develop more flexible ways to represent an identity, in which case I'd expect extension RFCs to describe the details of how they work.

draft-ietf-rats-corim.md
```diff
@@ -2089,112 +2063,139 @@ Logically, new ECT entries are appended to the existing ACS.
...
+For example, the verifier may contain code to compare a key fingerprint against the key which, when hashed, creates that fingerprint.
+
 ### Claims Comparison  {#sec-compare-claims}
 
-#### Comparison of measurement-values-map {#sec-match-one-codepoint}
+The verifier SHALL iterate over all entries in the candidate entries array and compare the condition ECT claims to the candidate entry authority claims.
```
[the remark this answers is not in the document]
andrew-draper: I will rewrite.

draft-ietf-rats-corim.md
```diff
@@ -2070,112 +2042,139 @@ Logically, new ECT entries are appended to the existing ACS.
...
```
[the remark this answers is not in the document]
andrew-draper: I think the current text is better, but I will remove candidate entry array

draft-ietf-rats-corim.md
```diff
@@ -2070,112 +2042,139 @@ Logically, new ECT entries are appended to the existing ACS.
...
+
+The Verifier SHALL iterate over the codepoints which are present in the condition ECT element's `measurement-values-map`.
```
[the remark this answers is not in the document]
andrew-draper: ok

draft-ietf-rats-corim.md
```diff
@@ -2089,112 +2063,139 @@ Logically, new ECT entries are appended to the existing ACS.
...
 
-[^issue] https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/203
+If any codepoint present in the condition ECT `measurement-values-map` does not match the same codepoint within the candidate entry `measurement-values-map` then verifier SHALL remove that candidate entry from the candidate entries array.
 
-If the `measurement-values-map` value has an associated CBOR tag, the comparison algorithm should comprehend the structure identified by the CBOR tag.
+#### Comparison of a single measurement-values-map codepoint{#sec-match-one-codepoint}
 
-If the Verifier does not recognize a CBOR tag value then the value MUST NOT match.
+If the codepoint key is zero or positive then the Verifier SHALL use a standard comparison algorithm, defined in this document, to compare the claims.
+The codepoint number and, if present, the tag value associated with the condition ECT codepoint SHALL be used to select the algorithm.
```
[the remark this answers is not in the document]
andrew-draper: This text exists because some standard codepoints use tags to select options

draft-ietf-rats-corim.md
andrew-draper: I have rewritten the text to try to make it clearer, hopefully without changing its meaning.

draft-ietf-rats-corim.md
```diff
@@ -2070,112 +2042,139 @@ Logically, new ECT entries are appended to the existing ACS.
...
 
-Note: CBOR tags are useful for discriminating values amongst alternates, but the comparison of tagged values is still determined by the codepoint they appear under. It is recommended but not required to compare values with a specific CBOR tag the same way across codepoints. For this reason, it is recommended to specify a default comparison algorithm with the CBOR tag's registration.
+If the codepoint key is negative then the Verifier MAY use a comparison algorithm defined by the appropriate CoRIM profile to compare the claims.
+The codepoint number, the profile associated with the condition ECT, and the tag value (if present) shall be used to select the algorithm.
```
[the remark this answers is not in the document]
andrew-draper: Text rewritten

draft-ietf-rats-corim.md
```diff
@@ -2070,133 +2042,195 @@ Logically, new ECT entries are appended to the existing ACS.
...
+
+If every entry in the condition ECT `authority` has a matching entry in the ACS entry `authority` field then the authorities match.
+The order of the fields in each `authority` field do not affect the result of the comparison.
+
+If any entry in the condition ECT `authority` does not have a matching entry in the ACS entry `authority` field then the authorities do not match.
+
+When comparing two `$crypto-key-type-choice` fields for equality, a Verifier SHALL treat them as equal if their deterministic CBOR encoding is binary equal.
+
+A Verifier MAY treat two keys as equal if they have different formats but represent the same key.
+For example, a Verifier may contain code to compare a key fingerprint against the key which, when hashed, creates that fingerprint.
```
nedmsmith, line 2092:
```suggestion

```
This kind of optional behavior could result in different verifiers arriving at a different result given the same inputs.
It is safer if the Verifier simply compared the "keys" (aka authority values) as opaque values rather than try to do format translation to a universal form.

draft-ietf-rats-corim.md
```diff
@@ -2070,133 +2042,195 @@ Logically, new ECT entries are appended to the existing ACS.
...
+If all entries of the condition ECTs array match a corresponding entry in the ACS array, then the `cryptokeys` condition ECT matches.
+Otherwise, `cryptokeys` does not match.
 
-### Conceptual Message Type Comparison  {#sec-compare-cm}
+##### Comparison for Integrity Registers {#sec-cmp-integrity-registers}
 
-The `cm` field comparison tests equality of one or more bits.
+For each Integrity Register entry in the condition ECT, the Verifier will use the associated identifier (i.e., `integrity-register-id-type-choice`) to look up the matching Integrity Register entry in the candidate entry.
+If no entry is found, the condition MUST return false.
+Instead, if an entry is found, the digest comparison proceeds as defined in {{sec-cmp-digests}} after equivalence has been found according to {{sec-comid-integrity-registers}}.
```
deeglaze, line 2225: Where is this equivalence defined? I don't see any "equivalence" wording in the referenced section. What does it mean for equivalence "to be found"?

I thought we'd describe the comparison the way we do with measurement-values-map, where if the key is in the condition, then the candidate must have a value for that key as well, and they compare thus.

draft-ietf-rats-corim.md
```diff
@@ -2070,133 +2042,195 @@ Logically, new ECT entries are appended to the existing ACS.
...
-The value stored under `measurement-values-map` key 2,
-or a value tagged with
-#6.TBD is a digest entry.
-It contains one or more digests, each measuring the
-same object. A Reference Value may contain multiple digests, each with a
-different algorithm acceptable to the Reference Value provider. If the
-digest in Evidence contains a single value with an algorithm and value
-matching one of the algorithms and values in the Reference Value then it
-matches.
+The value stored under `measurement-values-map` key 2 is a digest entry.
```
deeglaze, line 2173:
```suggestion
```
In the CDDL.

draft-ietf-rats-corim.md
```diff
@@ -2070,133 +2042,195 @@ Logically, new ECT entries are appended to the existing ACS.
...
+It contains one or more digests, each measuring the same object.
+When multiple digests are provided, each represents a different algorithm acceptable to the condition ECT author.
+
+In the simple case, a condition ECT digests entry containing one digest matches matches a candidate entry containing a single entry with the same algorithm and value.
+
+To prevent downgrade attacks, if there are multiple algorithms in common between the condition ECT and candidate entry then the digests calculated using all common algorithms must match.
```
deeglaze, line 2179:
```suggestion
To prevent downgrade attacks, if there are multiple algorithms in common between the condition ECT and candidate entry, then the bytes paired with common algorithms must be equal.
An `alg` matches another if they correspond to the same hash algorithm in the [IANA.named-information] registry.
A Verifier SHOULD fail the comparison if a condition `alg` does not have a corresponding entry in the IANA registry.
```

draft-ietf-rats-corim.md
```diff
@@ -2070,133 +2042,195 @@ Logically, new ECT entries are appended to the existing ACS.
...
+Otherwise, `cryptokeys` does not match.
 
-### Conceptual Message Type Comparison  {#sec-compare-cm}
+##### Comparison for Integrity Registers {#sec-cmp-integrity-registers}
 
-The `cm` field comparison tests equality of one or more bits.
+For each Integrity Register entry in the condition ECT, the Verifier will use the associated identifier (i.e., `integrity-register-id-type-choice`) to look up the matching Integrity Register entry in the candidate entry.
+If no entry is found, the condition MUST return false.
+Instead, if an entry is found, the digest comparison proceeds as defined in {{sec-cmp-digests}} after equivalence has been found according to {{sec-comid-integrity-registers}}.
+Note that it is not required for all the entries in the candidate entry to be used during matching: the condition ECT could consist of a subset of the device's register space. In TPM parlance, a TPM "quote" may report all PCRs in Evidence, while a condition ECT could describe a subset of PCRs.
```
deeglaze, line 2226:
```suggestion
```
I think this can be left unsaid, but I don't mind it.

deeglaze (changes requested):

draft-ietf-rats-corim.md
deeglaze: This decision could cause problems though. Since we’re requiring reference values to always include the expected root/intermediate/leaf expected signer from the attester (such as the sev-snp vcek, ask, or ark), then the reference value provider would want to avoid the bloat of copying the same full cert everywhere and just use a thumbprint. Is the expectation that the evidence translation will populate authorized-by with every format imaginable for describing the key in order to match? Or is the expectation for the profile to dictate a small number of supported formats that it will enumerate in the translation?

draft-ietf-rats-corim.md
```diff
@@ -2070,133 +2042,195 @@ Logically, new ECT entries are appended to the existing ACS.
...
+When comparing two `$crypto-key-type-choice` fields for equality, a Verifier SHALL treat them as equal if their deterministic CBOR encoding is binary equal.
+
+A Verifier MAY treat two keys as equal if they have different formats but represent the same key.
+For example, a Verifier may contain code to compare a key fingerprint against the key which, when hashed, creates that fingerprint.
+
+> I found this text later in the document. Is this what we want? Which ECTs are _first_ and _second_?
+>
+> The `a` field comparison tests for trust path termination.
+> If the authority of the first ECT is a trust anchor for the authority of the second ECT, the second ECT is valid.
+> If the authority values are identical, then the second ECT is valid.
```
deeglaze, line 2098: I've interpreted first as condition and second as candidate. I don't know where we're declaring trust anchors for evidence that is added to the ACS. Say an attestation key has multiple certificate paths. The verifier would know which are valid due to some configuration of trust anchor (I don't know where if anywhere we've specified what are the chosen trust anchors for a Verifier). For the purposes of explaining the abstract semantics in this document though, we can dictate that the whole trusted path is concretized in the ECT's authorities when added to the ACS as evidence. The concretization in this exposition can be abstract in a Verifier implementation.

If there is some connection to CoTS that I'm missing since I haven't been involved in that, I hope we can leave that abstraction to a mental exercise.

deeglaze (changes requested):

draft-ietf-rats-corim.md
```diff
@@ -2070,133 +2042,195 @@ Logically, new ECT entries are appended to the existing ACS.
...
+A Verifier SHALL compare the condition ECT expected `authority` value to the candidate entry `authority` value.
+
+If every entry in the condition ECT `authority` has a matching entry in the ACS entry `authority` field then the authorities match.
+The order of the fields in each `authority` field do not affect the result of the comparison.
+
+If any entry in the condition ECT `authority` does not have a matching entry in the ACS entry `authority` field then the authorities do not match.
+
+When comparing two `$crypto-key-type-choice` fields for equality, a Verifier SHALL treat them as equal if their deterministic CBOR encoding is binary equal.
+
+A Verifier MAY treat two keys as equal if they have different formats but represent the same key.
```
yogeshbdeshpande, line 2091: Remove the usage of MAY to remove line 2091 -2092. The Verifier should have the same format for comparision

draft-ietf-rats-corim.md
```diff
@@ -2070,133 +2042,195 @@ Logically, new ECT entries are appended to the existing ACS.
...
+For example, a Verifier may contain code to compare a key fingerprint against the key which, when hashed, creates that fingerprint.
+
+> I found this text later in the document. Is this what we want? Which ECTs are _first_ and _second_?
```
yogeshbdeshpande, line 2094: Remove the text on 2094 and retain text above!

draft-ietf-rats-corim.md
yogeshbdeshpande, line 2098:
```suggestion

```

yogeshbdeshpande (changes requested):
Please work on comments

yogeshbdeshpande:
Remove Raw Value comparison from the PR for a separate PR

draft-ietf-rats-corim.md
yogeshbdeshpande: @andrew-draper to correct digest comparison to match support both of text and integer for hash alg

yogeshbdeshpande (changes requested):
correct hash text and alg-id correction

draft-ietf-rats-corim.md
```diff
@@ -2070,133 +2042,195 @@ Logically, new ECT entries are appended to the existing ACS.
...
-### Name Space Comparison  {#sec-compare-ns}
+The value stored under `measurement-values-map` key 12 is an array of `$crypto-key-type-choice` entries. `$crypto-key-type-choice` entries are CBOR tagged values.
+The array contains one or more entries in sequence.
 
-The `ns` field comparison tests equality where the text values are identical.
+The CBOR tag of the first entry of the condition ECT `cryptokeys` array is compared with the CBOR tag of the first entry of the candidate entry `cryptokeys` value.
+If the CBOR tags match, then the bytes following the CBOR tag from the condition ECT entry are compared with the bytes following the CBOR tag from the candidate entry.
+If the byte strings match, and there is another array entry, then the next entry from the condition ECTs array is likewise compared with the next entry of the ACS array.
+If all entries of the condition ECTs array match a corresponding entry in the ACS array, then the `cryptokeys` condition ECT matches.
+Otherwise, `cryptokeys` does not match.
```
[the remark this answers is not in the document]
deeglaze: This is existing text, so I've moved this objection to a separate issue.

draft-ietf-rats-corim.md
```diff
@@ -2070,133 +2042,195 @@ Logically, new ECT entries are appended to the existing ACS.
...
+Before performing the binary comparison, a Verifier SHOULD convert both fields into a form which meets CBOR Core Deterministic Encoding Requirements {{-cbor}}.
+
+If any ACS entry `element-id` does not have a corresponding `element-id` in the ACS entry then the element map does not match.
+
+If any ACS entry has multiple corresponding `element-id`s then the element map does not match.
+
+Second, a Verifier SHALL compare the `element-claims` field within the condition ECT `element-list` and the corresponding field from the ACS entry (see {{sec-compare-mvm}}).
 
-### Environment Comparison {#sec-compare-env}
+### Measurement values map map Comparison {#sec-compare-mvm}
```
deeglaze, line 2135: map map
```suggestion
### Measurement values map Comparison {#sec-compare-mvm}
```

draft-ietf-rats-corim.md
```diff
@@ -2070,133 +2042,195 @@ Logically, new ECT entries are appended to the existing ACS.
...
 
 ##### Comparison for raw-value entries
 
-> [Andy] *I think this comparison method only works if the entry is at key 4 (because
-there needs to be a mask at key 5). Should we have a Reference Value of this
-which stores `[expect-raw-value raw-value-mask]` in an array?*
+The value stored under `measurement-values-map` key 4 is a raw-value entry, which must have type BSTR.
 
-[^issue] https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/71
+The value stored under the condition ECT `measurement-values-map` key 5 is a raw-value-mask entry, which must have type BSTR.
```
andrew-draper, line 2196:
```suggestion
[^issue] https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/71
```

draft-ietf-rats-corim.md
```diff
@@ -2070,133 +2042,195 @@ Logically, new ECT entries are appended to the existing ACS.
...
+A Verifier SHALL iterate over all ACS entries and SHALL attempte to match the condition ECT against each ACS entry using the algorithm described in {{sec-match-one-condition-ect}}.
+A Verifier SHALL create a "matched entries" set, and SHALL populate it with all ACS entries which matched the condition ECT.
+
+If the matched entries array is not empty, then the condition ECT matches the ACS.
+
+If the matched entries array is empty, then the condition ECT does not match the ACS.
+
+### Comparing a condition ECT against a single ACS entry {#sec-match-one-condition-ect}
+
+A Verifier SHALL perform all of the comparisons in sections {{sec-compare-environment}), {{sec-compare-authority}), {{sec-compare-cm}} and {{sec-compare-element-list}).
```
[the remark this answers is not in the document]
andrew-draper: I am not expecting evidence to always be marked with a profile, so doing a profile comparison first will sometimes fail.
If we have a profile comparison then should it be the other way round - a candidate ECT without a profile will match a condition ECT with a profile?

draft-ietf-rats-corim.md
```diff
@@ -2070,133 +2042,195 @@ Logically, new ECT entries are appended to the existing ACS.
...
+A Verifier SHALL compare each field which is present in the condition ECT `environment-map` against the corresponding field in the ACS entry `environment-map` using binary comparison.
+Before performing the binary comparison, a Verifier SHOULD convert both `environment-map` fields into a form which meets CBOR Core Deterministic Encoding Requirements {{-cbor}}.
+
+If all fields which are present in the condition ECT `environment-map` are present in the ACS entry and are binary identical, then the environments match.
+
+If any field which is present in the condition ECT `environment-map` is not present in the ACS entry then the environments do not match.
+
+If any field which is present in the condition ECT `environment-map` is not binary identical to the corresponding ACS entry field then the environments do not match.
+
+If a field is not present in the condition ECT `environment-map` then the presence of, and value of, the corresponding ACS entry field does not affect whether the environments match.
```
andrew-draper, line 2078:
```suggestion
If a field is not present in the condition ECT `environment-map` then the presence of, and value of, the corresponding ACS entry field SHALL NOT affect whether the environments match.
```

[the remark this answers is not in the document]
andrew-draper: Good suggestion, but I think the NOT in SHALL NOT is supposed to be in upper case.

draft-ietf-rats-corim.md
```diff
@@ -2070,133 +2042,195 @@ Logically, new ECT entries are appended to the existing ACS.
...
+
+### Authority comparison {#sec-compare-authority}
+
+A Verifier SHALL compare the condition ECT expected `authority` value to the candidate entry `authority` value.
```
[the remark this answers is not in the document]
andrew-draper: Does adding a hyphen here make the text clearer?

draft-ietf-rats-corim.md
andrew-draper: Removing using Ned's suggestion below.

draft-ietf-rats-corim.md
andrew-draper: I think we need a further change later to tidy up the processing of authority, I will remove this text now and we can add something like it later if that is the direction we want to follow.
I would like to see richer semantics available in the final draft.

draft-ietf-rats-corim.md
andrew-draper: Removed using suggested change below.

draft-ietf-rats-corim.md
andrew-draper: See comment about richer semantics. For now, I am removing it so we can make progress.

draft-ietf-rats-corim.md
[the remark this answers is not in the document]
deeglaze: I believe so. Also, it's an endash, and it's the grammatical thing to use to say that the expectation comes from the condition ECT (more than one word).

draft-ietf-rats-corim.md
```diff
@@ -2070,133 +2042,195 @@ Logically, new ECT entries are appended to the existing ACS.
...
+If any ACS entry has multiple corresponding `element-id`s then the element map does not match.
+
+Second, a Verifier SHALL compare the `element-claims` field within the condition ECT `element-list` and the corresponding field from the ACS entry (see {{sec-compare-mvm}}).
 
-### Environment Comparison {#sec-compare-env}
+### Measurement values map map Comparison {#sec-compare-mvm}
 
-### Claims Comparison  {#sec-compare-claims}
+A Verifier SHALL iterate over the codepoints which are present in the condition ECT element's `measurement-values-map`.
+Each of the codepoints present in the condition ECT is compared against the same codepoint in the candidate entry.
```
andrew-draper, line 2138:
```suggestion
Each of the codepoints present in the condition ECT is compared against the codepoint in the candidate entry with the same `element-id`.
```

draft-ietf-rats-corim.md
```diff
@@ -2070,133 +2042,195 @@ Logically, new ECT entries are appended to the existing ACS.
...
+### Element map comparison {#sec-compare-element-map}
+
+A Verifier shall compare each `element-map` within the condition ECT `element-list` against the ACS entry `element-list`.
+
+First, a Verifier SHALL locate the entries in the ACS entry `element-list` which have a matching `element-id` as the condition ECT `element-map`.
+Two `element-id` fields are the same if they are either both omitted, or both present with binary identical deterministic encodings.
+
+Before performing the binary comparison, a Verifier SHOULD convert both fields into a form which meets CBOR Core Deterministic Encoding Requirements {{-cbor}}.
+
+If any ACS entry `element-id` does not have a corresponding `element-id` in the ACS entry then the element map does not match.
```
andrew-draper, line 2129:
```suggestion
If any condition ECT `element-id` does not have a corresponding `element-id` in the ACS entry then the element map does not match.
```

draft-ietf-rats-corim.md
```diff
@@ -2070,133 +2042,195 @@ Logically, new ECT entries are appended to the existing ACS.
...
+
+If any ACS entry has multiple corresponding `element-id`s then the element map does not match.
```
andrew-draper, line 2131:
```suggestion
If any condition ECT entry `element-id` does not have a corresponding `element-id` in the ACS entry then the element map does not match.

If any condition ECT entry has multiple corresponding `element-id`s then the element map does not match.
```

draft-ietf-rats-corim.md
andrew-draper, line 2138:
```suggestion
Each of the codepoints present in the condition ECT `measurement-values-map` is compared against the same codepoint in the candidate entry `measurement-values-map`.
```

andrew-draper: I used slightly different text

draft-ietf-rats-corim.md
andrew-draper, line 2179:
```suggestion
To prevent downgrade attacks, if there are multiple algorithms in common between the condition ECT and candidate entry, then the bytes paired with common algorithms must be equal.
A Verifier SHALL treat two algorithm identifiers as equal if they have the same deterministic binary encoding.
A Verifier SHOULD treat two algorithm identifies as equal if one is a string and the other is an integer, and they correspond to the same hash algorithm in the [IANA.named-information] registry.
```

andrew-draper: I have been thinking about what we discussed in the call and I think there may be a security issue with it. As I remember, the decision in the call was that a verifier should treat the string and integer names for an algorithm as identical.
This leads to an issue if the registry is every extended, as some verifiers will not understand the new algorithms. If a condition ECT uses the numeric version and an ACS entry uses the string version then a downgrade attack is possible.
I am proposing text below which recommends that the integer version be used when it is available, and does not mandate comparing strings and integers.

draft-ietf-rats-corim.md
```diff
@@ -2070,133 +2042,195 @@ Logically, new ECT entries are appended to the existing ACS.
...
-between the Reference Value and the ACS entry then the
-Reference Value does not match.
+The comparison MUST return false if there are no common hash algorithms.
 
 ##### Comparison for raw-value entries
 
-> [Andy] *I think this comparison method only works if the entry is at key 4 (because
-there needs to be a mask at key 5). Should we have a Reference Value of this
-which stores `[expect-raw-value raw-value-mask]` in an array?*
+The value stored under `measurement-values-map` key 4 is a raw-value entry, which must have type BSTR.
```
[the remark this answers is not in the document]
andrew-draper: I have moved the raw-value changes to PR #294

andrew-draper, line 2194:
```suggestion
[^issue] https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/71
```

draft-ietf-rats-corim.md
andrew-draper: This is the same text as before the change.
Are we expecting to remove integrity registers? We discussed that in one call.
If we are then I think editorial improvements like the one above may be moot.

draft-ietf-rats-corim.md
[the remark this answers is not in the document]
andrew-draper: In this case, expected is a modifier of authority. If the "expected" there is causing confusion then I think it is probably best to just remove it.

draft-ietf-rats-corim.md
```diff
@@ -2070,82 +2042,148 @@ Logically, new ECT entries are appended to the existing ACS.
...
+A Verifier SHALL compare each condition ECT `measurement-values-map` value against the corresponding ACS entry value using the appropriate algorithm.
+
+Non-negative codepoints represent standard data representations.
+The comparison algorithms for these are defined in this document (in the sections below) or in other standards.
+For some non-negative codepoints their behavior is modified by the CBOR tag at the start of the condition ECT `measurement-values-map` value.
+
+Negative codepoints represent profile defined data representations.
+A Verifier SHALL use the codepoint number, the profile associated with the condition ECT, and the tag value (if present) to select the comparison algorithm.
+
+If a Verifier is unable to determine the comparison algorithm which applies to a codepoint then it SHALL behave as though the candidate entry does not match the condition ECT.
```
andrew-draper, line 2148: Is this the right thing to do if there is no defined comparison algorithm?
For most of the standard (non-negative) codepoints in the document today, a binary compare is what would be expected.
As that is a non-editorial change I propose making it in a different PR.

draft-ietf-rats-corim.md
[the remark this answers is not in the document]
deeglaze: Hmmmmmmmm that is very tricky indeed. That would mean that we have to be very careful that any CBOR that matches the base CDDL has a solid semantics that cannot be extended by a profile, such as with additional value encodings that allow for weaker pattern matching (e.g., adding any-of to standard codepoints if it weren't already there).

Say we have a condition ECT that has conditional endorsement series triple that on condition 3 includes nonstandard values, and condition 4 has standard values again and is the one we expect to match.
Do we permit the condition with a profile to be evaluated against the candidate without a profile until we get to the conditional series additions?

I understand -1 would never compare true to the nonnegative standard codepoints, but if there are extensible type choices in standard codepoints, we have to ensure that they are compared with encoding equality only to avoid overriding the base CoRIM semantics. `$version-scheme` for example permits any `int`, so how do we allow a profile to define a new one without messing with the base CoRIM semantics? I don't know. We haven't defined the semantics of version matching.

draft-ietf-rats-corim.md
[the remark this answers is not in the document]
deeglaze: Agreed. The condition's authority is already a condition you may be expecting.

draft-ietf-rats-corim.md
```diff
@@ -2070,82 +2042,147 @@ Logically, new ECT entries are appended to the existing ACS.
...
+
+If any field which is present in the condition ECT `environment-map` is not present in the ACS entry, then the environments do not match.
+
+If any field which is present in the condition ECT `environment-map` is not binary identical to the corresponding ACS entry field, then the environments do not match.
+
+If a field is not present in the condition ECT `environment-map` then the presence of, and value of, the corresponding ACS entry field SHALL NOT affect whether the environments match.
+
+### Authority comparison {#sec-compare-authority}
+
+A Verifier SHALL compare the condition ECT expected `authority` value to the candidate entry `authority` value.
```
deeglaze, line 2082:
```suggestion
A Verifier SHALL compare the condition ECT's `authority` value to the candidate entry's `authority` value.
```

draft-ietf-rats-corim.md
deeglaze: A profile may define only codepoints -1 to -15. If there's an update and -16 creeps in without the profile also getting a version update, do we want to say that additions are non-breaking, or that all changes require version updates and therefore new profile ids? I'd like that to be up to the profile designer, honestly, which would be within its rights to say that all code points below -15 or between -15 and -1000 are compared with binary equality to allow for opaque expansion but still allowing for a minor version advancement to add non-opaque comparisons to new code points.

I'd say no match. Keep this text.

draft-ietf-rats-corim.md
```diff
@@ -2070,82 +2042,147 @@ Logically, new ECT entries are appended to the existing ACS.
...
 
-If the CBOR encoding of the `digests` entry in the Reference Value or the
-ACS value with the same key is incorrect (for example if fields
-are missing or the wrong type) then the Reference Value does not match.
+The comparison MUST return false if the CBOR encoding of the `digests` entry in the condition ECT or the ACS value with the same key is incorrect (for example if fields are missing or the wrong type).
 
-The Verifier MUST iterate over the Reference Value `digests` array, locating
-hash algorithm identifiers that are present in the Reference Value and
-in the ACS entry.
+The comparison MUST return false if the condition ECT digests entry does not contain any digests.
```
deeglaze, line 2178:
```suggestion
The comparison MUST return false if the condition ECT digests entry does not contain any digests with an algorithm in common with the candidate entry.
```
The `digests-type` is already non-empty, so I assume you mean that there has to be at least 1 matching algorithm to compare digests of.

deeglaze: You say this below, so given that non-emptiness is required, I'd say delete the line.

draft-ietf-rats-corim.md
```diff
@@ -2157,46 +2194,26 @@ which stores `[expect-raw-value raw-value-mask]` in an array?*
...
-
-### Conceptual Message Type Comparison  {#sec-compare-cm}
-
-The `cm` field comparison tests equality of one or more bits.
+Note that it is not required for all the entries in the candidate entry to be used during matching: the condition ECT could consist of a subset of the device's register space. In TPM parlance, a TPM "quote" may report all PCRs in Evidence, while a condition ECT could describe a subset of PCRs.
 
 ### Profile-directed Comparison {#sec-compare-profile}
 
-A profile may specify handling for new CBOR tagged Reference Values.
+A profile may specify handling for new CBOR tagged condition ECTs.
```
deeglaze, line 2213:
```suggestion
A profile MUST specify comparison methods for its additions to `$`-prefixed CoRIM CDDL codepoints when this specification does not prescribe binary comparison.
```
Tagging is not required for extensions.

draft-ietf-rats-corim.md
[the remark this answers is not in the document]
nedmsmith: My expectation is if a profile defines an alternative matching algorithm from one defined in the "base" corim spec, the profile algorithm takes precedence.  A simple example might be svn comparisons (not min-svn) where current algorithm is equivalence, a RV profile might apply range semantics such that an actual svn=2 might be accepted by an algorithm that said (svn >= 1). Nevertheless, the ACS still contains svn=2 and any condition endorsement statement that compares for equality still uses the base spec's equality check algorithm. It suggests a constraint on profiles that the syntax for describing alternative matching should have a well-defined format. For example, the Intel profile defines a CBOR tag that prefaces the matching condition statement. Hence, the Verifier is not undecided about which algorithm to apply even if processing order varies, so long as the profile is parsed before processing begins.

draft-ietf-rats-corim.md
```diff
@@ -2070,82 +2042,147 @@ Logically, new ECT entries are appended to the existing ACS.
...
+If every entry in the condition ECT `authority` has a matching entry in the ACS entry `authority` field, then the authorities match.
+The order of the fields in each `authority` field do not affect the result of the comparison.
+
+If any entry in the condition ECT `authority` does not have a matching entry in the ACS entry `authority` field then the authorities do not match.
+
+When comparing two `$crypto-key-type-choice` fields for equality, a Verifier SHALL treat them as equal if their deterministic CBOR encoding is binary equal.
+
+
+
+### Conceptual Message Type Comparison  {#sec-compare-cm}
```
nedmsmith, line 2093: There isn't an external representation for cm-type in comid. Hence, there isn't a way to include it in a `condition` statement. The semantics for cm-type are intrinsic to processing semantics of CMs. For example, processing a RV message means the Evidence ECTs are the target ACS entries. For endorsements, Evidence and RV ECTs are the targets. Note: There is an unresolved issue as to whether Endorsement conditions can target other Endorsement ECTs.

Consequently, I don't believe this section is needed. The semantics of cm-type will be handled in the sections that describe processing phases.

draft-ietf-rats-corim.md
nedmsmith: I'm OK with the current text.

draft-ietf-rats-corim.md
```diff
@@ -2070,82 +2042,147 @@ Logically, new ECT entries are appended to the existing ACS.
...
+
+Negative codepoints represent profile defined data representations.
+A Verifier SHALL use the codepoint number, the profile associated with the condition ECT, and the tag value (if present) to select the comparison algorithm.
+
+If a Verifier is unable to determine the comparison algorithm which applies to a codepoint then it SHALL behave as though the candidate entry does not match the condition ECT.
 
 Profile writers SHOULD use CBOR tags for widely applicable comparison methods to ease Verifier implementation compliance across profiles.
 
-The following subsections define the comparison algorithms for the `measurement-values-map` keys defined by this specification.
+The following subsections define the comparison algorithms for the `measurement-values-map` codepoints defined by this specification.
```
nedmsmith, line 2152: We should use "key" or "code point" consistently. Line 2156 goes back to using "key".

draft-ietf-rats-corim.md
nedmsmith: The original text seems best IMO.

draft-ietf-rats-corim.md
```diff
@@ -2070,82 +2042,147 @@ Logically, new ECT entries are appended to the existing ACS.
...
+The comparison MUST return false if either digests entry contains multiple values for the same hash algorithm.
 
-If a hash algorithm identifier is present in both the Reference Value and
-the ACS, but the value of the hash is not binary identical
-between the Reference Value and the ACS entry then the
-Reference Value does not match.
+The Verifier MUST iterate over the condition ECT `digests` array, locating common hash algorithm identifiers (which are present in the condition ECT and in the candidate entry).
+If the value associated with any common hash algorithm identifier in the condition ECT differs from the value for the same algorithm identifier in the candidate entry then the comparison MUST return false.
+
+The comparison MUST return false if there are no common hash algorithms.
```
nedmsmith, line 2185:
```suggestion
The comparison MUST return false if there are no hash algorithms from the condition ECT in common with the hash algorithms from the candidate entry ECT.
```

nedmsmith (commented):
See suggested text

draft-ietf-rats-corim.md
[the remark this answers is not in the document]
deeglaze: svn-type-choice is not extensible, so your example doesn't track for me.

We shouldn't ever have certain condition tags in the ACS as actual state, such as `tagged-min-svn`.
Are we going to allow profiles to extend measurement-values-map values with their own pattern languages for more complex comparisons against the ACS? It becomes a kind of stored-procedure input at that point. It could be useful, but we can't permit for conditions to leak into actual state.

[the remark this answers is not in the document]
andrew-draper: I think the profile should take precedence if it exists, but that doesn't mean that we need to compare the profile between the condition ECT and the ACS entry.

Can we say something like this:
```suggestion
If the condition ECT contains a profile then the documentation for that profile describes the matching algorithm.
Profile authors SHOULD make their matching algorithms similar to the algorithm described in this document where possible.

A Verifier SHALL perform all of the comparisons in sections {{sec-compare-environment}), {{sec-compare-authority}), {{sec-compare-cm}} and {{sec-compare-element-list}).
```

This means that the profile can add new matching algorithms, like the ANY_OF<> algorithm proposed in another PR.

draft-ietf-rats-corim.md
andrew-draper: Ok, I will remove it.

draft-ietf-rats-corim.md
andrew-draper: There are conditions below which check for algorithms in common. I would prefer to keep the original text.

draft-ietf-rats-corim.md
andrew-draper: We used algorithm earlier, I will check in text using that work.

```suggestion
A profile MUST specify comparison algorithms for its additions to `$`-prefixed CoRIM CDDL codepoints when this specification does not prescribe binary comparison.
```

draft-ietf-rats-corim.md
```diff
@@ -2070,82 +2042,147 @@ Logically, new ECT entries are appended to the existing ACS.
...
 Profile writers SHOULD use CBOR tags for widely applicable comparison methods to ease Verifier implementation compliance across profiles.
 
-The following subsections define the comparison algorithms for the `measurement-values-map` keys defined by this specification.
+The following subsections define the comparison algorithms for the `measurement-values-map` codepoints defined by this specification.
 
 ##### Comparison for svn entries
 
-The value stored under `measurement-values-map` key 1 is an SVN, which must
-have type UINT.
+The value stored under `measurement-values-map` key 1 is an SVN, which must have type `uint`.
```
andrew-draper, line 2156:
```suggestion
The value stored under `measurement-values-map` codepoint 1 is an SVN, which must have type `uint`.
```

draft-ietf-rats-corim.md
andrew-draper: ok

draft-ietf-rats-corim.md
andrew-draper, line 2156:
```suggestion
The value stored under `measurement-values-map` codepoint 1 is an SVN, which must have type `uint`.
```

draft-ietf-rats-corim.md
```diff
@@ -2070,82 +2042,147 @@ Logically, new ECT entries are appended to the existing ACS.
...
 
-If the Reference value for `measurement-values-map` key 1 is an untagged UINT or
-a UINT tagged with #6.552 then an equality comparison is performed. If the value
-of the SVN in ACS is not equal to the value in the Reference
-Value then the Reference Value does not match.
+If the condition ECT value for `measurement-values-map` key 1 is an untagged `uint` or a `uint` tagged with #6.552 then an equality comparison is performed.
```
andrew-draper, line 2158:
```suggestion
If the condition ECT value for `measurement-values-map` codepoint 1 is an untagged `uint` or a `uint` tagged with #6.552 then an equality comparison is performed.
```

draft-ietf-rats-corim.md
```diff
@@ -2070,82 +2042,147 @@ Logically, new ECT entries are appended to the existing ACS.
...
+The comparison MUST return true if the value of the SVN in the candidate entry is equal to the value in the condition ECT.
 
-If the Reference value for `measurement-values-map` key 1 is a UINT tagged with
-#6.553 then a minimum comparison is performed. If the value of the SVN in
-ACS less than the value in the Reference Value then the
-Reference Value does not match.
+If the condition ECT value for `measurement-values-map` key 1 is a `uint` tagged with #6.553 then a minimum comparison is performed.
```
andrew-draper, line 2161:
```suggestion
If the condition ECT value for `measurement-values-map` codepoint 1 is a `uint` tagged with #6.553 then a minimum comparison is performed.
```

draft-ietf-rats-corim.md
```diff
@@ -2070,82 +2042,147 @@ Logically, new ECT entries are appended to the existing ACS.
...
-The value stored under `measurement-values-map` key 2,
-or a value tagged with
-#6.TBD is a digest entry.
-It contains one or more digests, each measuring the
-same object. A Reference Value may contain multiple digests, each with a
-different algorithm acceptable to the Reference Value provider. If the
-digest in Evidence contains a single value with an algorithm and value
-matching one of the algorithms and values in the Reference Value then it
-matches.
+It contains one or more digests, each measuring the same object.
```
andrew-draper, line 2166:
```suggestion
A digests entry contains one or more digests, each measuring the same object.
```

draft-ietf-rats-corim.md
```diff
@@ -2070,82 +2042,139 @@ Logically, new ECT entries are appended to the existing ACS.
...
+
+If the matched entries array is not empty, then the condition ECT matches the ACS.
+
+If the matched entries array is empty, then the condition ECT does not match the ACS.
+
+### Comparing a condition ECT against a single ACS entry {#sec-match-one-condition-ect}
+
+If the condition ECT contains a profile then the documentation for that profile describes the matching algorithm.
+Profile authors SHOULD make their matching algorithms similar to the algorithm described in this document where appropriate.
+
```
andrew-draper, line 2062:
```suggestion

```

draft-ietf-rats-corim.md
```diff
@@ -2070,82 +2042,139 @@ Logically, new ECT entries are appended to the existing ACS.
...
+If the matched entries array is empty, then the condition ECT does not match the ACS.
+
+### Comparing a condition ECT against a single ACS entry {#sec-match-one-condition-ect}
+
+If the condition ECT contains a profile then the documentation for that profile describes the matching algorithm.
+Profile authors SHOULD make their matching algorithms similar to the algorithm described in this document where appropriate.
+	
+A Verifier SHALL perform all of the comparisons in sections {{sec-compare-environment}), {{sec-compare-authority}) and {{sec-compare-element-list}).
+
```
andrew-draper, line 2064:
```suggestion

```

draft-ietf-rats-corim.md
```diff
@@ -2070,82 +2042,137 @@ Logically, new ECT entries are appended to the existing ACS.
...
+If the matched entries array is not empty, then the condition ECT matches the ACS.
+
+If the matched entries array is empty, then the condition ECT does not match the ACS.
+
+### Comparing a condition ECT against a single ACS entry {#sec-match-one-condition-ect}
+
+If the condition ECT contains a profile then the documentation for that profile describes the matching algorithm.
+Profile authors SHOULD make their matching algorithms similar to the algorithm described in this document where appropriate.
+A Verifier SHALL perform all of the comparisons in sections {{sec-compare-environment}), {{sec-compare-authority}) and {{sec-compare-element-list}).
```
andrew-draper, line 2062:
```suggestion

A Verifier SHALL perform all of the comparisons in sections {{sec-compare-environment}}, {{sec-compare-authority}} and {{sec-compare-element-list}}.

```

draft-ietf-rats-corim.md
```diff
@@ -2070,82 +2042,139 @@ Logically, new ECT entries are appended to the existing ACS.
...
+A Verifier SHALL iterate over all ACS entries and SHALL attempt to match the condition ECT against each ACS entry using the algorithm described in {{sec-match-one-condition-ect}}.
+A Verifier SHALL create a "matched entries" set, and SHALL populate it with all ACS entries which matched the condition ECT.
+
+If the matched entries array is not empty, then the condition ECT matches the ACS.
+
+If the matched entries array is empty, then the condition ECT does not match the ACS.
+
+### Comparing a condition ECT against a single ACS entry {#sec-match-one-condition-ect}
+
+If the condition ECT contains a profile then the documentation for that profile describes the matching algorithm.
```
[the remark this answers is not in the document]
deeglaze: In the space of candidate and condition having a profile id, we've discussed

both without: use base semantics
condition with, candidate without: use the profile matching algorithm for any extensions, but I think we're missing wording in the spec that if there are any non-base CDDL bits of a CoRIM that it MUST have a profile-id to give them meaning. Is the "not fo

We haven't discussed
condition without, candidate with: I'm guessing we use the base semantics' matching algorithm applies and rejects to match any nonstandard values.
condition with, candidate with: This one really does require mentioning in the document. I think that we need to require equality or at least condition profile-specified matching condition to allow for subsumption judgments.

deeglaze (changes requested):

draft-ietf-rats-corim.md
andrew-draper, line 2060: If the condition ECT contains a profile and the profile defines an algorithm for a codepoint and `environment-map`then a Verifier MUST use the algorithm defined by the profile (or a standard algorithm if the profile defines that).
If the condition ECT contains a profile, but the profile does not define an algorithm for a particular codepoint and `environment-map` then the verifier MUST use the standard algorithm described in this document to compare the data at that codepoint.

andrew-draper, line 2060:
```suggestion
If the condition ECT contains a profile and the profile defines an algorithm for a codepoint and `environment-map`then a Verifier MUST use the algorithm defined by the profile (or a standard algorithm if the profile defines that).
If the condition ECT contains a profile, but the profile does not define an algorithm for a particular codepoint and `environment-map` then the verifier MUST use the standard algorithm described in this document to compare the data at that codepoint.
```

nedmsmith (approved):

draft-ietf-rats-corim.md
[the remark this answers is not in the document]
deeglaze: Moving this objection to [an issue](https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/295) to not block.

deeglaze (approved):

draft-ietf-rats-corim.md
```diff
@@ -1940,39 +1940,11 @@ If the ECTs match, the `ev` `addition` ECT is added to the ACS.
 
 For each Conditional Endorsement Series Triple the Verifier iterates over the `conditional-series-record`s within the triple, stopping if it finds a match.
 
-For each iteration, the Verifier creates a temporary `stateful-environment-record` by merging the `stateful-environment-record` in the triple with the `refv` field in the `conditional-series-record`. It compares this temporary record against the ACS (see {{sec-match-one-se}}).
+For each iteration, the Verifier creates a temporary `stateful-environment-record` by merging the `stateful-environment-record` in the triple with the `refv` field in the `conditional-series-record`. It compares this temporary record against the ACS (see {{sec-match-condition-ect}}).
```
yogeshbdeshpande, line 1943: We need to correct these lines once PR #276 is merged, but i am fine with the text so far!

draft-ietf-rats-corim.md
```diff
@@ -2070,82 +2042,139 @@ Logically, new ECT entries are appended to the existing ACS.
...
-## ECT Comparison {#sec-ect-compare}
+## Comparing a condition ECT against the ACS {#sec-match-condition-ect}
 
 [^issue] https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/71
 
-This specification defines the comparison algorithm for the codepoints described in sub-sections below.
-A CoRIM profile may define additional negative codepoints and their matching algorithms.
-Specifications that extend CoMID MUST also define comparison algorithms for their added codepoints.
-If a codepoint's comparison algorithm is not stated or does not default to the comparison algorithms described in this specification, then the Verifier MUST compare the binary equality of the CBOR encodings of the values.
+The algorithm used to compare a condition ECT against the ACS is stateless; it depends only on the condition ECT being compared and the contents of the ACS at the time of the comparison.
```
yogeshbdeshpande, line 2049: Please check if line 2047 above is still applicable now that u have worked on this PR

draft-ietf-rats-corim.md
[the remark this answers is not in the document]
yogeshbdeshpande: > 
@deeglaze I think, I agree with @andrew-draper statement. A profile can modify/constrain a matching logic of the base parameter as well, in addition to specifying algorithm for extensions!

draft-ietf-rats-corim.md
```diff
@@ -2070,82 +2042,139 @@ Logically, new ECT entries are appended to the existing ACS.
...
 
-If the Verifier does not recognize a CBOR tag value then the value MUST NOT match.
+A Verifier SHALL iterate over all the entries in the condition ECT `element-list` and compare each one against the corresponding entry in the ACS entry `element-list`.
 
-Note: CBOR tags are useful for discriminating values amongst alternates, but the comparison of tagged values is still determined by the codepoint they appear under. It is recommended but not required to compare values with a specific CBOR tag the same way across codepoints. For this reason, it is recommended to specify a default comparison algorithm with the CBOR tag's registration.
+If every entry in the condition ECT `element-list` has a matching entry in the ACS entry `element-list` field then the element lists match.
+
+The order of the fields in each `element-list` field do not affect the result of the comparison.
+
+If any entry in the condition ECT `element-list` does not have a matching entry in the ACS entry `element-list` field then the authorities do not match.
```
yogeshbdeshpande, line 2103:
```suggestion
If any entry in the condition ECT `element-list` does not have a matching entry in the ACS entry `element-list` field then the `element-list` do not match.
```

draft-ietf-rats-corim.md
```diff
@@ -2070,82 +2042,139 @@ Logically, new ECT entries are appended to the existing ACS.
...
+If any codepoint present in the condition ECT `measurement-values-map` does not have a corresponding codepoint within the candidate entry `measurement-values-map` then Verifier SHALL remove that candidate entry from the candidate entries array.
+
+If any codepoint present in the condition ECT `measurement-values-map` does not match the same codepoint within the candidate entry `measurement-values-map` then Verifier SHALL remove that candidate entry from the candidate entries array.
+
+#### Comparison of a single measurement-values-map codepoint {#sec-match-one-codepoint}
+
+A Verifier SHALL compare each condition ECT `measurement-values-map` value against the corresponding ACS entry value using the appropriate algorithm.
+
+Non-negative codepoints represent standard data representations.
+The comparison algorithms for these are defined in this document (in the sections below) or in other standards.
```
yogeshbdeshpande, line 2134: Line 2134, we should clarify what other standards mean here ..? We can say, other standards as pointed in each individual sections

draft-ietf-rats-corim.md
```diff
@@ -2070,82 +2042,139 @@ Logically, new ECT entries are appended to the existing ACS.
...
 
-The value stored under `measurement-values-map` key 1 is an SVN, which must
-have type UINT.
+The value stored under `measurement-values-map` codepoint 1 is an SVN, which must have type `uint`.
 
-If the Reference value for `measurement-values-map` key 1 is an untagged UINT or
-a UINT tagged with #6.552 then an equality comparison is performed. If the value
-of the SVN in ACS is not equal to the value in the Reference
-Value then the Reference Value does not match.
+If the condition ECT value for `measurement-values-map` codepoint 1 is an untagged `uint` or a `uint` tagged with #6.552 then an equality comparison is performed.
```
yogeshbdeshpande, line 2150:
```suggestion
If the ECT value for `measurement-values-map` codepoint 1 is an untagged `uint` or a `uint` tagged with #6.552 then an equality comparison is performed.
```

yogeshbdeshpande: I believe this section of comparison of SVN Entries is a generic section, isn't it? So there may be condition ECT or normal ECT without a condition, we should be making it generic?

draft-ietf-rats-corim.md
```diff
@@ -2070,82 +2042,139 @@ Logically, new ECT entries are appended to the existing ACS.
...
+The comparison MUST return true if the value of the SVN in the candidate entry is equal to the value in the condition ECT.
 
-If the Reference value for `measurement-values-map` key 1 is a UINT tagged with
-#6.553 then a minimum comparison is performed. If the value of the SVN in
-ACS less than the value in the Reference Value then the
-Reference Value does not match.
+If the condition ECT value for `measurement-values-map` codepoint 1 is a `uint` tagged with #6.553 then a minimum comparison is performed.
```
yogeshbdeshpande, line 2153: Please check same comment as line 2150 applies to 2153 ?

yogeshbdeshpande (approved):
In general LGTM, unblocking the Pull Request. I have few questions, which we can conclude shortly in the CoRIM Design meeting today

draft-ietf-rats-corim.md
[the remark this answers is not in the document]
deeglaze: Matching logic I can get behind provided that we better distinguish possible and actual state in the representation. The profile types section text " A profile MUST NOT alter the syntax or semantics of CoRIM types defined in this document." is what led to my edit.

draft-ietf-rats-corim.md
deeglaze: I don't think that the algorithm is necessarily stateless if a profile introduces unification variables to enforce equalities of measurements without knowing their concrete values.

yogeshbdeshpande, line 2049:
```suggestion

```

draft-ietf-rats-corim.md
```diff
@@ -2070,82 +2042,139 @@ Logically, new ECT entries are appended to the existing ACS.
...
+A Verifier SHALL create a "matched entries" set, and SHALL populate it with all ACS entries which matched the condition ECT.
+
+If the matched entries array is not empty, then the condition ECT matches the ACS.
+
+If the matched entries array is empty, then the condition ECT does not match the ACS.
+
+### Comparing a condition ECT against a single ACS entry {#sec-match-one-condition-ect}
+
+If the condition ECT contains a profile then the documentation for that profile describes the matching algorithm.
+Profile authors SHOULD make their matching algorithms similar to the algorithm described in this document where appropriate.
```
deeglaze, line 2061:
```suggestion
```
Not meaningful.

draft-ietf-rats-corim.md
deeglaze: Conditions match against candidates. You don't consider the tagged-min to be a valid actual value in the ACS. I agree with the original wording.

draft-ietf-rats-corim.md
```diff
@@ -2070,82 +2042,138 @@ Logically, new ECT entries are appended to the existing ACS.
...
-The value stored under `measurement-values-map` key 2,
-or a value tagged with
-#6.TBD is a digest entry.
-It contains one or more digests, each measuring the
-same object. A Reference Value may contain multiple digests, each with a
-different algorithm acceptable to the Reference Value provider. If the
-digest in Evidence contains a single value with an algorithm and value
-matching one of the algorithms and values in the Reference Value then it
-matches.
+A digests entry contains one or more digests, each measuring the same object.
```
deeglaze, line 2157:
```suggestion
A `digests` entry contains one or more digests, each measuring the same object.
```

draft-ietf-rats-corim.md
```diff
@@ -2070,82 +2042,138 @@ Logically, new ECT entries are appended to the existing ACS.
...
+If any codepoint present in the condition ECT `measurement-values-map` does not have a corresponding codepoint within the candidate entry `measurement-values-map` then Verifier SHALL remove that candidate entry from the candidate entries array.
+
+If any codepoint present in the condition ECT `measurement-values-map` does not match the same codepoint within the candidate entry `measurement-values-map` then Verifier SHALL remove that candidate entry from the candidate entries array.
+
+#### Comparison of a single measurement-values-map codepoint {#sec-match-one-codepoint}
+
+A Verifier SHALL compare each condition ECT `measurement-values-map` value against the corresponding ACS entry value using the appropriate algorithm.
+
+Non-negative codepoints represent standard data representations.
+The comparison algorithms for these are defined in this document (in the sections below) or in other standards.
```
yogeshbdeshpande, line 2133:
```suggestion
The comparison algorithms for these are defined in this document (in the sections below) or in other specification.
```

henkbirkholz (approved):

thomas-fossati (approved):
LGTM

PR #292, opened 2024-09-27
Account for "activating" tags from CoBOM in theory of operation.

deeglaze:
The CoBOM tag can help narrow a search space and make an attestation result self-explanatory, such as "unsupported CoBOM" or "missing tags x,y,z from CoBOM".
Without a CoBOM, we wanted to still ensure that the appraisal process was repeatable. Add a SHOULD for tracking the head state of the knowledge base of all input CoRIMs to an appraisal session.

### DIFF
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -1466,6 +1466,10 @@ CoRIM-based data structures define an external representation of Conceptual Mess
 Appraisal processing describes both mapping transformations and Verifier reconciliation ({{sec-verifier-rec}}).
 Non-CoRIM-based data structures require mapping transformation, but these are out of scope for this document.
 
+If the Verifier requires CoBOMs, then the only CoRIM inputs are those activated by the CoBOM.
+If the Verifier does not require CoBOMs, then the CoRIM inputs consist of the knowledge base of CoRIMs available to the Verifier at the beginning of appraisal.
+The knowledge base state SHOULD have an immutable label for attestation results to have auditable provenance.
+
 If a CoRIM profile is specified, there are a few well-defined points in the procedure where Verifier behaviour depends on the profile.
 The CoRIM profile MUST provide a description of the expected Verifier behavior for each of those well-defined points.
 
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -1466,7 +1466,7 @@ CoRIM-based data structures define an external representation of Conceptual Mess
 Appraisal processing describes both mapping transformations and Verifier reconciliation ({{sec-verifier-rec}}).
 Non-CoRIM-based data structures require mapping transformation, but these are out of scope for this document.
 
-If the Verifier requires CoBOMs, then the only CoRIM inputs are those activated by the CoBOM.
+If a Verifier requires CoBOMs, then the CoRIM inputs indicated explicitly by the CoBOM are the only appropriate and required CoRIMs.
 If the Verifier does not require CoBOMs, then the CoRIM inputs consist of the knowledge base of CoRIMs available to the Verifier at the beginning of appraisal.
 The knowledge base state SHOULD have an immutable label for attestation results to have auditable provenance.
 


deeglaze:
Given that an appraisal session is meant to consider the knowledge base at a single auditable point of reference, I'm not sure if we need to keep the "must be activated atomically" wording.

deeglaze:
I keep falling back on "knowledge base" but I'm sure there's some term already decided upon that I can't recall. The immutable label recommendation apart from nebulous "auditability" is to support the new SLSA build environment track, where they're working to include remote attestation to identify build environments  https://github.com/slsa-framework/slsa/pull/1115

draft-ietf-rats-corim.md
```diff
@@ -1466,6 +1466,10 @@ CoRIM-based data structures define an external representation of Conceptual Mess
 Appraisal processing describes both mapping transformations and Verifier reconciliation ({{sec-verifier-rec}}).
 Non-CoRIM-based data structures require mapping transformation, but these are out of scope for this document.
 
+If a Verifier requires CoBOMs, then the CoRIM inputs indicated explicitly by the CoBOM are the only appropriate and required CoRIMs.
```
yogeshbdeshpande, line 1469: Pending a new issue where it is stated that a Verifier requires a CoBOM that text needs to be revisited before taking this PR further

yogeshbdeshpande (commented):
Please make a change!

nedmsmith:
Note: The CoBOM section states: "Verifier policy determines whether CoBOMs are required".  The last time this was discussed, there wasn't clarity AFAIK regarding what happens if a RIM supplier asserts a CoBOM but the Verifier doesn't require a CoBOM. Are all Verifiers required to support CoBOMs even if they don't "require" CoBOMs?

In the case where a Verifier requires a CoBOM but a supplier doesn't supply a CoBOM. What is the expected Verifier behavior? Should a Verifier create a logical CoBOM or should the Verifier refuse to process the RIM? 

The CoBOM section also says "If a Verifier Owner has a policy that does not require CoBOM, tags within a CoRIM received by a Verifier are activated immediately and treated valid for appraisal." It isn't clear whether a policy to "not require" means that the Verifier SHALL NOT process CoBOMs or that a Verifier will still process CoBOMs if they are supplied.

draft-ietf-rats-corim.md
thomas-fossati: @yogeshbdeshpande I don't understand this comment.  Could you please clarify?

draft-ietf-rats-corim.md
```diff
@@ -1466,6 +1466,10 @@ CoRIM-based data structures define an external representation of Conceptual Mess
...
+If the Verifier does not require CoBOMs, then the CoRIM inputs consist of the knowledge base of CoRIMs available to the Verifier at the beginning of appraisal.
+The knowledge base state SHOULD have an immutable label for attestation results to have auditable provenance.
```
thomas-fossati, line 1471: I like framing this in terms of "CoRIM knowledge base".

However, I'd scale this one up to also cover the "Verifier requires CoBOMs" case, e.g.:

> The CoRIMs available to the Verifier at the beginning of appraisal constitute the Verifier’s "knowledge base state."
> If a Verifier requires CoBOMs, its knowledge base state is made of the CoRIM inputs indicated explicitly by the CoBOMs.
> If the Verifier does not require CoBOMs, the knowledge base state consists of all the CoRIMs available to the Verifier.
> The knowledge base state SHOULD have an immutable label for attestation results to have auditable provenance.

draft-ietf-rats-corim.md
thomas-fossati: Thank you, but I still don't understand your point. L1469 takes no stance on who decides whether CoBOMs are used or not; it purely describes the mechanism. Who sets the boolean `use_cobom` on which the behaviour switches is out of scope for this paragraph.

draft-ietf-rats-corim.md
thomas-fossati, line 1471: I like this a lot. However, it's a requirement that affects the tracking done by the ACS to record the tag-id of the processed triples.

draft-ietf-rats-corim.md
yogeshbdeshpande: The confusion stems from the fact that:
`If a Verifier requires CoBOM`
this terminology is ambigous! as it dictates Verifier is the authority for deciding the need for CoBoM

draft-ietf-rats-corim.md
deeglaze: The entire attestation appraisal is frozen to the input state. The construction of the attestation context should be deterministic. The appraisal of the context should be deterministic. The result construction should be free to reference the label for the input state. I don't think we need to change ECT or ACS data types.

deeglaze:
I'll defer this to resolving "quantizing inputs"

PR #294, opened 2024-09-30
Move raw-value comparison algorithm to its own PR

andrew-draper:


draft-ietf-rats-corim.md
```diff
@@ -2149,11 +2149,21 @@ Reference Value does not match.
...
+
+If the lengths differ amongst the condition ECT `raw-values-mask`, condition ECT `raw-value` and candidate entry `raw-value`, then the comparison MUST return false.
+
+If all three lengths are the same, then a Verifier MUST iterate over the bits in the `raw-values-mask` which are 1, and compare the values for corresponding bit position in the two `raw-value` fields.
+
+If, for every bit position in `raw-values-mask` whose value is 1, the corresponding bits in both `raw-value`s are equal then the comparison MUST return true.
+
+If there is any bit position in `raw-values-mask` whose value is 1, but the two `raw-value`s differ at the same bit position, then the comparison MUST return false.
+
+Note that if a candidate entry contains a value for `raw-values-mask`, then this does not affect the result of the comparison.
```
deeglaze, line 2166:
```suggestion
If the condition ECT has a `raw-value-mask` and no `raw-value`, it is ignored.
If the condition ECT has both a `raw-value-mask`, `raw-value`, then the following conditions must be true for a match:

*  `raw-value` MUST be either a `bstr` or `tagged-bytes`. The compared value of `raw-value` is the `bstr` itself or the tagged `bstr` respectively.
*  the lengths of the condition `raw-value` `bstr`, condition `raw-value-mask`, and the candidate `raw-value` `bstr` are equal.
*  All bit positions in `raw-value-mask` whose value is 1 must imply that the bit values at the same position in both the condition `raw-value` `bstr` and the candidate `raw-value` `bstr` are equal.

Note that the candidate's `raw-value-mask` is ignored if it exists.
```
We can simplify if we want to treat `bstr` and `tagged-bytes` as different since they have different CBOR encodings, and since `bstr` isn't in the base CDDL's `$raw-value-type-choice`, but I figured there would likely be confusion about tagged-bytes vs non-tagged.

The "mask implies raw-value is a bstr" condition is to restrict the behavior of the mask to when the raw-value is truly just bytes that can be masked. Otherwise the type choice's alternatives wouldn't have a clear meaning with the mask's presence.

The quoted text is in 5.1.4.1.4.6 seems to contradict my proposal of limiting the scope to only `bstr`:
> When a new raw value type is defined, the convention for applying the mask is also defined. Typically, a CoRIM profile is used to define new raw values and mask semantics.

I disagree with this. The semantics of a standard codepoint should not be open to interpretation. A profile that wants nonstandard semantics needs to use a nonstandard codepoint.

deeglaze (changes requested):

draft-ietf-rats-corim.md
```diff
@@ -2149,11 +2149,21 @@ Reference Value does not match.
 
 ##### Comparison for raw-value entries
 
-> [Andy] *I think this comparison method only works if the entry is at key 4 (because
-there needs to be a mask at key 5). Should we have a Reference Value of this
-which stores `[expect-raw-value raw-value-mask]` in an array?*
+The value stored under `measurement-values-map` key 4 is a `raw-value` entry, which must have type BSTR.
```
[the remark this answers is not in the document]
nedmsmith: There is asymmetry in the definition of raw-value and raw-value-mask in that the mask is always `bstr` while raw-value-type-choice is extensible. The implication seems to be that the CBOR tag tells the Verifier how to map the raw value into a `bstr` so that the mask can be applied consistently.

If there is a way to write the text so that always applies regardless of $raw-value-type-choice extensions, then that is preferable. 

Nevertheless, before the raw value can be applied to the mask it must undergo a conversion to `bstr`. 

I think I would prefer a definition of raw-value that puts both the value and the mask behind the same CBOR tag. But what is there is OK for now IMHO.

draft-ietf-rats-corim.md
nedmsmith: > `raw-value` MUST be either a `bstr` or `tagged-bytes`.

The CDDL says the only `$raw-value-type-choice` defined is `tagged-bytes`. `bstr` only isn't defined.

nedmsmith: >the lengths of the condition `raw-value` `bstr`, condition `raw-value-mask`, and the candidate `raw-value` `bstr` are equal.

Is the "candidate" the same as the `additions`? If so, it would be more intuitive to say addition or addition candidate.

nedmsmith: >> When a new raw value type is defined, the convention for applying the mask is also defined. Typically, a CoRIM profile is used to define new raw values and mask semantics.

>I disagree with this. The semantics of a standard codepoint should not be open to interpretation. A profile that wants nonstandard semantics needs to use a nonstandard codepoint.

The extension point in `raw-value` isn't "non-standard" in that the CBOR tagging means there is a specification that defines the semantics of the structure. Nevertheless, the asymmetry of the extensibility seems awkward. Maybe an improvement is a mask type that puts both the value and the mask behind the same CBOR tag? But the convention for extensibility is the CBOR tag definition specifies the convention for representing both the value and the mask as well as the convention for converting the value and mask to a bstr.

deeglaze: > > `raw-value` MUST be either a `bstr` or `tagged-bytes`.
> 
> The CDDL says the only `$raw-value-type-choice` defined is `tagged-bytes`. `bstr` only isn't defined.

Yes, I included `bstr` to allow for a profile to specify it as an extension and to avoid footguns if a CoRIM issuer forgets the 560 tag on something that is reasonably expected to always be bytes. There are enough instances of x509 extensions that don't add the ASN.1 bytes tag byte in front of a byte string despite not strictly following the x509 spec for extensions that the value also needs to be an ASN.1 object that I accounted for that here. Might be overly cautious.

> Is the "candidate" the same as the ACS entry? If so, it would be more intuitive to say ACS entry.

The "candidate" word is to reference the induction variable for the iteration over all possible ACS entries. It helps disambiguate which entry we're talking about.

draft-ietf-rats-corim.md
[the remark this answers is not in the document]
deeglaze: @andrew-draper how disruptive is it to drop raw-value-mask in favor of a tagged-masked-bytes choice in raw-value-type-choice?

```
563([data: bstr, mask: bstr])
```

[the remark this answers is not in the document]
nedmsmith: > tagged-masked-bytes = 563([data: bstr, mask: bstr])

From the perspective of moving the Intel profile to a standard representation. This approach would be easier as the Intel profile already defines something similar:
```
mask-eq = 1
mask-operator = mask-eq
mask-type = bstr
mask-expression = [ mask-operator, value: mask-type, mask: mask-type ]
tagged-mask-expression = #6.60010( mask-expression )
```
I could see the following being easier to migrate to than the current raw-value:
```
mask-expression = [ mask-operator, 563([ data: bstr, mask: bstr]) ]
tagged-mask-expression = #6.60010( mask-expression )
```
In terms of disruptiveness, we should deprecate before deleting. Maybe the corim spec could add a second masked value type while discouraging use of raw-value?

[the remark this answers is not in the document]
yogeshbdeshpande: @andrew-draper : Add more explanation to explain how to use the `raw-value-type`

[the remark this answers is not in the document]
andrew-draper: I think tagged-mask-expresssion is a slightly nicer way to do the comparison. Since the mask only appears in CoRIM files (not in evidence) I don't think this has any outside effects.
I agree with deprecating before deleting.

cddl/raw-value-compare.cddl
```diff
@@ -0,0 +1,4 @@
+raw-value-compare = [
```
andrew-draper, line 1: There should be a tag before the array. I will add this if the direction is agreed.

draft-ietf-rats-corim.md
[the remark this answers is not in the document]
andrew-draper: The most recent proposal uses a mask expression (named `raw-value-compare`) similar to what is proposed by Ned.

draft-ietf-rats-corim.md
andrew-draper: I think we need to include non-tagged bstr because some evidence might be in that format. The place where tagging is useful is in raw-values within stateful-environment or similar.

draft-ietf-rats-corim.md
```diff
@@ -869,14 +869,9 @@ The following describes each member of the `measurement-values-map`.
...
-  An optional `raw-value-mask` (index 5) indicates which bits in the
-  `raw-value` field are relevant for verification. A mask of all ones ("1")
-  means all bits in the `raw-value` field are relevant. Multiple values could
-  be combined to create a single `raw-value` attribute. The vendor determines
-  how to pack multiple values into a single `raw-value` structure. The same
-  packing format is used when collecting Evidence so that Reference Values and
-  collected values are bit-wise comparable. The vendor determines the encoding
-  of `raw-value` and the corresponding `raw-value-mask`.
+  The vendor determines the encoding of `raw-value`.
+  When used for comparison, a mask may be provided indicting which bits in the `raw-value` field must be compared.
```
deeglaze, line 873:
```suggestion
  When used for comparison, a mask may be provided indicating which bits in the `raw-value` field must be compared.
```

draft-ietf-rats-corim.md
```diff
@@ -995,19 +990,25 @@ the sensitive values in memory are encrypted.
 Raw value measurements are typically vendor defined values that are checked by Verifiers
 for consistency only, since the security relevance is opaque to Verifiers.
 
-There are two parts to a `raw-value-group`, a measurement and an optional mask.
+A raw value from Evidence, or an Endorsement, is a value of type `bytes` with an optional tag value.
```
deeglaze, line 993:
```suggestion
A `raw-value` from Evidence or an Endorsement is a value of type `bytes` with an optional tag value.
```

draft-ietf-rats-corim.md
```diff
@@ -995,19 +990,25 @@ the sensitive values in memory are encrypted.
...
+Additional raw value types can be defined, but must be a CBOR tagged `bstr`.
+This allows parsers to distinguish between the various semantics of raw values, and lets Verifiers compare raw values without understanding their contents.
+
+A raw value intended for comparison, for example in `stateful-environment-record` can include a mask value, which selects the bits to compare during appraisal.
 The mask is applied by the Verifier as part of appraisal.
 Only the raw value bits with corresponding TRUE mask bits are compared during appraisal.
 
-When a new raw value type is defined, the convention for applying the mask is also defined.
-Typically, a CoRIM profile is used to define new raw values and mask semantics.
+When a new raw value type is defined, a comparison algorithm may also be defined.
```
deeglaze, line 1003: Nooo. This should be a measurement value map extensionnn

draft-ietf-rats-corim.md
```diff
@@ -2448,8 +2449,27 @@ The comparison MUST return false if there are no hash algorithms from the condit
 
 ##### Comparison for raw-value entries
 
+A `raw-value` entry contains binary data whose meaning the Verifier is not required to understand.
```
deeglaze, line 2452: "Not required" is different than MUST treat opaquely, or only with mask comparison. We don't want this type to have different meanings in different verifiers.

draft-ietf-rats-corim.md
```diff
@@ -2448,8 +2449,27 @@ The comparison MUST return false if there are no hash algorithms from the condit
...
 
-[^issue] https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/71
+The value stored under `measurement-values-map` key 4 in an ACS entry must be a `raw-value` entry, which must have type `bytes`, with an optional tag.
+
+The value stored under the condition ECT `measurement-values-map` key 4 may additionally be a `raw-value-compare` entry, which specifies an expected value and an optional mask.
```
deeglaze, line 2456:
```suggestion
```suggestion
The value stored under `measurement-values-map` key 4 in an ACS entry MUST have type `bytes` with an optional tag, or type `raw-value-compare`.
```
```

File: draft-ietf-rats-corim.md [hunk 9]
```diff
@@ -2448,8 +2449,27 @@ The comparison MUST return false if there are no hash algorithms from the condit
...
+
+If the condition ECT `measurement-value-map` key 4 is of type `bytes` or `tagged-bytes`, then the Verifier treats it in the same way as a `raw-value-compare` with the `value` field holding the same contents and the `mask` field omitted.
+The tag value is not used in the standard comparison function defined in this document.
```
deeglaze, line 2459:
```suggestion
The tag value MUST NOT be used in the standard comparison function defined in this document.
```

draft-ietf-rats-corim.md
```diff
@@ -2448,8 +2449,27 @@ The comparison MUST return false if there are no hash algorithms from the condit
...
+
+The comparison MUST return false if the lengths of the candidate entry value and the condition ECT value are different.
+
+If a mask is provided, then the comparison MUST return false if the lengths of the condition ECT mask and value are different.
+
+If a mask is not provided, then the comparison MUST return true if the condition ECT value is binary equal to the candidate entry value.
```
deeglaze, line 2465:
```suggestion
If a mask is not provided, then the comparison MUST return true if and only if the condition ECT value is binary equal to the candidate entry value.
```

draft-ietf-rats-corim.md
```diff
@@ -2448,8 +2449,27 @@ The comparison MUST return false if there are no hash algorithms from the condit
...
+
+If a mask field is provided, then the comparison MUST use the mask to determine which bits to compare.
+If a bit in the mask is 0 then this indicates that the corresponding bit in the ACE entry value may have either value.
```
deeglaze, line 2468: What's an ACE entry?

draft-ietf-rats-corim.md
```diff
@@ -995,19 +990,25 @@ the sensitive values in memory are encrypted.
 Raw value measurements are typically vendor defined values that are checked by Verifiers
 for consistency only, since the security relevance is opaque to Verifiers.
 
-There are two parts to a `raw-value-group`, a measurement and an optional mask.
+A raw value from Evidence, or an Endorsement, is a value of type `bytes` with an optional tag value.
 The default raw value measurement is of type `tagged-bytes` ({{sec-common-tagged-bytes}}).
-Additional raw value types can be defined, but must be CBOR tagged so that parsers can distinguish
-between the various semantics of type values.
 
+Additional raw value types can be defined, but must be a CBOR tagged `bstr`.
```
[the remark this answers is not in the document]
andrew-draper: I'm trying to keep the spirit of what was written before. The intention is that tags other than 6.560 can be used, providing the tagged object is `bstr`. A Verifier following this spec looks at the `bstr` as opaque binary data. A relying party following an extension can use the tag to understand what is inside the `bstr`.

draft-ietf-rats-corim.md
andrew-draper: Trying to keep the spirit of what was there before. Happy to lose this sentence.

draft-ietf-rats-corim.md
andrew-draper, line 2452:
```suggestion
A `raw-value` entry contains binary data which a Verifier following this specification will treat as opaque.
```

draft-ietf-rats-corim.md
andrew-draper: I think it is worthwhile for the spec to call out the difference between codepoint 4 in Evidence/Endorsements and in reference values (eg in stateful-environment-map).

draft-ietf-rats-corim.md
andrew-draper, line 2459:
```suggestion
The standard comparison function defined in this document does not use the tag value.
```

draft-ietf-rats-corim.md
andrew-draper, line 2468:
```suggestion
If a bit in the mask is 0 then this indicates that the corresponding bit in the ACS entry value may have either value.
```

draft-ietf-rats-corim.md
andrew-draper: I changed the text to say that a Verifier following this spec must treat `raw-value` as opaque data.

cddl/raw-value-compare.cddl
deeglaze: Agreed.

andrew-draper: TODO: Rename this object to a name which is a noun rather than a verb

cddl/raw-value.cddl
```diff
@@ -1,3 +1,5 @@
+$raw-value-type-choice /= bytes
```
thomas-fossati, line 1: why is this needed? is `tagged-bytes` not sufficient?

andrew-draper, line 1:
```suggestion
$raw-value-type-choice /= bytes
```
Should we remove the untagged bytes?

deeglaze: I see no strong reason to potentially break spdm evidence.

nedmsmith: > I see no strong reason to potentially break spdm evidence.

I prefer to keep it. This saves a couple of bytes, and it so happens I've been in conversations where I'm being asked to shave a handful of bytes.

draft-ietf-rats-corim.md
```diff
@@ -2448,8 +2449,27 @@ The comparison MUST return false if there are no hash algorithms from the condit
...
+
+If a mask is provided, then the comparison MUST return false if the lengths of the condition ECT mask and value are different.
+
+If a mask is not provided, then the comparison MUST return true if and only if the condition ECT value is binary equal to the candidate entry value.
+
+If a mask field is provided, then the comparison MUST use the mask to determine which bits to compare.
+If a bit in the mask is 0 then this indicates that the corresponding bit in the ACS Entry value may have either value.
+If, for every bit position in the mask whose value is 1, the corresponding bits in both values are equal then the comparison MUST return true.
+
+For backwards compatibility, a Verifier SHOULD treat a mask provided in condition ECT `measurement-values-mask` key 5 in the same way as a mask provided in the `mask` entry of `raw-value-compare`.
```
nedmsmith, line 2471: There's the potential for the mask in 4.raw-value-compare (or better name) to be different from the mask in 5 of both are populated. The language should be clear that if 4.raw-value-compare is used then 5 MUST/SHOULD not be used.  Or if 5 is used then 4.raw-value-compare MUST/SHOULD not be used.

SHOULD makes it harder to get verifier logic right. I'm leaning toward MUST.

draft-ietf-rats-corim.md
```diff
@@ -2448,8 +2449,27 @@ The comparison MUST return false if there are no hash algorithms from the condit
...
+This backwards compatibility option will be removed before publication.
```
nedmsmith, line 2472:
```suggestion

```

A previous paragraph already said 5 is deprecated. Don't need to repeat it here (in possibly more ambiguous text).

nedmsmith (commented):
see multiple comments

cddl/raw-value.cddl
andrew-draper: Untagged bytes removed

draft-ietf-rats-corim.md
```diff
@@ -995,19 +990,25 @@ the sensitive values in memory are encrypted.
...
 The mask is applied by the Verifier as part of appraisal.
 Only the raw value bits with corresponding TRUE mask bits are compared during appraisal.
 
-When a new raw value type is defined, the convention for applying the mask is also defined.
-Typically, a CoRIM profile is used to define new raw values and mask semantics.
+When a new raw value type is defined, a comparison algorithm may also be defined.
+Typically, a CoRIM profile is used to define new raw values and comparison algorithms.
+
+For backwards compatibility, the syntax where the mask appears under codepoint 5 is still supported.
+This will be removed before publication.
```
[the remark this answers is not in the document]
andrew-draper: Changed to use Ned's proposed text.

draft-ietf-rats-corim.md
```diff
@@ -2448,8 +2449,27 @@ The comparison MUST return false if there are no hash algorithms from the condit
 
 ##### Comparison for raw-value entries
 
+A `raw-value` entry contains binary data which a Verifier following this specification will treat as opaque.
 
-[^issue] https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/71
+The value stored under `measurement-values-map` key 4 in an ACS entry must be a `raw-value` entry, which must have type `bytes`, with an optional tag.
```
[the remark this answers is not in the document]
andrew-draper: I have changed "key" in this section to "codepoint". Should there be a space?

My take on the use of reference is that if a reference value with mask matches an ACS entry, then the reference value provider is corroborating the ACS entry which exists.

draft-ietf-rats-corim.md
andrew-draper: Removed

draft-ietf-rats-corim.md
```diff
@@ -2448,8 +2449,27 @@ The comparison MUST return false if there are no hash algorithms from the condit
 
 ##### Comparison for raw-value entries
 
+A `raw-value` entry contains binary data which a Verifier following this specification will treat as opaque.
```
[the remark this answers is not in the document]
andrew-draper: I have removed the second part of the sentence.
Each subsection in this area has an introductory paragraph and I think it would be odd for raw-value not to have one.

draft-ietf-rats-corim.md
```diff
@@ -995,19 +990,25 @@ the sensitive values in memory are encrypted.
...
 for consistency only, since the security relevance is opaque to Verifiers.
 
-There are two parts to a `raw-value-group`, a measurement and an optional mask.
+A `raw-value` from Evidence or an Endorsement is a value of type `bytes` that is optionally tagged #6.560.
 The default raw value measurement is of type `tagged-bytes` ({{sec-common-tagged-bytes}}).
-Additional raw value types can be defined, but must be CBOR tagged so that parsers can distinguish
-between the various semantics of type values.
 
+Additional raw value types can be defined, but must be a CBOR tagged `bstr`.
+This allows parsers to distinguish between the various semantics of raw values, and lets Verifiers compare raw values without understanding their contents.
```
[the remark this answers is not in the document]
andrew-draper: Yes, this is ambiguous. Rewritten.

draft-ietf-rats-corim.md
```diff
@@ -995,19 +990,25 @@ the sensitive values in memory are encrypted.
...
+
+A raw value intended for comparison, for example in `stateful-environment-record` can include a mask value, which selects the bits to compare during appraisal.
```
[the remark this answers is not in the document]
andrew-draper: Example removed

draft-ietf-rats-corim.md
```diff
@@ -995,19 +990,22 @@ the sensitive values in memory are encrypted.
 Raw value measurements are typically vendor defined values that are checked by Verifiers
 for consistency only, since the security relevance is opaque to Verifiers.
 
-There are two parts to a `raw-value-group`, a measurement and an optional mask.
+A `raw-value` from Evidence or an Endorsement is a value of type `bytes` that is optionally tagged #6.560.
 The default raw value measurement is of type `tagged-bytes` ({{sec-common-tagged-bytes}}).
-Additional raw value types can be defined, but must be CBOR tagged so that parsers can distinguish
-between the various semantics of type values.
 
+Additional raw value types can be defined, but must be a CBOR tagged `bstr`.
```
andrew-draper, line 996:
```suggestion
Additional value types can be added to `$raw-value-type-choice`, but must be a CBOR tagged `bstr`.
```

draft-ietf-rats-corim.md
```diff
@@ -995,19 +990,22 @@ the sensitive values in memory are encrypted.
 Raw value measurements are typically vendor defined values that are checked by Verifiers
 for consistency only, since the security relevance is opaque to Verifiers.
 
-There are two parts to a `raw-value-group`, a measurement and an optional mask.
+A `raw-value` from Evidence or an Endorsement is a value of type `bytes` that is optionally tagged #6.560.
```
yogeshbdeshpande, line 993: please remove optionally here!

draft-ietf-rats-corim.md
```diff
@@ -2448,8 +2446,24 @@ The comparison MUST return false if there are no hash algorithms from the condit
...
 ##### Comparison for raw-value entries
 
+A `raw-value` entry contains binary data.
 
-[^issue] https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/71
+The value stored under `measurement-values-map` codepoint 4 in an ACS entry must be a `raw-value` entry, which must have type `bytes`, with an optional tag.
+
+The value stored under the condition ECT `measurement-values-map` codepoint 4 may additionally be a `masked-raw-value` entry, which specifies an expected value and a mask.
+
+If the condition ECT `measurement-value-map` codepoint 4 is of type `bytes` or `tagged-bytes`, and there is no value stored under codepoint 5, then the Verifier treats it in the same way as a `masked-raw-value` with the `value` field holding the same contents and a `mask` of the same length as the value with all bits set.
```
yogeshbdeshpande, line 2455:
```suggestion
If the condition ECT `measurement-value-map` codepoint 4 is of `tagged-bytes`, and there is no value stored under codepoint 5, then the Verifier treats it in the same way as a `masked-raw-value` with the `value` field holding the same contents and a `mask` of the same length as the value with all bits set.
```

draft-ietf-rats-corim.md
andrew-draper, line 993:
```suggestion
A `raw-value` from Evidence or an Endorsement is a tagged value of type `bytes`.
This specification defines tag #6.560.
```

deeglaze (approved):

nedmsmith (approved):

henkbirkholz (approved):

draft-ietf-rats-corim.md
```diff
@@ -2448,8 +2447,24 @@ The comparison MUST return false if there are no hash algorithms from the condit
...
 
+A `raw-value` entry contains binary data.
 
-[^issue] https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/71
+The value stored under `measurement-values-map` codepoint 4 in an ACS entry must be a `raw-value` entry, which must be tagged and have type `bytes`.
+
+The value stored under the condition ECT `measurement-values-map` codepoint 4 may additionally be a `masked-raw-value` entry, which specifies an expected value and a mask.
+
+If the condition ECT `measurement-value-map` codepoint 4 is of `tagged-bytes`, and there is no value stored under codepoint 5, then the Verifier treats it in the same way as a `masked-raw-value` with the `value` field holding the same contents and a `mask` of the same length as the value with all bits set.
+The standard comparison function defined in this document does not use the tag value.
```
thomas-fossati, line 2457: in CBOR, the tag value is the whole thing.  "number" and "content" are the parts that we want to refer to here.


```suggestion
The standard comparison function defined in this document only uses the tag content (i.e., the tag number is stripped off).
```

draft-ietf-rats-corim.md
```diff
@@ -2448,8 +2447,24 @@ The comparison MUST return false if there are no hash algorithms from the condit
...
+
+For backwards compatibility, if the condition ECT `measurement-value-map` codepoint 4 is of type `bytes` or `tagged-bytes`, and there is a mask stored under codepoint 5, then the Verifier treats it in the same way as a `masked-raw-value` with the `value` field holding the same contents and a `mask` holding the contents of codepoint 5.
```
thomas-fossati, line 2459: I think we agreed that simple `bytes` is not a valid type:

```suggestion
For backwards compatibility, if the condition ECT `measurement-value-map` codepoint 4 is of type `tagged-bytes`, and there is a mask stored under codepoint 5, then the Verifier treats it in the same way as a `masked-raw-value` with the `value` field holding the same contents and a `mask` holding the contents of codepoint 5.
```

draft-ietf-rats-corim.md
```diff
@@ -2448,8 +2447,24 @@ The comparison MUST return false if there are no hash algorithms from the condit
...
+
+The comparison MUST return false if the lengths of the candidate entry value and the condition ECT value are different.
+
+The comparison MUST return false if the lengths of the condition ECT mask and value are different.
```
thomas-fossati, line 2463: hmm, isn't this an inconsistency that should be flagged as an error instead?

draft-ietf-rats-corim.md
```diff
@@ -2597,7 +2612,8 @@ IANA is requested to allocate the following tags in the "CBOR Tags" registry {{!
 |     560 | `bytes`             | tagged-bytes, see {{sec-common-tagged-bytes}}                 | {{&SELF}} |
 |     561 | `digest`            | tagged-cert-path-thumbprint-type, see {{sec-crypto-keys}}     | {{&SELF}} |
 |     562 | `bytes`             | tagged-pkix-asn1der-cert-type, see {{sec-crypto-keys}}        | {{&SELF}} |
-| 563-599 | `any`               | Earmarked for CoRIM                                           | {{&SELF}} |
+|     563 | `masked-raw-value`  | masked-raw-value, see {{sec-comid-raw-value-types}}           | {{&SELF}} |
```
thomas-fossati, line 2615: If you change this table you also need to recompute the cbor-tags file.

thomas-fossati (commented):
LGTM, apart from a few small issues that should be easy to fix

draft-ietf-rats-corim.md
```diff
@@ -995,19 +990,23 @@ the sensitive values in memory are encrypted.
 Raw value measurements are typically vendor defined values that are checked by Verifiers
 for consistency only, since the security relevance is opaque to Verifiers.
 
-There are two parts to a `raw-value-group`, a measurement and an optional mask.
+A `raw-value` from Evidence or an Endorsement is a tagged value of type `bytes`.
```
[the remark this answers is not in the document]
yogeshbdeshpande: Agree with Thomas, we should remove Evidence from the equation, but just focus on Reference Values and Endorsed Values

draft-ietf-rats-corim.md
```diff
@@ -995,19 +990,23 @@ the sensitive values in memory are encrypted.
...
 The default raw value measurement is of type `tagged-bytes` ({{sec-common-tagged-bytes}}).
-Additional raw value types can be defined, but must be CBOR tagged so that parsers can distinguish
-between the various semantics of type values.
 
+Additional value types can be added to `$raw-value-type-choice`, these additional values must be CBOR tagged `bstr`s.
+Constraining all raw value types to be `bstr` lets Verifiers compare raw values without understanding their contents.
+
+A raw value intended for comparison can include a mask value, which selects the bits to compare during appraisal.
 The mask is applied by the Verifier as part of appraisal.
 Only the raw value bits with corresponding TRUE mask bits are compared during appraisal.
```
yogeshbdeshpande, line 1002:
```suggestion
A raw value intended for comparison can also include a mask value, using the type `masked-raw-value` which selects the bits to compare during appraisal.
The mask is applied by the Verifier as part of appraisal.
Only the raw value bits with corresponding mask bits set to TRUE are compared during appraisal.
```

draft-ietf-rats-corim.md
```diff
@@ -995,19 +990,23 @@ the sensitive values in memory are encrypted.
...
 
-When a new raw value type is defined, the convention for applying the mask is also defined.
-Typically, a CoRIM profile is used to define new raw values and mask semantics.
+The `raw-value-mask` in `measurement-values-map` is deprecated, but retained for backwards compatibility.
```
yogeshbdeshpande, line 1004:
```suggestion
The `raw-value-mask` (code point 5) in `measurement-values-map` is deprecated, but retained for backwards compatibility.
```

draft-ietf-rats-corim.md
```diff
@@ -2448,8 +2447,24 @@ The comparison MUST return false if there are no hash algorithms from the condit
 
 ##### Comparison for raw-value entries
 
+A `raw-value` entry contains binary data.
 
-[^issue] https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/71
+The value stored under `measurement-values-map` codepoint 4 in an ACS entry must be a `raw-value` entry, which must be tagged and have type `bytes`.
```
yogeshbdeshpande, line 2452: ACS been an array of ECT, should this also not have either `tagged-bytes` or `masked-raw-value` ?

yogeshbdeshpande (commented):
Left some comments, happy to discuss before the meeting, to make progress...

draft-ietf-rats-corim.md
andrew-draper: We decided not to include non-tagged bstr

draft-ietf-rats-corim.md
andrew-draper, line 993:
```suggestion
A `raw-value` measurement, or an Endorsement, is a tagged value of type `bytes`.
```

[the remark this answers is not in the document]
andrew-draper: I have committed a different change which talks about measurements rather than evidence

draft-ietf-rats-corim.md
```diff
@@ -995,19 +990,23 @@ the sensitive values in memory are encrypted.
...
 for consistency only, since the security relevance is opaque to Verifiers.
 
-There are two parts to a `raw-value-group`, a measurement and an optional mask.
+A `raw-value` from Evidence or an Endorsement is a tagged value of type `bytes`.
+This specification defines tag #6.560.
 The default raw value measurement is of type `tagged-bytes` ({{sec-common-tagged-bytes}}).
-Additional raw value types can be defined, but must be CBOR tagged so that parsers can distinguish
-between the various semantics of type values.
 
+Additional value types can be added to `$raw-value-type-choice`, these additional values must be CBOR tagged `bstr`s.
```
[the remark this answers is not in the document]
andrew-draper: Applied, thanks

draft-ietf-rats-corim.md
andrew-draper: Syntactically yes, but it doesn't make sense. Do we want a way for evidence or an endorsement to say "I know these bits but not those bits?"

draft-ietf-rats-corim.md
andrew-draper, line 2457:
```suggestion
The standard comparison function defined in this document removes the tag before performing the comparison.
```

draft-ietf-rats-corim.md
andrew-draper: I am not sure how to do this, I will move this comment to an issue.

draft-ietf-rats-corim.md
andrew-draper: I made a similar change with different text.

draft-ietf-rats-corim.md
andrew-draper: Done

draft-ietf-rats-corim.md
yogeshbdeshpande: Yes, please raise a new issue, so that we address it in next iteration!

yogeshbdeshpande (approved):
LGTM!

draft-ietf-rats-corim.md
```diff
@@ -869,14 +869,9 @@ The following describes each member of the `measurement-values-map`.
...
 * `raw-value` (index 4): Contains the actual (not hashed) value of the element.
-  An optional `raw-value-mask` (index 5) indicates which bits in the
-  `raw-value` field are relevant for verification. A mask of all ones ("1")
-  means all bits in the `raw-value` field are relevant. Multiple values could
-  be combined to create a single `raw-value` attribute. The vendor determines
-  how to pack multiple values into a single `raw-value` structure. The same
-  packing format is used when collecting Evidence so that Reference Values and
-  collected values are bit-wise comparable. The vendor determines the encoding
-  of `raw-value` and the corresponding `raw-value-mask`.
+  The vendor determines the encoding of `raw-value`.
```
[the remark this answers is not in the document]
deeglaze: I don’t think 1. is an acceptable design decision. It’s not raw if it’s interpreted.

draft-ietf-rats-corim.md
```diff
@@ -995,19 +990,23 @@ the sensitive values in memory are encrypted.
...
 for consistency only, since the security relevance is opaque to Verifiers.
 
-There are two parts to a `raw-value-group`, a measurement and an optional mask.
+A `raw-value` measurement, or an Endorsement, is a tagged value of type `bytes`.
+This specification defines tag #6.560.
 The default raw value measurement is of type `tagged-bytes` ({{sec-common-tagged-bytes}}).
-Additional raw value types can be defined, but must be CBOR tagged so that parsers can distinguish
-between the various semantics of type values.
 
+Additional value types can be added to `$raw-value-type-choice`, these additional values must be CBOR tagged `bstr`s.
```
deeglaze, line 997: When you say "can be added", do you mean that a base CoRIM is permitted to have any tagged byte string as registered in https://www.iana.org/assignments/cbor-tags/cbor-tags.xhtml ? That would be convenient for me (and Ned, I think). If you mean that any profile extensions to `$raw-value-type-choice` are restricted to specify only tagged byte strings, that's less appealing.

draft-ietf-rats-corim.md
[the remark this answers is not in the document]
nedmsmith: The Verifier in phase <= 4 treats all measurements as semantically opaque. Whether a bstr or some other measurement type (say ip-address) is "fully" opaque hasn't been defined AFAIK. But maybe it doesn't need to be as the Verifier ignores ip-address semantics anyway. The only difference is how a comparison operation is defined. 

A profile can describe semantics for phases above 4 and for attestation results too. A section on profiles can document this. But there's little value in sprinkling bits from that section in other parts of the CoRIM spec especially in sections pertaining to phases 1-4. There's nothing really special about raw-value measurement type other than its name doesn't have semantic bias.

nedmsmith:
I'm OK to merge this PR as is.

draft-ietf-rats-corim.md
```diff
@@ -995,19 +990,23 @@ the sensitive values in memory are encrypted.
...
+Constraining all raw value types to be `bstr` lets Verifiers compare raw values without understanding their contents.
+
+A raw value intended for comparison can include a mask value, which selects the bits to compare during appraisal.
 The mask is applied by the Verifier as part of appraisal.
 Only the raw value bits with corresponding TRUE mask bits are compared during appraisal.
 
-When a new raw value type is defined, the convention for applying the mask is also defined.
-Typically, a CoRIM profile is used to define new raw values and mask semantics.
+The `raw-value-mask` in `measurement-values-map` is deprecated, but retained for backwards compatibility.
```
steven-bellock, line 1004: What do "deprecated" and "backwards compatible" refer to here? Previous drafts of this document?

cddl/masked-raw-value.cddl
```diff
@@ -0,0 +1,4 @@
+masked-raw-value = #6.563([
```
thomas-fossati, line 1: This deviates from the typographical conventions.
```suggestion
tagged-masked-raw-value = #6.563([
```

draft-ietf-rats-corim.md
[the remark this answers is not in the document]
thomas-fossati: > A profile can describe semantics for phases above 4

That is true, but a profile can also describe Evidence transformations. In that context, a semantic mapping from an Evidence claim to a `raw-value` can occur, with the `raw-value` being assigned a certain "name" (i.e., "semantics provided by the profile").
In Dionna's model (i.e., "fully opaque semantics"), this doesn't happen because `mkey` is already the name.

draft-ietf-rats-corim.md
deeglaze: That's my understanding.

steven-bellock: Ok then presumably the code point, and its deprecation, will (not may) be removed before ratification. No point in publishing a first version with deprecated content.

yogeshbdeshpande:
@andrew-draper I have rebased the PR to the main and it compiles. Do you have any outstanding work left to this PR? if so please do it now! Otherwise, we are happy to squash and merge!

Between, @thomas-fossati I do not see your appoval, so I assume you will review it again?

draft-ietf-rats-corim.md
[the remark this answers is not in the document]
thomas-fossati: What do we do with this?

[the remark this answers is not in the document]
yogeshbdeshpande: I do not have a particular objection to line 872, as truly it is up to a Vendor to decide how it uses `raw-value`

I would say to resolve the deadlock, shall we say?
The vendor determines the encoding of `raw-value` (as per the profile).

As the bottom text clarifies already, Thomas:
`Constraining all raw value types to be bstr lets Verifiers compare raw values without understanding their contents.`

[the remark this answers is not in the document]
thomas-fossati: I have no objections to what is in the PR.
I only asked to highlight that there are two cases: fully opaque semantics and semantics driven by profile.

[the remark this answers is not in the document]
andrew-draper: Does semantics here refer to data format or to matching rules?
Considering matching rules, for all types, I think the profile author can define new matching rules in their profile.
They can also define new data types, but should try to avoid defining a data type on a codepoint which conflicts with an existing data type on the same codepoint.

draft-ietf-rats-corim.md
andrew-draper: It is describing common usage today. I would be ok with removing codepoint 5 before ratification.

draft-ietf-rats-corim.md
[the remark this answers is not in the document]
thomas-fossati: > Does semantics here refer to data format or to matching rules?

Primarily semantics.  But that could also have a fall out on matching rules.

draft-ietf-rats-corim.md
nedmsmith: We want to avoid conflicts with implementations that want to be compatible with both TCG and IETF. If IETF omits CP 5, then it should be marked as "reserved" so that a future spec doesn't squat on it. 

Alternatively, the IANA registry should include measurement-values-map registry where cp 5 can be marked as in use by the TCG spec.

draft-ietf-rats-corim.md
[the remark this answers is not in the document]
nedmsmith: > In Dionna's model (i.e., "fully opaque semantics"), this doesn't happen because `mkey` is already the name.

`mkey` could specify semantics. If there are no standard defined mkey values, then a profile might define them. Either way, it doesn't matter to the processing rules for matching because it only cares to test mkey for equality. As long as the equality test respects data type it doesn't matter (to the verifier) that mkey might have deeper semantics. A conditional endorsement might care about deeper semantics, but again, Verifier doesn't care.

PR #304, opened 2024-10-03
Add DER certificate tag type

deeglaze:
CBOR is binary, so we shouldn't have to bloat up to base64 PEM encodings.


draft-ietf-rats-corim.md
```diff
@@ -1023,6 +1024,9 @@ A cryptographic key can be one of the following formats:
   certificates MUST be concatenated in order so that each directly certifies
   the one preceding.
 
+* `tagged-pkix-asn1der-key-type`: ASN.1 DER encoded X.509 public key certificate.
```
thomas-fossati, line 1027: We want to say this DER-encoded cert is `bstr`-wrapped.

Also, it looks like `cddl/crypto-key-type-choice.cddl` needs to be informed about this new type.

deeglaze: Done.

nedmsmith (approved):
LGTM

thomas-fossati:
Thanks!

I forgot one more thing.  Each time we add a new tag request, we should also:
```sh
make -C cddl cbor-tags.txt
```
to update the CBOR tags inventory (`cddl/cbor-tags.txt`).

thomas-fossati (approved):
LGTM, thanks!

yogeshbdeshpande (approved):
LGTM!

yogeshbdeshpande:
Thanks for this change @deeglaze 

@henk, please review, we should make every effort to speed up reviews so that we do not only do that during the meeting.

PR #305, opened 2024-10-04
Move contributors to front matter

deeglaze:
There is an extraneous Contributors section that went unnoticed since there was also a typographical mistake that demoted the section heading to normal text.

### DIFF
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -54,6 +54,18 @@ contributor:
     email: cabo@tzi.org
     contribution: >
       Carsten Bormann contributed to the CDDL specifications and the IANA considerations.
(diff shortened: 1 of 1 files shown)


thomas-fossati (changes requested):
I disagree with this PR.  Given the substantial and continuous contributions to the spec, both Dionna and Andy should be listed as co-authors rather than contributors.

nedmsmith (approved):
LGTM

thomas-fossati (approved):
I am OK with contributors (for now).

I want to have the discussion with the RATS chairs in Dublin to allow >5 editors.

yogeshbdeshpande (approved):
Thanks Dionna for the tidy up!

PR #306, opened 2024-10-04
Allow implicit environment for simple conditions

deeglaze:
Suggested by Yogesh

Will need to be accounted for after PR#299.


deeglaze:
Meeting suggested this was too casey.

PR #307, opened 2024-10-04
Quantize inputs with a session initialization phase.

deeglaze:
Closes Issue#242

The text already refers to an appraisal session and an appraisal context, but they are not introduced as concepts the same way the ACS is. Add Context and Session as specific constructs that serve distinct purposes.

I think the "CoRIMs are collected" text is a bit too vague, so I put all input fetching in phase 0. Validation and transformation is in phase 1.

I found it weird that Evidence Collection was not part of input gathering, so I've moved it.

This change clarifies the meaning of "active" and "inactive", and what "requires CoBOMs" means. It's just whether all available tags are included in the search space or only referenced ones are. A CoBOM is not a bundle of tags the way a CoRIM is. A CoBOM must reference tags that are already in the input set. It just flips them to being usable in the appraisal procedure.

I did not provide CDDL for the session or context since I think that needs to be sequenced after PR#299.


draft-ietf-rats-corim.md
```diff
@@ -1733,19 +1741,35 @@ An ARS is a list of ECTs that describe ACS entries that are selected for use as
...
 ## Input Validation and Transformation (Phase 1) {#sec-phase1}
 
-During the initialization phase, the CoRIM Appraisal Context is loaded with various conceptual message inputs such as CoMID tags ({{sec-comid}}), CoSWID tags {{-coswid}}, CoBOM tags {{-cobom}}, and cryptographic validation key material (including raw public keys, root certificates, intermediate CA certificate chains), and Concise Trust Anchor Stores (CoTS) {{-ta-store}}.
-These objects will be utilized in the Evidence Appraisal phase that follows.
-The primary goal of this phase is to ensure that all necessary information is available for subsequent processing.
+In Phase 1 the Verifier constructs an Appraisal Context that will serve as the set of valid sources of information for the Appraisal Procedure.
+The primary goal of this phase is to ensure that all necessary information is valid and available for subsequent processing.
 
-After context initialization, additional inputs are held back until appraisal processing has completed.
+No inputs other than dynamically-fetched information about certificate revocation status (see [RFC6960]) may be collected in Phase 1.
```
thomas-fossati, line 1766:
```suggestion
No inputs other than dynamically-fetched information about certificate revocation status (see {{?RFC6960}}) may be collected in Phase 1.
```

draft-ietf-rats-corim.md
```diff
@@ -1756,43 +1780,38 @@ For example, if the Evidence format is known in advance, CoRIMs using a profile
...
+If CoBOMs are required by policy, the selected tags are considered *inactive*.
+An *inactive* tag MUST NOT be used during the Appraisal Procedure.
+If CoBOMs are not required by policy, the selected tags are considered *active* and thus MUST be used during the Appraisal Procedure.
+
 Later stages will further select the CoRIMs appropriate to the Evidence Appraisal stage.
 
 #### Tags Extraction and Validation
 
-The Verifier chooses tags from the selected CoRIMs - including CoMID, CoSWID, CoBOM, and CoTS.
+The Verifier chooses tags from the selected CoRIMs—including CoMID, CoSWID, CoBOM, and CoTS.
```
thomas-fossati, line 1792:
```suggestion
The Verifier chooses tags from the selected CoRIMs, including CoMID, CoSWID, CoBOM, and CoTS.
```

thomas-fossati (commented):
Nice.

draft-ietf-rats-corim.md
```diff
@@ -1733,19 +1741,35 @@ An ARS is a list of ECTs that describe ACS entries that are selected for use as
 {::include cddl/intrep-ars.cddl}
 ~~~
 
+## Session setup (Phase 0) {#sec-phase0}
+
+The exchange of a request for attestation appraisal for a response of Attestation Results corresponds to a single Attestation Session.
+
+During this setup phase, the Verifier populates its Appraisal Session with a consistent view of all its inputs to the Appraisal Procedure.
+Inputs are various conceptual messages collected from Reference Value Providers, Endorsers, Verifier Owners, and Attesters.
+Conceptual messages may include Attestation Evidence, CoMID tags {{sec-comid}}, CoSWID tags {{-coswid}}, CoBOM tags {{sec-cobom}}, and static cryptographic validation key material (including raw public keys, root certificates, intermediate CA certificate chains, and Concise Trust Anchor Stores (CoTS) {{-ta-store}}.
```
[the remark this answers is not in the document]
deeglaze: Yeah.. Ned's change to add parentheses didn't treat this one line consistently, and adding parens only for CoBOM seems weird, no?

draft-ietf-rats-corim.md
```diff
@@ -1733,19 +1741,35 @@ An ARS is a list of ECTs that describe ACS entries that are selected for use as
...
+It is left to Verifier Policy to determine if input sources must use supply chain transparency constructs (see {{-scitt-arch}}) to track input provenance.
```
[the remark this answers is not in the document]
deeglaze: Verifier Policy can be to follow regulations or not. I didn't want to add speculation about this spec going to regulators as a framework.

draft-ietf-rats-corim.md
```diff
@@ -1733,19 +1741,35 @@ An ARS is a list of ECTs that describe ACS entries that are selected for use as
...
+
+The time at which the Verifier evaluates certificate- and tag-validity is an input.
+Once the Appraisal Session inputs are collected, no more may be added to the Appraisal Session apart from one exception.
+Certificate revocation status results may be collected during Phase 1, since there is no collective simultaneity of all responses.
```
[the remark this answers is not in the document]
deeglaze: Sounds like you're saying I should allow for the wall clock input to also be collected multiple times during Phase 1? I'm just trying to temper all sources of external influence on the hermeticity of an appraisal.

draft-ietf-rats-corim.md
```diff
@@ -1733,19 +1741,35 @@ An ARS is a list of ECTs that describe ACS entries that are selected for use as
...
+
+It is left to Verifier Policy to determine if or how to log the inputs used for a given Appraisal Session for optional use in Attestation Results.
```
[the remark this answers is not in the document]
deeglaze: Do you want different text?

draft-ietf-rats-corim.md
deeglaze: Should I add RFC6960 to the normative frontmatter instead? I don't know this {{?ref}} syntax.

draft-ietf-rats-corim.md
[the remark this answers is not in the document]
thomas-fossati: > [...] adding parens only for CoBOM seems weird, no?

It makes sense to add parentheses because external refs are rendered differently from internal refs:

* `{{EXT}}` becomes `[EXT]`
* `{{INT}}` becomes `INT`

draft-ietf-rats-corim.md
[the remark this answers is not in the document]
thomas-fossati: Understood. The thing is 9334's definition of policy is:

> _"A set of rules that a Verifier uses to evaluate the validity of information about an Attester."_

which would make any regulatory aspects out of scope.

draft-ietf-rats-corim.md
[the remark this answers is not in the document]
deeglaze: Added to both cobom and comid references then.

draft-ietf-rats-corim.md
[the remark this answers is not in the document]
thomas-fossati: I am just pointing out that creating an exception just for revocation checks to the general rule that _"Phase 0 locks all inputs"_ is probably unnecessary.  If we go without, the loss of functionality doesn't seem too big, given:
a) Temporal distance between Phases 0 and 1 is likely to be (very) short, and
b) Some clock skew tolerance will be in place anyway, making the boundaries between 0 and 1 even fuzzier.

draft-ietf-rats-corim.md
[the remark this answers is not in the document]
thomas-fossati: Yes, just a small tweak to acknowledge external factors other than the Verifier Owner's will may be at play.

draft-ietf-rats-corim.md
thomas-fossati: `{{?ref}}` automatically creates an informational reference entry for `ref`.
`{{!ref}}` automatically creates a normative reference entry for `ref`.

It's handy for GitHub's "suggested changes" because you don't have to fiddle with the front matter when you are miles deep in the scroll.

thomas-fossati: > Should I add RFC6960 to the normative frontmatter instead? 

I would make OCSP informative.  Using CRL distribution points is also an option.

deeglaze: Done.

draft-ietf-rats-corim.md
```diff
@@ -1539,12 +1542,18 @@ Appraisal Claims Set (ACS):
...
+: A structure that contains all state needed for performing the Appraisal Procedure.
+
 Appraisal Policy:
 : A description of the conditions that, if met, allow acceptance of Claims. Typically, the entity asserting a Claim should have knowledge, expertise, or context that gives credibility to the assertion. Appraisal Policy resolves which entities are credible and under what conditions.  See also "Appraisal Policy for Evidence" in {{-rats-arch}}.
 
 Attestation Results Set (ARS):
 : A structure that holds results of Appraisal and ECTs that are to be conveyed to a Relying Party.
 
+Appraisal Session:
+: A structure that tracks all state that corresponds to a single request for attestation appraisal. This includes the Appraisal Claims Set and Appraisal Context.
```
nedmsmith, line 1555:
```suggestion
: A structure that tracks all state that corresponds to a single request for attestation appraisal.
This includes the Appraisal Claims Set and Appraisal Context.
```
We're breaking lines on sentences right?

draft-ietf-rats-corim.md
```diff
@@ -1733,19 +1743,35 @@ An ARS is a list of ECTs that describe ACS entries that are selected for use as
...
 ## Input Validation and Transformation (Phase 1) {#sec-phase1}
 
-During the initialization phase, the CoRIM Appraisal Context is loaded with various conceptual message inputs such as CoMID tags ({{sec-comid}}), CoSWID tags {{-coswid}}, CoBOM tags {{-cobom}}, and cryptographic validation key material (including raw public keys, root certificates, intermediate CA certificate chains), and Concise Trust Anchor Stores (CoTS) {{-ta-store}}.
-These objects will be utilized in the Evidence Appraisal phase that follows.
-The primary goal of this phase is to ensure that all necessary information is available for subsequent processing.
+In Phase 1 the Verifier constructs an Appraisal Context that will serve as the set of valid sources of information for the Appraisal Procedure.
+The primary goal of this phase is to ensure that all necessary information is valid and available for subsequent processing.
 
-After context initialization, additional inputs are held back until appraisal processing has completed.
+No inputs other than dynamically-fetched information about certificate revocation status (see {{RFC6960}}, and {{Section 4.2.1.13 of -pkix-cert}}) may be collected in Phase 1.
```
nedmsmith, line 1768: I thought the previous text was better / orthogonal to the suggested text.
```suggestion
After the Appraisal Context is initialized, appraisal processing is allowed to complete before new inputs are considered.
```
I thought the previous text was better / orthogonal to the suggested text.

draft-ietf-rats-corim.md
```diff
@@ -1756,43 +1782,38 @@ For example, if the Evidence format is known in advance, CoRIMs using a profile
...
-
-During the Evidence collection phase, the Verifier communicates with Attesters to gather Evidence.
-The first part of this phase does not require any cryptographic validation.
-This means that Verifiers can use untrusted code to discover Evidence sources.
-Attesters are Evidence sources.
-
-Verifiers may rely on conveyance protocol specific context to identify an Evidence source, which is the Evidence input oracle for appraisal.
-
-The collected Evidence is then transformed to an internal representation, making it suitable for appraisal processing.
+If the Verifier Policy requires CoBOMs, then after the Verifier has processed all CoBOMs it MUST discard any tags which have not been activated by a CoBOM.
```
nedmsmith, line 1811: Removal of sec-ev-coll goes too far. This section corresponds to the CoRIM Selection, only for Evidence. I suggest:
```
#### Evidence Selection

All available Evidence in the Appraisal Session's inputs are checked for validity.

Evidence that is not within it's validity period, or that cannot be associated with an authenticated and authorized source MUST be discarded.

Evidence that has been secured by a cryptographic mechanism, such as a signature, that fails validation MUST be discarded.

Evidence selection MUST yield at least one usable entry.

Selected Evidence is transformed into an internal representation (see {{sec-phase1-trans}}).
```
The new text doesn't seem to belong in this section.

draft-ietf-rats-corim.md
```diff
@@ -1756,43 +1782,38 @@ For example, if the Evidence format is known in advance, CoRIMs using a profile
...
 
 #### Cryptographic Validation of Evidence {#sec-crypto-validate-evidence}
 
+The collected Evidence is transformed to an internal representation, making it suitable for appraisal processing.
+
```
nedmsmith, line 1816:
```suggestion

```

nedmsmith (changes requested):

draft-ietf-rats-corim.md
deeglaze: The “allowed” worries me, since we don’t require monotonic matching conditions for profiles and mid-appraisal additions of tags can lead to inconsistent ACS results.

draft-ietf-rats-corim.md
```diff
@@ -87,10 +87,12 @@ normative:
   IANA.named-information: named-info
 
 informative:
+  RFC6960:
```
thomas-fossati, line 90:
```suggestion
  RFC6960: ocsp
```

draft-ietf-rats-corim.md
thomas-fossati, line 1768:
```suggestion
No inputs other than dynamically-fetched information about certificate revocation status (see {{-ocsp}}, and {{Section 4.2.1.13 of -pkix-cert}}) may be collected in Phase 1.
```

thomas-fossati: Party pooper mode on. (For different reasons) I don't like either 🤣

draft-ietf-rats-corim.md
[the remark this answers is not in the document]
deeglaze: Added.

draft-ietf-rats-corim.md
deeglaze: Added.

draft-ietf-rats-corim.md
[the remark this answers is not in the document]
deeglaze: Moved the lock to be back at the end of Appraisal Context construction then.

draft-ietf-rats-corim.md
deeglaze: I've added a hermeticity subsection to describe this without any need for operational suggestions like having inputs be "held back".

draft-ietf-rats-corim.md
```diff
@@ -1882,6 +1900,13 @@ The Evidence ECT fields are populated as described in {{sec-phase1-trans}} and {
 Evidence transformation algorithms may be well-known, may be defined by a CoRIM profile ({{sec-corim-profile-types}}), or may be supplied dynamically.
 The handling of dynamic Evidence transformation algorithms is out of scope for this document.
 
+### Appraisal hermeticity
+
+The Appraisal Context at the and of Phase 1 constitutes all inputs to the Appraisal Procedure.
+The same Appraisal Context processed at any time MUST produce the same Attestation Results.
+
+The reason to lock the inputs before Attestation Appraisal is to ensure consistent ACS construction when accounting for any profile-driven semantics.
```
deeglaze, line 1908: Please help wordsmith this. There are obviously different levels of hermeticity and auditability that Verifier policy will impose. I don't want to require absolute determinism since that imposes a requirement on policy evaluation.
I don't like "same" since we're not prescribing any Attestation Result construction or representation, so there is no strong notion of sameness.

If we go with 
```suggestion
The reason to lock the inputs before Attestation Appraisal is for all Appraisal Procedure dependencies to be accounted for before interpreting them.                                                                                            
Fully determined inputs SHOULD ensure consistent ACS construction when accounting for any profile-driven semantics.                                                                                                                             
```

draft-ietf-rats-corim.md
```diff
@@ -1464,6 +1463,10 @@ They are not required to use the same internal representation or evaluation orde
 
 The appraisal procedure is divided into several logical phases for clarity.
 
++ **Phase 0**: Session setup
+
+During Phase 0, the Verifier collects its set of inputs from external sources that will be used for the remainder of the Appraisal Procedure to ensure the Procedure is deterministic.
```
thomas-fossati, line 1468: The deterministic nature is ensured by the consistency of the appraisal algorithm.
```suggestion
During Phase 0, the Verifier collects its set of inputs from external sources that will be used for the remainder of the Appraisal Procedure.
```

draft-ietf-rats-corim.md
thomas-fossati: ah, got it.  I thought it was a typo; my bad.

draft-ietf-rats-corim.md
```diff
@@ -1756,40 +1777,45 @@ For example, if the Evidence format is known in advance, CoRIMs using a profile
 
 The selection process MUST yield at least one usable tag.
 
+Selected CoRIMs are transformed into an internal representation (see {{sec-phase1-trans}}).
+If CoBOMs are required by policy, the selected tags are considered *inactive*.
+An *inactive* tag MUST NOT be used during the Appraisal Procedure.
```
thomas-fossati, line 1782: I find this way of presenting activation/disactivation a bit puzzling.

I see the flow from undecided/unknown to either active or inactive.
It'd be probably easier to grok if we introduced an initial "unknown" state and describe the decision tree as:

```mermaid
flowchart TD
    A[unknown] --> C{use_cobom}
    C -->|true| D{is_listed}
    C -->|false| E[active]
    D -->|true| F[active]
    D -->|false| G[inactive]
```

draft-ietf-rats-corim.md
```diff
@@ -1756,40 +1777,45 @@ For example, if the Evidence format is known in advance, CoRIMs using a profile
...
 
 The Verifier MUST discard all tags which are not syntactically and semantically valid.
-In particular, any cross-referenced triples (e.g., CoMID-CoSWID linking triples) MUST be successfully resolved.
+In particular, any cross-referenced triples (e.g., CoMID-CoSWID linking triples and CoBOM-listed tags) MUST be successfully resolved.
 
-#### CoBOM Extraction
+#### Tag activation by CoBOM
 
-This section is not applicable if the Verifier appraisal policy does not require CoBOMs.
+This step can be skipped of all tags are implicitly active.
```
thomas-fossati, line 1796: Doesn't read right to this non-native speeker. Did you mean s/of/if/? Or something else?

Regardless, I prefer it stupid and explicit, as it was before :-)
Unless you see a case that would not be covered using the old phrasing.

draft-ietf-rats-corim.md
thomas-fossati: I agree that the old L1906 is not 100% correct: it should probably have said instead:

> The same Appraisal Context processed at any time MUST produce the same ACS.

I like the text you suggest.  The only thing is we should state more strongly that what we are trying to achieve is _deterministic ACS construction._

draft-ietf-rats-corim.md
deeglaze: The appraisal algorithm is deterministic for the base CDDL, but profiles can define their own comparison algorithms that can be complex and have computational timeouts that lead to nondeterministic results (for example). Considering the SLSA level 4 language about hermeticity and reproducibility, we should have wording about best-effort reproducibility https://slsa.dev/spec/v0.1/requirements

If we consider a Verifier to be a "builder" of attestation results, then we can apply the same requirements whole sale.

draft-ietf-rats-corim.md
deeglaze: Yes s/of/if/, typo.

draft-ietf-rats-corim.md
thomas-fossati: Hmm, even with profile-specific comparison algorithms (or other kinds of custom behaviour), _successful_ ACS computation should be deterministic.

If computational timeouts, or other conditions that make the ACS "incomplete", happen, they must be made fully visible to the downstream appraisal policy.

deeglaze: Suppose we have a profile that has unification variables that are allowed in conditions and in additions. The logic it employs for unification does not have principal types, so there are multiple incomparable correct solutions to a system of constraints. Which substitution it resolves on can be dependent on something nondeterministic like hash map iteration order. The resulting ACS is not consistently reproducible in this case, the way a compiler can choose which way it allocates registers or orders if/then/else basic blocks.

I'm not saying it's sensible, I'm saying it's not ruled out and it is very difficult to rule out. Thus a SHOULD be reproducible.

deeglaze: We've already said that certain decisions can be make by policy. Are those decisions made strictly after phase 4? If we can say phases 2, 3, and 4 must be deterministic without imposing policy burden by the way we define semantics, then I'd certainly like to make that declaration.

I don't think profiles would be introducing randomized behavior. Non-deterministic isn't necessarily random. You can have "the same" outcome by means of equivalence but not syntactic equality. When we're talking about ACS construction we're making representation choices that may be more conducive to equal outputs than something that uses, say, hash maps or algebraic constructions that represent "the same object" in multiple ways.

thomas-fossati: > We've already said that certain decisions can be make by policy. Are those decisions made strictly after phase 4?

yes

> If we can say phases 2, 3, and 4 must be deterministic without imposing policy burden by the way we define semantics, then I'd certainly like to make that declaration.

👍 

> I don't think profiles would be introducing randomized behavior. Non-deterministic isn't necessarily random. You can have "the same" outcome by means of equivalence but not syntactic equality. When we're talking about ACS construction we're making representation choices that may be more conducive to equal outputs than something that uses, say, hash maps or algebraic constructions that represent "the same object" in multiple ways.

Got it.  Then:

> _"profiles MUST NOT introduce non-deterministic behaviour."_

draft-ietf-rats-corim.md
nedmsmith: > Fully determined inputs SHOULD ensure consistent ACS construction when accounting for any profile-driven semantics.

```suggestion
Given the same Appraisal Context, different Verifier appraisals MUST produce deterministic results.

Verifier inputs are locked (see Step 1) before Attestation Appraisal to ensure ACS construction is deterministic.
```
Text above addresses the question of profile or policy inputs rendering non-deterministic ACS results.

draft-ietf-rats-corim.md
```diff
@@ -1733,19 +1743,30 @@ An ARS is a list of ECTs that describe ACS entries that are selected for use as
 {::include cddl/intrep-ars.cddl}
 ~~~
 
-## Input Validation and Transformation (Phase 1) {#sec-phase1}
+## Session setup (Phase 0) {#sec-phase0}
```
nedmsmith, line 1746: Why change from phase 1 to phase 0?

draft-ietf-rats-corim.md
```diff
@@ -1733,63 +1742,113 @@ An ARS is a list of ECTs that describe ACS entries that are selected for use as
...
 
-The collected Evidence is then transformed to an internal representation, making it suitable for appraisal processing.
+A tag's activity is determined by the flowchart in figure {#fig-tag-activity}
+
+flowchart TD
+    A[unknown] --> C{use_cobom}
+    C -->|true| D{is_listed}
+    C -->|false| E[active]
+    D -->|true| F[active]
+    D -->|false| G[inactive]
```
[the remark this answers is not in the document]
deeglaze: The referenced issue does not appear to have solved the problem. Do you know if this change would work?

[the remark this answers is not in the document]
thomas-fossati: I believe it wouldn't because of point 3 (i.e., `mmdc` not producing conformant SVG)

nedmsmith:
I tried to build on local environment after installing mermaid-cli but still get errors. 
Also, the branch seems to be significantly behind main - getting build warnings that have been fixed in main.

draft-ietf-rats-corim.md
nedmsmith: > A diagram like this would be helpful IMO.

But maybe the diagram should be backed out of this PR until the tooling issues can be addressed. It's difficult to suggest changes when it doesn't build.

draft-ietf-rats-corim.md
```diff
@@ -1733,63 +1741,111 @@ An ARS is a list of ECTs that describe ACS entries that are selected for use as
...
 
-The selection process MUST yield at least one usable tag.
+During this setup phase, the Verifier populates its Appraisal Session with a consistent view of all its inputs to the Appraisal Procedure.
+Inputs are various conceptual messages collected from Reference Value Providers, Endorsers, Verifier Owners, and Attesters.
+Conceptual messages may include Attestation Evidence, CoMID tags ({{sec-comid}}), CoSWID tags {{-coswid}}, CoBOM tags ({{sec-cobom}}), Policy, and cryptographic validation key material (including raw public keys, root certificates, intermediate CA certificate chains, certificate revocation data (see {{-ocsp}} or {{Section 4.2.1.13 of -pkix-cert}}), and Concise Trust Anchor Stores (CoTS) {{-ta-store}}.
+The clock time used for validity judgments and policy evaluation is an input.
 
-Later stages will further select the CoRIMs appropriate to the Evidence Appraisal stage.
+How the Verifier collects its inputs is out of scope of this document.
+There will be some amount of CoRIMs and standalone tags available as inputs that make an `asession`:
```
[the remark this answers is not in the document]
deeglaze: Okay, I'll remove the internal representation and leave the description to the prose of the Verifier Abstraction.

cddl/intrep-actx.cddl
```diff
@@ -0,0 +1,8 @@
+ir-appraisal-context = {
```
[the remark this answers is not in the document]
deeglaze: Tracked as Issue #319

draft-ietf-rats-corim.md
```diff
@@ -1756,40 +1777,45 @@ For example, if the Evidence format is known in advance, CoRIMs using a profile
...
-#### CoBOM Extraction
+#### Tag activation by CoBOM
 
-This section is not applicable if the Verifier appraisal policy does not require CoBOMs.
+This step can be skipped of all tags are implicitly active.
 
 CoBOMs which are not within their validity period are discarded.
-
-The Verifier processes all CoBOMs that are valid at the point in time of Evidence Appraisal and activates all tags referenced therein.
+The Verifier processes all CoBOMs that are valid at the point in time of Evidence Appraisal.
```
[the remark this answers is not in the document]
deeglaze: Removed.

draft-ietf-rats-corim.md
```diff
@@ -1733,63 +1741,111 @@ An ARS is a list of ECTs that describe ACS entries that are selected for use as
...
 
-During the initialization phase, the CoRIM Appraisal Context is loaded with various conceptual message inputs such as CoMID tags ({{sec-comid}}), CoSWID tags {{-coswid}}, CoBOM tags {{-cobom}}, and cryptographic validation key material (including raw public keys, root certificates, intermediate CA certificate chains), and Concise Trust Anchor Stores (CoTS) {{-ta-store}}.
-These objects will be utilized in the Evidence Appraisal phase that follows.
-The primary goal of this phase is to ensure that all necessary information is available for subsequent processing.
+An Appraisal Session includes Verifier-specific session state as well as a collection of inputs to process.
+Conceptual Messages are given explicit representation in the session.
 
-After context initialization, additional inputs are held back until appraisal processing has completed.
+~~~ cddl
+{::include cddl/intrep-asession.cddl}
```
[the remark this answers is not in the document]
deeglaze: Removed appraisal session representation.

draft-ietf-rats-corim.md
```diff
@@ -1539,12 +1540,19 @@ Appraisal Claims Set (ACS):
...
+: A structure that contains all state needed for performing the Appraisal Procedure.
+
 Appraisal Policy:
 : A description of the conditions that, if met, allow acceptance of Claims. Typically, the entity asserting a Claim should have knowledge, expertise, or context that gives credibility to the assertion. Appraisal Policy resolves which entities are credible and under what conditions.  See also "Appraisal Policy for Evidence" in {{-rats-arch}}.
 
 Attestation Results Set (ARS):
 : A structure that holds results of Appraisal and ECTs that are to be conveyed to a Relying Party.
 
+Appraisal Session:
+: A structure that tracks all state that corresponds to a single request for attestation appraisal.
```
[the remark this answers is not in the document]
deeglaze: Removed.

cddl/intrep-actx.cddl
```diff
@@ -0,0 +1,8 @@
...
+  ae : [ + ae ]
+  rv : [ rv ]
+  ev : [ ev ]
+  evs : [ evs ]
+  ACS: ACS
+  extra: any
```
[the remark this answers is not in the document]
deeglaze: Extra in actx is meant to capture implementation-specific details. I could leave it unsaid, but I don't want to make appraisal seem like it can only consider the spec's types.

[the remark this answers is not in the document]
deeglaze: I don't think actx should contain ar or ARS since that's a phase after the ACS is fully constructed. actx should just be for processing determined-good inputs to a final ACS.

draft-ietf-rats-corim.md
```diff
@@ -1733,63 +1741,111 @@ An ARS is a list of ECTs that describe ACS entries that are selected for use as
...
+~~~ cddl
+{::include cddl/intrep-actx.cddl}
+~~~
 
-Any CoRIM that has been secured by a cryptographic mechanism, such as a signature, that fails validation MUST be discarded.
+### Input Collection {#sec-phase1-collect}
 
-Other selection criteria MAY be applied.
-For example, if the Evidence format is known in advance, CoRIMs using a profile that is not understood by a Verifier can be readily discarded.
+The exchange of a request for attestation appraisal for a response of Attestation Results corresponds to a single Attestation Session.
```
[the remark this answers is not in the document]
deeglaze: It's not. The results are produced at the end of a session. An Appraisal Context is only for getting to the end of constructing the ACS. It's a bit matryoshka doll-ish.

draft-ietf-rats-corim.md
```diff
@@ -1998,11 +2029,22 @@ If the Evidence does not have a value for the mandatory `ae` fields, the Verifie
 Evidence transformation algorithms may be well-known, defined by a CoRIM profile ({{sec-corim-profile-types}}), or supplied dynamically.
 The handling of dynamic Evidence transformation algorithms is out of scope for this document.
 
+### Appraisal hermeticity
+
+The Appraisal Context at the and of Phase 1 constitutes all inputs to the Appraisal Procedure.
+
+Given the same Appraisol Context, different Verifier appraisals MUST produce deterministic results for phases 2, 3, and 4.
```
yogeshbdeshpande, line 2036:
```suggestion
Given the same Appraisal Context, different Verifier appraisals MUST produce deterministic results for phases 2, 3, and 4.
```

draft-ietf-rats-corim.md
```diff
@@ -1759,63 +1768,85 @@ An ARS is a list of ECTs that describe ACS entries that are selected for use as
...
+In Phase 1 the Verifier constructs an Appraisal Context that contains valid inputs to the Appraisal Procedure.
+The Appraisal Context is available to the Appraisal Procedure throughout the various phases.
+
+~~~ cddl
+{::include cddl/intrep-actx.cddl}
+~~~
+
+### Input Collection {#sec-phase1-collect}
+
+The exchange of a request for attestation appraisal for a response of Attestation Results corresponds to a single Attestation Session.
```
yogeshbdeshpande, line 1782:
```suggestion
A single attestation session is a period when a attestation appraisal process is initiated upon Verifier receiving Evidence from Attester or Relying Party and generating the results of the Appraisal.
```

draft-ietf-rats-corim.md
```diff
@@ -1759,63 +1768,85 @@ An ARS is a list of ECTs that describe ACS entries that are selected for use as
...
 
 Other selection criteria MAY be applied.
-For example, if the Evidence format is known in advance, CoRIMs using a profile that is not understood by a Verifier can be readily discarded.
+For example, if the Evidence format is known in advance, CoRIMs using a profile that is not understood by a Verifier for that Evidence format MAY be discarded.
+Selection policies model the trust relationships between the Verifier Owner and the relevant suppliers, and are out of the scope of the present document.
+For example, a composite device ({{Section 3.3 of -rats-arch}}) is likely to be fully described by multiple CoRIMs, each signed by a different supplier.
+In such case, the Verifier Owner may instruct the Verifier to discard tags activated by supplier CoBOMs that are not also activated by the trusted integrator.
 
-The selection process MUST yield at least one usable tag.
+> The selection process MUST yield at least one usable conceptual message.
```
yogeshbdeshpande, line 1816: Need to discuss this..

draft-ietf-rats-corim.md
```diff
@@ -2341,7 +2400,7 @@ If a triple condition does not match, then the Verifier continues to process oth
 
 ### Reference Values Corroboration and Augmentation (Phase 3) {#sec-phase3}
 
-Reference Value Providers (RVP) publish Reference Values using the Reference Values Triple ({{sec-comid-triple-refval}}) which are transformed ({{sec-ref-trans}}) into an internal representation ({{sec-ir-ref-val}}).
+Reference Value Providers (RVP) publish Reference Values using the Reference Values Triple ({{sec-comid-triple-refval}}) which are transformed ({{sec-ref-trans}}) into an internal representation ({{sec-ir-ref-val}}) of `rv` relations of `rv` relations.
```
nedmsmith, line 2403:
```suggestion
Reference Value Providers (RVP) publish Reference Values using the Reference Values Triple ({{sec-comid-triple-refval}}) which are transformed ({{sec-ref-trans}}) into an internal representation ({{sec-ir-ref-val}}) of `rv` relations.
```

nedmsmith:
Fixed build issues with this PR which included merging from main.
However, there are still issues with this PR having a different heading structure from the current main. Possibly, it makes sense to re-apply the changes starting from a current snapshot of main?

nedmsmith:
Some examples should be developed that test the new CDDL this PR introduces.

draft-ietf-rats-corim.md
```diff
@@ -1733,63 +1741,111 @@ An ARS is a list of ECTs that describe ACS entries that are selected for use as
...
+There will be some amount of CoRIMs and standalone tags available as inputs that make an `asession`:
 
-#### Tags Extraction and Validation
+~~~ cddl
+{::include cddl/intrep-asession.cddl}`
+~~~
+
+Initially all inputs are in `cms` if interpreted by this specification, or `extra` if not.
+
+It is left to Verifier Policy to determine if input sources must use supply chain transparency constructs (see {{-scitt-arch}}) to track input provenance.
```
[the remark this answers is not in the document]
deeglaze: I lost track of what this line used to refer to.

cddl/intrep-actx.cddl
```diff
@@ -0,0 +1,4 @@
+ir-appraisal-context = {
```
yogeshbdeshpande, line 1: I am not clear why this data structure is needed ..?

We should just have a Staging Area, kept somewhere out side of ACS and then highlight one concept that is known as ACS.

Active Context of Appraisal is the ACS Data Structure, which tracks the Appraisal.

Appraisal starts with reception of Evidence and ACS starts getting populated.

Then ACS keeps growing and at the end, ACS is converted into Attestation Results.

deeglaze: I have it as a data structure primarily to have a single object of process. We then say that Appraisal = phases567(leastfixedpoint(Phases234)(Phase1(inputs)))

if the staging area and ACS don't change after running through phases 2, 3, and 4, then there are no more conditional rules that can apply, and you're ready to move to the later phases.

yogeshbdeshpande (changes requested):
Please see my latest comments!

PR #308, opened 2024-10-04
Add version-map comparison and fix svn comparison

deeglaze:
The svn comparison text states that an svn must be a uint, which it isn't. It also accounts for untagged uint as a possible value, which it isn't.

Add a section about version-map comparison requiring equality comparison only. We aren't expanding the notion of version-map to have $version-scheme-specific comparison algorithms.

### DIFF
--- a/cddl/svn-type-choice.cddl
+++ b/cddl/svn-type-choice.cddl
@@ -3,4 +3,4 @@ svn = svn-type
 min-svn = svn-type
 tagged-svn = #6.552(svn)
 tagged-min-svn = #6.553(min-svn)
-svn-type-choice = tagged-svn / tagged-min-svn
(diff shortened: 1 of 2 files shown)


nedmsmith:
> The svn comparison text states that an svn must be a uint

I think the text intends to say that the comparison applies to the svn value after the tagging is stripped away. That would mean the comparison applies to the uint part.

nedmsmith:
> It also accounts for untagged uint as a possible value, which it isn't.

To support different Evidence formats, it is anticipated that the CBOR tagging may not be supplied. It may be that the verifier comparison algorithm is given an untagged uint. It seems quite reasonable to support this.

thomas-fossati (approved):

henkbirkholz (approved):

nedmsmith (approved):

PR #313, opened 2024-10-09
Define raw public key

deeglaze:
Raw still needs an encoding. Taking this to mean the RawPublicKey encoding defined in TLS 1.3.

### DIFF
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -70,7 +70,6 @@ contributor:
 normative:
   RFC4122: uuid
   RFC5280: pkix-cert
-  RFC7250: raw-pk
   RFC7468: pkix-text
   RFC8610: cddl
   RFC9090: cbor-oids
@@ -1026,10 +1025,13 @@ A cryptographic key can be one of the following formats:
 * `tagged-cose-key-type`: CBOR encoded COSE_Key or COSE_KeySet.
   Defined in {{Section 7 of -cose}}.
 
+* `tagged-pkix-asn1der-cert-type`: a `bstr` of ASN.1 DER encoded X.509 public key certificate.
+  Defined in {{Section 4 of -pkix-cert}}.
+
 A cryptographic key digest can be one of the following formats:
 
-* `tagged-thumbprint-type`: a `digest` of a raw public key as defined in {{-raw-pk}}. The digest value may
-  be used to find the public key if contained in a lookup table.
+* `tagged-thumbprint-type`: a `digest` of a raw public key.
+  The digest value may be used to find the public key if contained in a lookup table.
 
 * `tagged-cert-thumbprint-type`: a `digest` of a certificate.
   The digest value may be used to find the certificate if contained in a lookup table.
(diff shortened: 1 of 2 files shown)


thomas-fossati:
It's SubjectPublicKeyInfo from RFC5280.

Haven't checked but we should have that defined somewhere already - I am pretty sure, because it's what we use for PSA/CCA

nedmsmith:
The corim spec and the TCG endorsement spec clarifies it as a PEM encoded SubjectPublicKeyInfo and points to Section 13 in RFC7468.

draft-ietf-rats-corim.md
```diff
@@ -70,6 +70,7 @@ contributor:
 normative:
   RFC4122: uuid
   RFC5280: pkix-cert
+  RFC7250: raw-pk
```
nedmsmith, line 73: This RFC doesn't define raw public key directly. Rather it defines: 
```
case RawPublicKey:
              opaque ASN.1_subjectPublicKeyInfo<1..2^24-1>;
```
But this seems to be referring to a TLS encoding of ASN.1 that is defined in RFC5280.

The key thumbprint refers to a digest of a raw key (as defined by `tagged-pkix-base64-key-type`) 
https://datatracker.ietf.org/doc/html/rfc7093 describes techniques for generating digests of SubjectPublicKeyInfo that is DER encoding. Since another PR added DER encoded key type. Possibly, it makes sense to define a thumbprint type that is the digest of the DER encoded SubjectPublicKeyInfo?

I'm not aware of an RFC that defines a thumbprint for a PEM encoded SubjectPublicKeyInfo, but it may be sufficient for say that any digest algorithm can be applied to the PEM key. There isn't a need to reference anything as the output of a hash algorithm is bytes (bstr).

nedmsmith:
It would be helpful to file an issue before creating a PR so that there is opportunity to explore the problem a bit and to determine what/if a PR is needed.

draft-ietf-rats-corim.md
deeglaze: I believe there is a need to describe the algorithm since key fingerprinting is ambiguous and folks will want to get around the rule that there should be no nontrivial key comparisons by converting their keys into multiple formats (also is there a meaningful difference to diverge from the more common term 'fingerprint' with 'thumbprint'?)

thomas-fossati: > I believe there is a need to describe the algorithm since key fingerprinting is ambiguous and folks will want to get around the rule that there should be no nontrivial key comparisons by converting their keys into multiple formats

Method 1 described in [S4.2.1.2 of RFC5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.2)?

> (also is there a meaningful difference to diverge from the more common term 'fingerprint' with 'thumbprint'?)

Maybe we could just call it "key identifier" and avoid any colloquialisms.

thomas-fossati (approved):
LGTM

nedmsmith (approved):

henkbirkholz (approved):
we'll come back to the use of normative language her via #317

Issue #317, opened 2024-10-11
Remove key lookup normative text around key digests


deeglaze:
>  The digest value may be used to find the public key if contained in a lookup table.

We've decided that crypto-key-type-choice can only be binary-compared, so lookups are not meaningful to appraisal.

deeglaze:
Key lookup may still be usable for resolving attest-key-triple keys to construct certificate paths. See proposed text in PR #318.

nedmsmith:
> I think we need to keep the language that keys are binary-compared. We can rely on the translation to internal representation to add all the representations of a key that may need to be compared.

I think there are three cases where binary-compared keys makes sense; (i) environment-map.instance-id|group-id, (ii) measurement-values-map.cryptokeys, (iii) in conditions containing authorized-by. The cases where translation and lookup make sense are with identity-triples and attest-key-triples. The most compelling use case is when a key is a thumbprint and key verification is needed. The thumbprint can be used as a kid to find a suitable key (typically the cert path - x5chain or x5bag) but the certificate / signature should include a kid so the verifier can just look it up as opposed to having to compute the digest. Nevertheless, the kid computation algorithm has to be common between the kid / thumbprint or they won't match. RFC5280 defines a sha1 for alg and RFC5758 defines one for ECDSA keys with SHA2 variants.

yogeshbdeshpande:
@henkbirkholz  will take care of this prior to IETF 124

Issue #314, opened 2024-10-09
Attest-key-triple validation needed for ae construction?


deeglaze:
If we're dropping the need for RVPs to specify the attester authority, we ought to mention in Phase 1 that Evidence can only be transformed an ECT about environment E only if the evidence was signed by a key bound to E with an attest-key-triple, no? It probably works differently for SPDM since the measurement is over an authenticated channel rather than signed, but unfortunately that detail is missing from the TCG document https://trustedcomputinggroup.org/wp-content/uploads/TCG-DICE-Concise-Evidence-Binding-for-SPDM-Version-1.0-Revision-54_pub.pdf

nedmsmith:
> It probably works differently for SPDM since the measurement is over an authenticated channel rather than signed

SPDM defines something called a "transcript" which is something like a signed block that can be taken out of the channel context to be used to verify the integrity of offline measurements. They could be forwarded to a proxy to verify etc. This seems a bit like an IETF "token" but the spec isn't written that way.

nedmsmith:
> but unfortunately, that detail is missing from the TCG document 

The TCG binding spec anticipates that a transcript can be used.
The spec doesn't preclude use of a token or a detached signature (e.g., EAT) as the `spdm-toc-map` is extensible.

deeglaze:
Do we have any plan for normative language about attest-key-triple and how it's expected to bind an evidence-signing key to an environment?

nedmsmith:
> Do we have any plan for normative language about attest-key-triple and how its use is meant to bind an evidence-signing key to an environment?

By "bind" do you mean some form of proof-of-possession?

deeglaze:
No I mean more of a binding contract in the Verifier to not accept evidence from keys not associated with the environment the reference values are for. I want to make sure that if I’m not specifying authorized-by for my reference values, then the environment must be tied to the manufacturer RoT somehow

nedmsmith:
[...]

PR #318, opened 2024-10-11
Cryptographic validation of Evidence with attest-key-triple

deeglaze:
Validation is specific to a method. This reorganizes example methods into subsections and adds a new example based on attest-key-triple and CoTS.

This is an attempt to make actionable "key lookup" normative language as marked in Issue #317.

This is an attempt to resolve the ae verification concern in Issue #314.

### DIFF
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -1824,14 +1824,6 @@ If Evidence is cryptographically signed, its validation is applied before transf
 If Evidence is not cryptographically signed, the security context of the conveyance protocol that collected it is used to cryptographically validate Evidence.
 
 The way cryptographic signature validation works depends on the specific Evidence collection method used.
-For example, in DICE, a proof of liveness is carried out on the final key in the certificate chain (a.k.a., the alias certificate).
-If this is successful, a suitable certification path is looked up in the Appraisal Context, based on linking information obtained from the DeviceID certificate.
-See Section 9.2.1 of {{DICE.Layer}}.
-If a trusted root certificate is found, the usual X.509 certificate validation is performed.
-
-As a second example, in PSA {{-psa-token}} the verification public key is looked up in the appraisal context using the `ueid` claim found in the PSA claims-set.
-If found, COSE Sign1 verification is performed accordingly.
-
 Regardless of the specific integrity protection method used, the Evidence's integrity MUST be validated successfully.
 
 > If a CoRIM profile is supplied, it MUST describe:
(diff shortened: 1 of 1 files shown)


draft-ietf-rats-corim.md
```diff
@@ -1840,6 +1832,26 @@ Regardless of the specific integrity protection method used, the Evidence's inte
...
+In PSA {{-psa-token}} the verification public key is looked up in the appraisal context using the `ueid` claim found in the PSA claims-set.
+If found, COSE Sign1 verification is performed accordingly.
+
+##### Example: `attest-key-triple`
+
+A more general method of verifying evidence purely with CoRIM constructs is by verifying the signature with a trusted attestation key associated with the Evidence's Target Environment.
+The Verifier MAY use environment-identifying elements of unverified Evidence to determine the Target Environment.
+The Verifier MAY establish trust in an attestation key by constructing a certificate path to an accepted trust anchor whose constraints apply to the Target Environment.
+For example, the Verifier has accepted a CoTS tag {{-ta-store}} with an environment that matches the `attest-key-triple`'s `environment-map` and a listed purpose is `"eat"`.
+The Verifier MAY associate an attestation key with the Target Environment if a valid `attest-key-triple`'s `environment-map` successfully compares with the Evidence's Target Environment.
```
thomas-fossati, line 1851: Given this topic is still relatively fresh, I understand the reluctance to add strong statements.  However, these MAYs totally lack bite.  I think a better way to phrase this is to state clear requirements under a set of well-defined circumstances.  For example:

> If `attest-key-triple`s are used during phase 1 for evidence verification, then:
> * The verifier MUST ...
> * The verifier SHOULD ...

deeglaze: Okay, if I use this method in the SEV-SNP CoRIM profile, then I can add that normative language for the profile, but this is still an example, and that profile doesn't _actually_ use CoRIMs in this way.

For that profile I will still be making suggestions that Verifiers include a CoTS tag binding the SEV-SNP attestation class-id for the AMD-published keys as an implicit input, and then treating VCEK certificates from KDS as "virtual" attest-key-triples, since they don't issue CoRIMs.

When I describe translating an ATTESTATION_REPORT to claim structure that lines up with reference values in the CoRIM profile, I still describe the whole verification scheme.

It all just feels more like a mental exercise than an actual use of the CoRIM representation.

thomas-fossati: We seem to have similar requirements for PSA/CCA.  Maybe we could sit down and see if we can extract a core set of requirements that are profile-independent.

nedmsmith: The verifier will accept Evidence assertions so long as the channel to the Attester is authenticated (but not attested). Attest-key triple is not second guessing this assumption.

Phase 3 (corroboration of ref values) may be a better phase to process attest-key triples since the enforcement of a key measurement that fails key verification would reasonably result in that measurement NOT being corroborated. 

It might make sense to require the RIM author to include RV triples that do binary compare operations as a pre-requisite to applying attest-key processing.

deeglaze: RV triples that do binary compare operations of what? I'm not following.

nedmsmith, line 1851:
```suggestion

```

draft-ietf-rats-corim.md
```diff
@@ -1824,14 +1824,6 @@ If Evidence is cryptographically signed, its validation is applied before transf
...
 The way cryptographic signature validation works depends on the specific Evidence collection method used.
-For example, in DICE, a proof of liveness is carried out on the final key in the certificate chain (a.k.a., the alias certificate).
-If this is successful, a suitable certification path is looked up in the Appraisal Context, based on linking information obtained from the DeviceID certificate.
-See Section 9.2.1 of {{DICE.Layer}}.
-If a trusted root certificate is found, the usual X.509 certificate validation is performed.
-
-As a second example, in PSA {{-psa-token}} the verification public key is looked up in the appraisal context using the `ueid` claim found in the PSA claims-set.
-If found, COSE Sign1 verification is performed accordingly.
-
 Regardless of the specific integrity protection method used, the Evidence's integrity MUST be validated successfully.
```
nedmsmith, line 1827:
```suggestion
Nevertheless, Evidence integrity MUST be validated successfully.
```

draft-ietf-rats-corim.md
```diff
@@ -1840,6 +1832,26 @@ Regardless of the specific integrity protection method used, the Evidence's inte
...
+If found, COSE Sign1 verification is performed accordingly.
+
+##### Example: `attest-key-triple`
+
+A more general method of verifying evidence purely with CoRIM constructs is by verifying the signature with a trusted attestation key associated with the Evidence's Target Environment.
+The Verifier MAY use environment-identifying elements of unverified Evidence to determine the Target Environment.
+The Verifier MAY establish trust in an attestation key by constructing a certificate path to an accepted trust anchor whose constraints apply to the Target Environment.
+For example, the Verifier has accepted a CoTS tag {{-ta-store}} with an environment that matches the `attest-key-triple`'s `environment-map` and a listed purpose is `"eat"`.
+The Verifier MAY associate an attestation key with the Target Environment if a valid `attest-key-triple`'s `environment-map` successfully compares with the Evidence's Target Environment.
+If the `$crypto-key-type-choice` is an identifier type (e.g., `tagged-thumbprint-type`) and the Verifier cannot retrieve the associated public key material, then the key identifier MUST be discarded.
```
nedmsmith, line 1852:
```suggestion

```

This logic shouldn't apply when crypto-key-type-choice is used as a measurement-values-map value. If the value is corroborated by RVs then that should be good enough.

If the context is a key verification triple like attest-key-triple, then it makes sense to consider this text as part of the processing steps. But this is one of many possible failure conditions. Maybe there isn't a need to callout this condition explicitly?

nedmsmith:
Does PR #341 close this PR?

PR #325, opened 2024-10-15
Allow mkey tstr

### LINKED ISSUE #312: Allow `mkey`s to be `tstr`
@jraman567 has a use case that needs this.

An alternative would be for Jag's profile to extend `$measured-element-type-choice`, but it seems that `tstr` would be common enough to deserve being brought into the base schema.

nedmsmith:
Added tstr to $measured-element-type-choice.

Fix #312


deeglaze (approved):

yogeshbdeshpande (approved):
LGTM!

Issue #333, opened 2024-10-23
Nonsense: corim = tagged extensible type choice of tags 


deeglaze:
We need to remove the extensibility of $concise-rim-type-choice since extensibility is driven by profile, but profile isn't known at this point.

We also don't need to tag the type choice. Let's release the 500 tag?

`corim = concise-rim-type-choice`

thomas-fossati:
> We need to remove the extensibility of $concise-rim-type-choice since extensibility is driven by profile, but profile isn't known at this point.

Indeed, that `$` looks like a typo to me.  `$concise-rim-type-choice` should be a closed type choice which, per [our typographical conventions](https://ietf-rats-wg.github.io/draft-ietf-rats-corim/draft-ietf-rats-corim.html#section-3), should read `concise-rim-type-choice`.

> We also don't need to tag the type choice. Let's release the 500 tag?
> 
> `corim = concise-rim-type-choice`

The two-level tagging system effectively smells like overengineering.

I don't recall why we originally decided to use this tagging scheme.

While I am not opposed to dropping it, this could produce wire-image incompatibility with the TCG's spec, which we agreed must be avoided.

thomas-fossati:
> TCG Endorsement spec defines:
> 
> ```
> corim = #6.500($concise-reference-integrity-manifest-type-choice)
> tagged-corim-map = #6.501(corim-map)
> $concise-reference-integrity-manifest-type-choice /= tagged-corim-map
> $concise-reference-integrity-manifest-type-choice /= #6.502(signed-corim)
> 
> COSE-Sign1-corim = [
>   protected: bstr .cbor protected-corim-header-map
>   unprotected: unprotected-corim-header-map
>   payload: bstr .cbor tagged-corim-map
>   signature: bstr
> ]
> ```
> 
> The TCG is doing errata for this spec. Maybe it makes sense to influence this? But it would need to happen quickly and be minimally invasive.

Then I'd do:

```
corim = tagged-corim-map / tagged-cose-sign1-corim
```

which uses 501 for the unprotected and 18 for the sign1-protected `corim-map`.

deeglaze:
Is it common for 

> TCG Endorsement spec defines:
> 
> ```
> corim = #6.500($concise-reference-integrity-manifest-type-choice)
> tagged-corim-map = #6.501(corim-map)
> $concise-reference-integrity-manifest-type-choice /= tagged-corim-map
[...]

PR #337, opened 2024-10-23
Drop top level tagging requirement

deeglaze:
Tagged type choices are not typical.
I would go so far as to drop the 500 tag as the entrypoint to CoRIM altogether. NVIDIA is creating CoRIMs this way, but they are using a different content-type in the protected header. I think we can drop it in an follow-up.

This patch drops

* the need to tag the type choice
* the extensibility of concise-rim-type-choice, since extensibility is governed by a profile, and the profile is not known at this point in parsing.
* the need to tag the signed corim, since it is a COSE-sign1 with an unambigiuous content-type, and COSE-sign1 already has its own tag.

Addresses Issue #333, but 500 removal is TBD.


nedmsmith:
This PR probably should have some examples that tests the new functionality.

deeglaze:
Added a small change to the section about the specific #6.501 tag.

deeglaze:
> Note that this PR would allow four different tags (500, 501, 502, 18) instead of one (500) to represent a top-level CoRIM.

I think this is the pattern for CBOR though. There’s no single top level tag for COSE objects either.

500 and 502 are for theoretical encoding options that I don’t understand the justification for either.

thomas-fossati:
> > Note that this PR would allow four different tags (500, 501, 502, 18) instead of one (500) to represent a top-level CoRIM.
> 
> I think this is the pattern for CBOR though. There’s no single top level tag for COSE objects either.
> 
> 500 and 502 are for theoretical encoding options that I don’t understand the justification for either.

In fact, I'd drop 500 for sure, and may also be convinced to drop 502 if offered one or two pints of Guinness :-)

deeglaze:
We can say 502 is for signing envelopes that don't have their own CBOR tag or need further disambiguation.

500 is generally a confusing representation choice, so I think we've lost that one..

thomas-fossati:
> We can say 502 is for signing envelopes that don't have their own CBOR tag or need further disambiguation.
> 
> 500 is generally a confusing representation choice, so I think we've lost that one.. I can be convinced to have the payload always be 501-tagged for simplicity, it'll just invalidate cocli-signed CoRIMs. That's fine.

The easiest, and most consistent, would be to always require 501 (even when found in 502) and 502.

deeglaze:
Am I to understand that the MUST language in sec-corim-signed is not intended, and 502 is an anything-goes tag so long as the protected payload is a 501-tagged CBOR-encoded corim-map?

thomas-fossati:
> Am I to understand that the MUST language in sec-corim-signed is not intended,

Which MUST-language are you referring to?

> and 502 is an anything-goes tag so long as the protected payload is a 501-tagged CBOR-encoded corim-map?

Not sure, but at the moment we seem to have an inconsistency between CDDL and prose.  In S4.2 we say: _"payload: A CBOR encoded *tagged* CoRIM."_, whilst the CDDL allows a "naked" `corim-map` alongside the tagged one.  Besides, "CoRIM" should be "`corim-map`".

yogeshbdeshpande:
It was concluded in CoRIM Meeting 30th october to drop tag for #6.500 and propose the same change in TCG.

Also remove type extensibility from top level $concise-rim-type-choice, i.e. make it
only: `concise-rim-type-choice`

deeglaze:
We should return to this when the DICE errata has been accepted to remove the top level tags as well.

cddl/cbor-tags.txt
```diff
@@ -1,6 +1,6 @@
-tagged-concise-rim-type-choice = #6.500($concise-rim-type-choice)
+tagged-concise-rim-type-choice = #6.500(concise-rim-type-choice)
 tagged-corim-map = #6.501(corim-map)
-tagged-signed-corim = #6.502(signed-corim)
+tagged-signed-corim = #6.502($signed-corim)
```
deeglaze, line 3: I think we've agreed to drop the extensibility of the signing format.

draft-ietf-rats-corim.md
```diff
@@ -259,7 +259,7 @@ For more detail, see {{sec-corim-profile-types}}.
 
 A CoRIM can be signed ({{sec-corim-signed}}) using COSE Sign1 to provide end-to-end security to the CoRIM contents.
 When CoRIM is signed, the protected header carries further identifying information about the CoRIM signer.
-Alternatively, CoRIM can be encoded as a CBOR-tagged payload ({{sec-corim-map}}) and transported over a secure channel.
+Alternatively, CoRIM can be encoded as a #6.501 CBOR-tagged payload ({{sec-corim-map}}) and transported over a secure channel.
```
deeglaze, line 262: TODO: Drop 500, 502, deprecate 501 but allow it for the TCG spec conformance.

We change the IANA request to have a single content type, application/rim+cbor.

deeglaze:
Discussed in our meeting that we're going with 1 content type and dropping 500 and 502.

deeglaze:
@nedmsmith @henkbirkholz @thomas-fossati @yogeshbdeshpande @andrew-draper please take a look.

cddl/cbor-tags.txt
```diff
@@ -16,3 +14,4 @@ tagged-cert-thumbprint-type = #6.559(digest)
 tagged-bytes = #6.560(bytes)
 tagged-cert-path-thumbprint-type = #6.561(digest)
 tagged-pkix-asn1der-cert-type = #6.562(bstr)
+tagged-integral-predicate= #6.563(integral-predicate)
```
thomas-fossati, line 17: This looks spurious.

cddl/examples/corim-4.diag
```diff
@@ -0,0 +1,60 @@
...
+  / protected / << / header_map / {
+  / alg: / 1 : -35,
+  / content-type: / 3 : "application/rim+cbor",
+  / corim-meta: / 8 : << / corim-meta / {
+    / signer: / 0 : / corim-signer-map / {
+      / signer: / 0 : "ACME Ltd."
+    }
+  } >>,
+  / kid: / 4 : h'f8ccd2b49fdba32cd94498030fdc8e5010358919'
+} >>,
```
thomas-fossati, line 11: Indentation needs some care

cddl/examples/payload-corim-4.diag
```diff
@@ -0,0 +1,46 @@
+501(/ corim-map / {
```
thomas-fossati, line 1: send the corim-map to L2

Also, `ts=2 et`

draft-ietf-rats-corim.md
```diff
@@ -2690,78 +2688,17 @@ IANA is requested to add the following media types to the "Media Types"
...
-
-### corim-unsigned+cbor {#sec-mt-corim-unsigned}
-
-{:compact}
-Type name:
-: `application`
-
-Subtype name:
-: `corim-unsigned+cbor`
+: `rim+cbor`
```
thomas-fossati, line 2701: Two observations:

1. The registration is now invalid: we cannot just drop all the registration template fields except the subtype name
2. Let's call it `corim` rather than `rim` - we don't want to create confusion with TCG's "Reference Integrity Manifest (RIM)"

draft-ietf-rats-corim.md
thomas-fossati: > TODO: Drop 500, 502, deprecate 501 but allow it for the TCG spec conformance.
> 
> We change the IANA request to have a single content type, application/rim+cbor.

IIUC the gist of the proposed way forward is:

| media type | CBOR tag | data item |
| -- | -- | -- |
| `application/rim+cbor` | 501 | `tagged-corim-map` | 
| `application/rim+cbor` | 18 | `signed-corim` | 

This looks asymmetrical because in the case of `18`, absent a media type indication from the transport/wrapping channel, the parser needs to unpack the potential `signed-corim`, unpack the protected header, fetch the `content-type` element value and make sure it's `application/rim+cbor`.

I suggest (as I did in https://github.com/ietf-rats-wg/draft-ietf-rats-corim/pull/337#issuecomment-2447335956) to keep both 501 and 502 instead.

draft-ietf-rats-corim.md
nedmsmith: I didn't see that application/rim+cbor was taken. What made it invalid?

nedmsmith: > Let's call it `corim` rather than `rim` - we don't want to create confusion with TCG's "Reference Integrity Manifest (RIM)."

There are a variety of TCG specs that refer to themselves as "rim", but I don't believe any have registered media types. The DICE Endorsement Architecture spec (that defines corim) includes "application/rim+cbor" which is published (but didn't yet request a media type - this could be done immediately if necessary). 
TCG regards the published spec as having priority over subsequent specs. But it really comes down to what's registered by IANA. 
Pragmatically, should the TCG corim and the IETF corim have different media types. For any given release of the specs one spec may be a subset of the other's functionality (due to processing idiosyncrasies). Can they both have the same media type name? When the specs are aligned, and they have different media type names - is that desirable?

nedmsmith (approved):
I basically approve but need to resolve thread related to which media type string to request.

draft-ietf-rats-corim.md
deeglaze: The media type for signed-corim would be application/cose. The unprotected header for the payload's media type would be application/rim+cbor.

cddl/examples/payload-corim-4.diag
deeglaze: I don't understand your second comment. ts=2?

cddl/examples/corim-4.diag
deeglaze: Fixed?

cddl/cbor-tags.txt
deeglaze: Removed.

draft-ietf-rats-corim.md
```diff
@@ -1183,13 +1182,13 @@ If the search criteria are satisfied, the `endorsements` entries are asserted wi
 
 #### Conditional Endorsement Series Triple {#sec-comid-triple-cond-series}
 
-A Conditional Endorsement Series triple uses a "stateful environment" that identifies a Target Environment plus the measurements that have matching Evidence.	
+A Conditional Endorsement Series triple uses a "stateful environment" that identifies a Target Environment plus the measurements that have matching Evidence.
```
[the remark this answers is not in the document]
deeglaze: My text editor trims trailing whitespace automatically. Removed.

deeglaze:
I'm guessing we're delaying merging due to dropping 502, so I'm opening a dialogue with IANA MIME types and RFC9052 to consider adding content-type as an optional parameter for application/cose.

deeglaze:
Discussed in the thread that `application/rim+cose` with CoRE C-F registration and without the 502 tag in the CoRIM representation is the most viable path forward.

draft-ietf-rats-corim.md
thomas-fossati: **Update**

| media type | CBOR tag | data item |
| -- | -- | -- |
| `application/rim+cbor` | 501 | `tagged-corim-map` | 
| `application/rim+cose` | 18 | `signed-corim` | 

Also

* The content type (key=3) in the `protected-corim-header-map` is `application/rim+cbor`
* The `.corim` extension is registered for both `rim+cbor` and `rim+cose`

draft-ietf-rats-corim.md
```diff
@@ -2799,17 +2797,17 @@ Published specification:
 
 Applications that use this media type:
 : Attestation Verifiers, Endorsers and Reference-Value providers that need to
-  transfer unprotected CoRIM payloads over HTTP(S), CoAP(S), and other
+  transfer protected CoRIM payloads over HTTP(S), CoAP(S), and other
```
thomas-fossati, line 2800:
```suggestion
  transfer CoRIM payloads protected using COSE Sign1 over HTTP(S), CoAP(S), and other
```

draft-ietf-rats-corim.md
```diff
@@ -2799,17 +2797,17 @@ Published specification:
...
   transports.
 
 Fragment identifier considerations:
 : n/a
 
 Magic number(s):
-: `D9 01 F5`, `D9 01 F4 D9 01 F5`
+: `D2` (for COSE), and TBD given CoAP C-F assignment with TN transformation {{?RFC9277}} (e.g., for C-F 141 `D9 63 74 01 8F D2`).
```
thomas-fossati, line 2807: There is no TN here, it's just the tagged COSE Sign1:

```suggestion
: `D2 84`
```

cddl/examples/payload-corim-4.diag
```diff
@@ -0,0 +1,46 @@
...
+    / corim.id / 0 : h'284e6c3e5d9f4f6b851f5a4247f243a7',
```
thomas-fossati, line 2: I meant:

```suggestion
501(
  / corim-map / {
    / corim.id / 0 : h'284e6c3e5d9f4f6b851f5a4247f243a7',
```

cddl/examples/payload-corim-4.diag
thomas-fossati: see https://github.com/ietf-rats-wg/draft-ietf-rats-corim/pull/337/files#r1925086997

cddl/Makefile
```diff
@@ -42,6 +42,24 @@ CLEANFILES += $(3:.diag=.pretty)
...
+
+#examples/corim-4.sig: examples/sig-structure.cbor examples/testkey.pem
+#	openssl dgst -sha384 -sign examples/testkey.pem -out examples/corim-4.sig examples/sig-structure.cbor
+
+#examples/corim-4.diag: examples/corim-4.sig examples/corim-4.diag.tmpl examples/payload-corim-4.diag examples/protected-header-map.diag
+#	payload="$$(cat examples/payload-corim-4.diag)" \
+#	protected="$$(cat examples/protected-header-map.diag)" \
+#	signature="h'$$(cat examples/corim-4.sig | xxd -p -c 128)'" \
+#	envsubst < examples/corim-4.diag.tmpl > examples/corim-4.diag
+
```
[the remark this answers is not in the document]
thomas-fossati: Raised #364

cddl/cbor-tags.txt
thomas-fossati: Sure?

thomas-fossati (commented):
LGTM, apart from a couple of editorial fixes.

draft-ietf-rats-corim.md
```diff
@@ -2769,14 +2767,14 @@ Author/Change controller:
...
 {:compact}
 Type name:
 : `application`
 
 Subtype name:
-: `corim-unsigned+cbor`
+: `rim+cose`
 
 Required parameters:
 : n/a
```
thomas-fossati, line 2780: Here we should explicitly say that the `cose-type` parameter is not used.

draft-ietf-rats-corim.md
deeglaze: I thought you wanted the +cose CoAP C-F number to serialize the .corim at rest. The COSE_Sign1-only form is just for on-the-wire.

cddl/cbor-tags.txt
deeglaze: Oh it was an unadded file on my side that `make` readded. Removed.

draft-ietf-rats-corim.md
```diff
@@ -2591,10 +2590,9 @@ IANA is requested to allocate the following tags in the "CBOR Tags" registry {{!
 
 |     Tag | Data Item           | Semantics                                                     | Reference |
 |     --- | ---------           | ---------                                                     | --------- |
-|     500 | `tag`               | A tagged-concise-rim-type-choice, see {{sec-corim-tags}}      | {{&SELF}} |
+|     500 | `tag`               | Earmarked for CoRIM                                           | {{&SELF}} |
```
andrew-draper, line 2593:
```suggestion
|     500 | `tag`               | Reserved for backward compatibility                   | {{&SELF}} |
```

thomas-fossati (approved):
LGTM, thanks!

Issue #330, opened 2024-10-16
Transformation and processing sections needed for attest key and identity key triples


nedmsmith:
Also relates to #314 
A transformation section is needed for attest key and identity key triples that describes how attest/identity key triples are transformed into an internal representation.  

A processing section is needed that describes how the internal representation is processed. For example, it might impact phase 3 where a negative result of the AKT prevents the reference values from being added to the ACS.

Note: Given verifier behavior differs based on the type of key (identity vs. attest), there needs to be a way to represent the key type across external to internal processing. One approach is to modify the internal representation to include `key-type` info along with the `$crypto-key-type-choice` key.

nedmsmith:
There is a thread related to the semantics of what it means for a key to belong to an environment. 
The diag snippet below shows a key identifier (e.g., cert path) as a measurement of the environment class-id ending in ...39. This means the key is in the target environment named ...39. Since the private key is a secret the cert / public key serves as its alias. 
Example:
```
  / comid.triples / 4 : {
    / reference-triples / 0 : [
      [
        / environment-map / {
          / class / 0 : {
            / class-id / 0 :
              / tagged-uuid-type / 37( h'67b28b6c34cc40a19117ab5b05911e39' )
          }
        },
        [
          / measurement-map / {
            / mkey / 0 : "thing 2",
            / mval / 1 : {
              / cryptokeys / 13 : [
                / tagged-pkix-base64-cert-path-type / 556("base64_cert_path_X"),
                / tagged-pkix-base64-cert-path-type / 556("base64_cert_path_Y")
              ]
            }
          }
        ]
      ]
    ],
```
The next snippet shows an identity-triple that matches the keys that were named above as measurements and instructs the verifier to do key verification.
```
    / identity-triples / 2 : [ 
      [
        / environment-map / {
          / class / 0 : {
            / class-id / 0 :
              / tagged-uuid-type / 37( h'67b28b6c34cc40a19117ab5b05911e39' )
          }
        },
        / key-list / [
          / tagged-pkix-base64-cert-path-type / 556("base64_cert_path_X"),
          / tagged-pkix-base64-cert-path-type / 556("base64_cert_path_Y")
        ],
        / conditions / {
          / mkey / 0 : "thing 2"
        }
      ]
```
[...]

PR #341, opened 2024-10-26
Verify key triples

nedmsmith:
Addresses the majority of issue #330 (but one challenge remains).

### DIFF
--- a/cddl/attest-key-triple-record.cddl
+++ b/cddl/attest-key-triple-record.cddl
@@ -1,4 +1,6 @@
 attest-key-triple-record = [
-  environment-map
-  [ + $crypto-key-type-choice ]
+  environment: environment-map
+  ? mkey: $measured-element-type-choice
+  key-list: [ + $crypto-key-type-choice ]
+  ? authority-list: [ + $crypto-key-type-choice ]
 ]
--- a/cddl/identity-triple-record.cddl
+++ b/cddl/identity-triple-record.cddl
@@ -1,4 +1,6 @@
 identity-triple-record = [
-  environment-map
-  [ + $crypto-key-type-choice ]
+  environment: environment-map
+  ? mkey: $measured-element-type-choice
+  key-list: [ + $crypto-key-type-choice ]
+  ? authority-list: [ + $crypto-key-type-choice ]
 ]
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -114,6 +114,13 @@ informative:
   I-D.ietf-rats-eat: eat
   I-D.ietf-rats-concise-ta-stores: ta-store
   I-D.ietf-rats-ar4si: ar4si
+  DICE.cert:
+    title: DICE Certificate Profiles
+    author:
+      org: Trusted Computing Group
+    seriesinfo: Version 1.0, Revision 0.01
+    date: July 2020
+    target: https://trustedcomputinggroup.org/wp-content/uploads/DICE-Certificate-Profiles-r01_pub.pdf
 
 entity:
   SELF: "RFCthis"
@@ -683,7 +690,7 @@ The following describes each member of the `triples-map`:
   stateful environment records.
   Described in {{sec-comid-triple-cond-endors}}.
 
-##### Environments
+##### Environments {#sec-environments}
 
 An `environment-map` may be used to represent a whole Attester, an Attesting
 Environment, or a Target Environment.  The exact semantic depends on the
@@ -1203,49 +1210,68 @@ The first `series` entry that successfully matches the `selection` criteria term
 
 #### Device Identity Triple {#sec-comid-triple-identity}
 
-A Device Identity triple record relates one or more cryptographic keys to a device identity.
-The cryptographic keys are bound to or associated with a Target Environment that is within the device.
-The device identifier may be part of the Target Environment's `environment-map` or may be part of some other device identity credential, such as a certificate.
-The cryptographic keys are expected to be used to authenticate the device.
+A Device Identity triple (`identity-triples` in {{sec-comid-triples}}) relates one or more cryptographic keys to a device identity.
+The identity keys are bound to or associated with a Target Environment (as identified by `environment` and `mkey` - see below) within the device.
+The identity keys may be asserted via Evidence or Reference Values.
+
+The device identity keys may have been used to authenticate the Attester device or may be held in reserve for use at a later time.
 
 Device Identity triples instruct a Verifier to perform key validation checks, such as revocation, certificate path construction & verification, or proof of possession.
 The Verifier SHOULD verify keys contained in Device Identity triples.
 
-A Device Identity triple endorses that the keys were securely provisioned to the named Target Environment.
+A Device Identity triples endorse that the keys were securely provisioned to the named Target Environment.
 Additional details about how a key was provisioned or is protected may be asserted using Endorsements such as `endorsed-triples`.
 
 Depending on key formatting, as defined by `$crypto-key-type-choice`, the Verifier may take different steps to locate and verify the key.
 
-If a key has usage restrictions that limit its use to device identity challenges, Verifiers SHOULD check for key use that violates usage restrictions.
+If a key has usage restrictions that limit its use to device identity challenges, Verifiers SHOULD check for key use restrictions.
 
-Verification of a key, including its use restrictions, MAY produce Claims that are added to the ACS.
-Alternatively, Verifiers MAY report key verification results as part of an error reporting function.
+Each successful verification of a key in `key-list` SHALL produce Endorsement Claims that are added to the ACS.
+Claims are asserted with the joint authority of the Endorser (CoRIM signer) and the Verifier.
+Additionally, Verifiers MAY report key verification results as part of an error reporting function.
 
 ~~~ cddl
 {::include cddl/identity-triple-record.cddl}
 ~~~
 
+* `environment`: An `environment-map` condition used to identify the target Evidence or Reference Value.
+  See {{sec-environments}}.
+
+* `mkey`: An optional `$measured-element-type-choice` condition used to identify the element within the target Evidence or Reference Value.
+  See {{sec-comid-mkey}}.
+
+* `key-list`: A list of `$crypto-key-type-choice` keys that identifies which keys are to be verified.
+  See {{sec-crypto-keys}}.
+
+* `authority-list`: An optional list of `$crypto-key-type-choice` keys that identifies the authorities that asserted the key-list in the target Evidence or Reference Values.
+
 #### Attest Key Triple {#sec-comid-triple-attest-key}
 
-An Attest Key triple record relates one or more cryptographic keys to an Attesting Environment.
-The cryptographic keys are wielded by an Attesting Environment that collects measurements from a Target Environment.
-The cryptographic keys sign Evidence.
+An Attest Key triple (`attest-key-triples` in {{sec-comid-triples}}) relates one or more cryptographic keys to an Attesting Environment (as identified by `environment` and `mkey`).
+The cryptographic attestation keys are wielded by an Attesting Environment.
+Attestation keys may be asserted via Evidence or Reference Values.
+
+The attestation keys may have been used to sign Evidence or may be held in reserve for use at a later time.
 
 Attest Key triples instruct a Verifier to perform key validation checks, such as revocation, certificate path construction & verification, or proof of possession.
 The Verifier SHOULD verify keys contained in Attest Key triples.
 
-Attest Key triples endorse that the keys were securely provisioned to the named (identified via an `environment-map`) Attesting Environment.
+Attest Key triples endorse that the keys were securely provisioned to the named Attesting Environment.
 Additional details about how a key was provisioned or is protected may be asserted using Endorsements such as `endorsed-triples`.
 
 Depending on key formatting, as defined by `$crypto-key-type-choice`, the Verifier may take different steps to locate and verify the key.
-If a key has usage restrictions that limits its use to Evidence signing, Verifiers SHOULD check for key use that violates usage restrictions.
+If a key has usage restrictions that limits its use to Evidence signing (e.g., see Section 5.1.5.3 in [DICE.cert]).
+Verifiers SHOULD check for key use restrictions.
 
-Verification of a key, including its use restrictions, MAY produce Claims that are added to the ACS.
-Alternatively, Verifiers MAY report key verification results as part of an error reporting function.
+Each successful verification of a key in `key-list` SHALL produce Endorsement Claims that are added to the ACS.
+Claims are asserted with the joint authority of the Endorser (CoRIM signer) and the Verifier.
+Additionally, Verifiers MAY report key verification results as part of an error reporting function.
 
 ~~~ cddl
 {::include cddl/attest-key-triple-record.cddl}
 ~~~
+
+See {{sec-comid-triple-identity}} for additional details.
 
 #### Domain Dependency Triple {#sec-comid-triple-domain-dependency}
 
@@ -1896,7 +1922,7 @@ The selected tags are mapped to an internal representation, making them suitable
 
 #### Endorsement Triples Transformations {#sec-end-trans}
 
-Endorsed Values Triple Transformation :
+##### Endorsed Values Triple Transformation {#sec-end-trans-evt}
 
 {:ett-enum: counter="ett" style="format Step %d."}
 
@@ -1926,7 +1952,7 @@ Endorsed Values Triple Transformation :
 
 * If the Endorsement conceptual message has a profile, the profile is copied to the `ev`.`addition`.`profile` field.
 
-Conditional Endorsement Triple Transformation :
+##### Conditional Endorsement Triple Transformation {#sec-end-trans-cet}
 
 {:cett-enum: counter="cett" style="format Step %d."}
 
@@ -1959,7 +1985,7 @@ Conditional Endorsement Triple Transformation :
 
 * If the Endorsement conceptual message has a profile, the profile is copied to the `ev`.`addition`.`profile` field.
 
-Conditional Endorsement Series Triple Transformation :
+##### Conditional Endorsement Triple Transformation {#sec-end-trans-cest}
 
 {:cestt-enum: counter="cestt" style="format Step %d."}
 
@@ -1994,6 +2020,37 @@ Conditional Endorsement Series Triple Transformation :
 * The signer of the Conditional Endorsement conceptual message is copied to the `evs`.`series`.`addition`.`authority` field.
 
 * If the Endorsement conceptual message has a profile, the profile is copied to the `evs`.`series`.`addition`.`profile` field.
+
+##### Key Verification Triples Transformation {#sec-end-trans-kvt}
+
+[^issue] https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/330
+
+The following transformation steps are applied for both the `identity-triples` and `attest-key-triples`:
+
+{:kvt-enum: counter="ckvt" style="format Step %d."}
+
+{: kvt-enum}
+* An `ev` entry ({{sec-ir-end-val}}) is allocated.
+
+* The `cmtype` of the `evs` entry's `addition` ECT is set to `endorsements`.
+
+* Populate the `ev` `condition` ECT using either the `identity--triple-record` or `attest-key-triple-record` ({{sec-comid-triple-identity}}) as follows:
+
+{:kvt2-enum: counter="kvt2" style="format %i."}
+
+{: kvt2-enum}
+* **copy**(`environment-map`, `ev`.`condition`.`environment`.`environment-map`).
+
+* If populated, **copy**(`mkey`, `ev`.`condition`.`element-list`.`element-map`.`element-id`).
+
+* **copy**(`key-list`, `ev`.`condition`.`element-list`.`element-map`.`element-claims`.`measurement-values-map`.`cryptokeys`).
+
+* If populated, **copy**(`authority-list`, `ev`.`condition`.`authority`).
+
+{: kvt-enum}
+* The signer of the Identity or Attest Key Endorsement conceptual message is copied to the `ev`.`addition`.`authority` field.
+
+* If the Endorsement conceptual message has a profile, the profile is copied to the `ev`.`addition`.`profile` field.
 
 #### Evidence Tranformation
 
@@ -2151,7 +2208,7 @@ Endorsements are added to the ACS if the Endorsement condition is satisifed by t
 #### Processing Endorsements {#sec-process-end}
 
 Endorsements are matched with ACS entries by iterating through the `ev` list.
-For each `ev` entry, the `condition` ECT is compared with an ACS ECT, where the ACS ECT `cmtype` contains either `evidence` or `endorsements`.
+For each `ev` entry, the `condition` ECT is compared with an ACS ECT, where the ACS ECT `cmtype` contains either `evidence`, `reference-values`, or `endorsements`.
 If the ECTs match ({{sec-match-condition-ect}}), the `ev` `addition` ECT is added to the ACS.
 
 #### Processing Conditional Endorsements {#sec-process-cond-end}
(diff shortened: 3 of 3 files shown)


cddl/identity-triple-record.cddl
```diff
@@ -1,4 +1,6 @@
 identity-triple-record = [
-  environment-map
-  [ + $crypto-key-type-choice ]
+  environment: environment-map
+  ? mkey: $measured-element-type-choice
```
deeglaze, line 3: This optional mkey is a parsing annoyance. Can we move the non-environment fields to a map? Backward compatibility can just select between array and map.

draft-ietf-rats-corim.md
```diff
@@ -1203,50 +1210,69 @@ The first `series` entry that successfully matches the `selection` criteria term
...
+The identity keys are bound to or associated with a Target Environment (as identified by `environment` and `mkey` - see below) within the device.
+The identity keys may be asserted via Evidence or Reference Values.
+
+The device identity keys may have been used to authenticate the Attester device or may be held in reserve for use at a later time.
 
 Device Identity triples instruct a Verifier to perform key validation checks, such as revocation, certificate path construction & verification, or proof of possession.
 The Verifier SHOULD verify keys contained in Device Identity triples.
 
-A Device Identity triple endorses that the keys were securely provisioned to the named Target Environment.
+A Device Identity triples endorse that the keys were securely provisioned to the named Target Environment.
```
deeglaze, line 1222: I prefer the older wording

draft-ietf-rats-corim.md
```diff
@@ -1203,50 +1210,69 @@ The first `series` entry that successfully matches the `selection` criteria term
...
 Additional details about how a key was provisioned or is protected may be asserted using Endorsements such as `endorsed-triples`.
 
 Depending on key formatting, as defined by `$crypto-key-type-choice`, the Verifier may take different steps to locate and verify the key.
 
-If a key has usage restrictions that limit its use to device identity challenges, Verifiers SHOULD check for key use that violates usage restrictions.
+If a key has usage restrictions that limit its use to device identity challenges, Verifiers SHOULD check for key use restrictions.
```
deeglaze, line 1227: I find this harder to understand

draft-ietf-rats-corim.md
```diff
@@ -1203,50 +1210,69 @@ The first `series` entry that successfully matches the `selection` criteria term
...
+* `environment`: An `environment-map` condition used to identify the target Evidence or Reference Value.
+  See {{sec-environments}}.
+
+* `mkey`: An optional `$measured-element-type-choice` condition used to identify the element within the target Evidence or Reference Value.
+  See {{sec-comid-mkey}}.
+
+* `key-list`: A list of `$crypto-key-type-choice` keys that identifies which keys are to be verified.
+  See {{sec-crypto-keys}}.
+
+* `authority-list`: An optional list of `$crypto-key-type-choice` keys that identifies the authorities that asserted the key-list in the target Evidence or Reference Values.
```
deeglaze, line 1246:
```suggestion
* `authority-list`: An optional list of `$crypto-key-type-choice` keys that identifies the authorities that asserted the `key-list` in the target Evidence or Reference Values.
```

draft-ietf-rats-corim.md
```diff
@@ -1203,50 +1210,69 @@ The first `series` entry that successfully matches the `selection` criteria term
...
 The Verifier SHOULD verify keys contained in Attest Key triples.
 
-Attest Key triples endorse that the keys were securely provisioned to the named (identified via an `environment-map`) Attesting Environment.
+Attest Key triples endorse that the keys were securely provisioned to the named Attesting Environment.
 Additional details about how a key was provisioned or is protected may be asserted using Endorsements such as `endorsed-triples`.
 
 Depending on key formatting, as defined by `$crypto-key-type-choice`, the Verifier may take different steps to locate and verify the key.
-If a key has usage restrictions that limits its use to Evidence signing, Verifiers SHOULD check for key use that violates usage restrictions.
+If a key has usage restrictions that limits its use to Evidence signing (e.g., see Section 5.1.5.3 in [DICE.cert]).
+Verifiers SHOULD check for key use restrictions.
```
deeglaze, line 1264: Similar to prior comment. Would like to check for violations of the restriction

draft-ietf-rats-corim.md
```diff
@@ -1995,6 +2021,37 @@ Conditional Endorsement Series Triple Transformation :
...
+[^issue] https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/330
+
+The following transformation steps are applied for both the `identity-triples` and `attest-key-triples`:
+
+{:kvt-enum: counter="ckvt" style="format Step %d."}
+
+{: kvt-enum}
+* An `ev` entry ({{sec-ir-end-val}}) is allocated.
+
+* The `cmtype` of the `evs` entry's `addition` ECT is set to `endorsements`.
```
deeglaze, line 2035:
```suggestion
* The `cmtype` of the `ev` entry's `addition` ECT is set to `endorsements`.
```

draft-ietf-rats-corim.md
```diff
@@ -1995,6 +2021,37 @@ Conditional Endorsement Series Triple Transformation :
...
+
+* Populate the `ev` `condition` ECT using either the `identity--triple-record` or `attest-key-triple-record` ({{sec-comid-triple-identity}}) as follows:
```
deeglaze, line 2037:
```suggestion
* Populate the `ev` `condition` ECT using either the `identity-triple-record` or `attest-key-triple-record` ({{sec-comid-triple-identity}}) as follows:
```

cddl/identity-triple-record.cddl
nedmsmith: > This optional mkey is a parsing annoyance. Can we move the non-environment fields to a map? Backward compatibility can just select between array and map.

Can you suggest what the map might look like?

deeglaze: I want to first know why we need these additions to begin with. mkeys are meant for measured components of an attested environment, so there should not be different attestation keys at the mkey level. Why do we need authorities when attestation keys should be accepted based an trust anchors?

nedmsmith: > there should not be different attestation keys at the mkey level. Why do we need authorities when attestation keys should be accepted based an trust anchors?

The mkey and authority fields (along with the other fields) are condition values used for searching / matching ACS entries. The matched entry defines the context for the addition entry which is to add an endorsement under verifier authority for the subset of keys that satisfy verification checks. The addition ACS entry is otherwise identical to the matched ACS entry. Since the current ECT structure has mkey/element it makes sense to enable searching based on the mkey/element.

deeglaze: okay, then I'd suggest the following:

```
identity-triple = identity-triple-record / identity-triple-map
identity-triple-map = {
  &(environment: 0) => environment-map
  ? &(mkey: 1) => $measured-element-type-choice
  &(key-list: 2) => [ + $crypto-key-type-choice ]
  ? &(authority-list: 3) => [ + $crypto-key-type-choice ]
}
```
and not change identity-triple-record.

nedmsmith: What is wrong with the current proposal?

deeglaze: Suppose $measured-element-type-choice is a list of tagged bytes that get misinterpreted as crypto-key-type-choice. Do you have mkey,key-list or key-list,authority-list?

nedmsmith: The triples map defines the identity triple as an array of `identity-triple-record`. All of the triples are defined in terms of records (and not maps). This was discussed at length.

If the new fields are placed at the end of the existing fields then the ambiguity goes away.
```
? &(identity-triples: 2) =>
    [ + identity-triple-record ]

identity-triple-record = [
  environment: environment-map
  key-list: [ + $crypto-key-type-choice ]
  ? mkey: $measured-element-type-choice
  ? authority-list: [ + $crypto-key-type-choice ]
]
```
If mkey is omitted then the second list of crypto-key-type-choice is the authority-list. If mkey is included then it isn't confused with a key list.

deeglaze: I don't think the reordering avoids the ambiguity. I think you'd need to require that measured-element-type-choice must not be an array.

draft-ietf-rats-corim.md
```diff
@@ -1203,10 +1210,11 @@ The first `series` entry that successfully matches the `selection` criteria term
 
 #### Device Identity Triple {#sec-comid-triple-identity}
 
-A Device Identity triple record relates one or more cryptographic keys to a device identity.
-The cryptographic keys are bound to or associated with a Target Environment that is within the device.
-The device identifier may be part of the Target Environment's `environment-map` or may be part of some other device identity credential, such as a certificate.
-The cryptographic keys are expected to be used to authenticate the device.
+A Device Identity triple (`identity-triples` in {{sec-comid-triples}}) relates one or more cryptographic keys to a device identity.
+The identity keys are bound to or associated with a Target Environment (as identified by `environment` and `mkey` - see below) within the device.
```
deeglaze, line 1214:
```suggestion
The identity keys are bound to or associated with a Target Environment (as identified by `environment` and `mkey`—see below) within the device.
```

deeglaze: emdash

draft-ietf-rats-corim.md
```diff
@@ -2163,12 +2220,39 @@ Conditional endorsements have the same processing steps as shown in ({{sec-proce
...
+where for each `evs` entry, the `condition` ECT is compared with an ACS ECT, where the ACS ECT `cmtype` contains either `evidence`, `reference-values`, or `endorsements`.
 If the ECTs match ({{sec-match-condition-ect}}), the `evs` `series` array is iterated,
 where for each `series` entry, if the `selection` ECT matches an ACS ECT,
 the `addition` ECT is added to the ACS.
 Series processing terminates when the first series entry matches.
 
+#### Processing Key Verification Endorsements {#sec-process-keys}
+
+For each `ev` entry, the `condition` ECT is compared with an ACS ECT, where the ACS ECT `cmtype` contains either `evidence`, `reference-values`, or `endorsements`.
+If the ECTs match ({{sec-match-condition-ect}}), for each key in `ev`.`condition`.`element-claims`.`measurement-values-map`.`crypto-keys`:
```
deeglaze, line 2232: We're using a mix of . and / for following object structure. Looks like '.' is winning out, but please change the remaining / use "ACS processing requirements":

> measurement-map/authorized-by asserts that the actual state of one or more Claims within the Target Environment, as identified by environment-map, have the measurement values in measurement-map/mval.

draft-ietf-rats-corim.md
```diff
@@ -2163,12 +2220,39 @@ Conditional endorsements have the same processing steps as shown in ({{sec-proce
...
+
+* Verify the certificate signatures for the certification path.
+
+* Verify certificate revocation status for the certification path.
+
+* Verify key usage restrictions appropriate for the type of key.
+
+[^issue] https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/330
+
+* If key verification succeeds, **copy**(key, `ev`.`addition`.`element-list`.`element-map`.`element-claims`.`measurement-values-map`.`cryptokeys`).
```
deeglaze, line 2242: This is a type mismatch. Do you want [key] or do you want to append key into `cryptokeys`?

nedmsmith: It should be append.

draft-ietf-rats-corim.md
nedmsmith: should be an issue unless the cited text is part of the paragraphs touched by this PR.

draft-ietf-rats-corim.md
henkbirkholz: okay

draft-ietf-rats-corim.md
henkbirkholz: In which cases do you not follow the SHOULD as an implementer?

draft-ietf-rats-corim.md
henkbirkholz: @deeglaze seems to have fixed the last remaining "/" in #346

This is resolved.

draft-ietf-rats-corim.md
deeglaze:
```suggestion
* If key verification succeeds, **extend**(`ev`.`addition`.`element-list`.`element-map`.`element-claims`.`measurement-values-map`.`cryptokeys`, key).
```

Not sure what we're doing about adding to a list. Append? Extend?

draft-ietf-rats-corim.md
```diff
@@ -1216,37 +1224,55 @@ Additional details about how a key was provisioned or is protected may be assert
...
 The Verifier SHOULD verify keys contained in Attest Key triples.
 
-Attest Key triples endorse that the keys were securely provisioned to the named (identified via an `environment-map`) Attesting Environment.
+Attest Key triples endorse that the keys were securely provisioned to the named Attesting Environment.
 Additional details about how a key was provisioned or is protected may be asserted using Endorsements such as `endorsed-triples`.
 
 Depending on key formatting, as defined by `$crypto-key-type-choice`, the Verifier may take different steps to locate and verify the key.
-If a key has usage restrictions that limits its use to Evidence signing, Verifiers SHOULD check for key use that violates usage restrictions.
+If a key has usage restrictions that limits its use to Evidence signing (e.g., see Section 5.1.5.3 in [DICE.cert]).
+Verifiers SHOULD check for key use restrictions.
```
yogeshbdeshpande, line 1264:
```suggestion
Verifiers SHOULD enforce key use restrictions.
```

deeglaze (approved):

cddl/attest-key-triple-record.cddl
```diff
@@ -1,4 +1,8 @@
 attest-key-triple-record = [
-  environment-map
-  [ + $crypto-key-type-choice ]
+  environment: environment-map
+  key-list: [ + $crypto-key-type-choice ]
+non-empty<{
+   ? &(mkey: 0) => $measured-element-type-choice,
+   ? &(authorized-by: 2) => [ + $crypto-key-type-choice ] 
+}>
```
henkbirkholz, line 7: Not sure how the suggestion was pulled in. This was:

```suggestion
  ? mkey: $measured-element-type-choice
  ? authority-list: [ + $crypto-key-type-choice ]
```

draft-ietf-rats-corim.md
```diff
@@ -1995,6 +2021,37 @@ Conditional Endorsement Series Triple Transformation :
...
+{:kvt2-enum: counter="kvt2" style="format %i."}
+
+{: kvt2-enum}
+* **copy**(`environment-map`, `ev`.`condition`.`environment`.`environment-map`).
+
+* If populated, **copy**(`mkey`, `ev`.`condition`.`element-list`.`element-map`.`element-id`).
+
+* **copy**(`key-list`, `ev`.`condition`.`element-list`.`element-map`.`element-claims`.`measurement-values-map`.`cryptokeys`).
+
+* If populated, **copy**(`authority-list`, `ev`.`condition`.`authority`).
```
deeglaze, line 2048: Needs updating with the added nesting.

draft-ietf-rats-corim.md
```diff
@@ -2163,12 +2220,39 @@ Conditional endorsements have the same processing steps as shown in ({{sec-proce
...
+
+* Verify the certificate signatures for the certification path.
+
+* Verify certificate revocation status for the certification path.
+
+* Verify key usage restrictions appropriate for the type of key.
+
+[^issue] https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/330
+
+* If key verification succeeds, **extend**(`ev`.`addition`.`element-list`.`element-map`.`element-claims`.`measurement-values-map`.`cryptokeys`, key).
```
deeglaze, line 2242: I'm thoroughly confused what point an attest-key-triple serves if it is exactly the same as an endorsement of a cryptokey. It loses information about the key usage for verifying evidence to just being about having some key provisioned to it.

deeglaze (changes requested):

draft-ietf-rats-corim.md
deeglaze: @yogeshbdeshpande when PSA (not CCA) has its token verified with the IAK, "the verification public key is looked up in the appraisal context using the ueid claim found in the PSA claims-set."

This appraisal context is separate from the ACS? It's some unmentioned data structure that associates an environment with the key that should be used to verify the PSA? If not, and there is an ECT representation of it, then how is the transformation in #341 that strips the `attest-key-triple` quality to make it just another `cryptokey` measurement supposed to maintain the integrity of that lookup process?

yogeshbdeshpande: Hi @deeglaze : Veraison started when we did not had the concept of ACS within CoRIM draft, forget matching the exact data structure! However the concept of Appraisal context is the same as ACS. I would not prefer to explain too many details about Veraison implementation here, been this is a more generic CoRIM repo, but happy to explain it in a separate issue under Veraison repository!

deeglaze: Appraisal Context in this document is used as a nebulous, "stuff related to appraisal that we didn't represent in the ACS".
If we're making attest-key-triple mean something more than an endorsement of an (attesting) environment holding a key, we need to disambiguate it in the ACS to make the evidence verification section self-contained.

cddl/attest-key-triple-record.cddl
nedmsmith: > What was the difference between authorized-by and authority-list?

nothing.

draft-ietf-rats-corim.md
nedmsmith: The original text was actually wrong (append direction was reversed). The new text fixes this. 

The other thread doesn't seem related to the cited text. Maybe it belongs somewhere else?

nedmsmith: > If we're making attest-key-triple mean something more than an endorsement of an (attesting) environment holding a key, we need to disambiguate it in the ACS to make the evidence verification section self-contained.

We're not. The ACS has an ECT that associates the name of the key with an environment. This exists under the authority of the RVP as well as the Attester.

The attest-key-triple asked the verifier to do key verification. As a result, the ACS has a third ECT entry identical to the first two, this time under the authority of the Endorser and the Verifier. The Endorser is the entity asking to do the key verify checks, and the Verifier is the entity that did the check. 

Putting it all together:
step 1: Attester says key _X exists in environment _E.
step 2: RVP says key _X is supposed to exist in env _E. (aka step 1 is corroborated)
step 3: Endorser says key _X is supposed to be verified.
step 4: Verifier says key _X was verified. 

An RP may have a policy that wants to see all 4 steps. There are entries in the ACS for all 4.

deeglaze: The triples are named wrong, then. We shouldn't say anything about attesting or identity of the triple just means the signer demands a check to be done. I don't like that though, since there's still nothing that cleanly says "to get the attesting key for evidence of Target Environment E{layer n+1}, use the first cryptokey in E{layer n} (no mkey)"

A profile can of course say do something else, but as long as we're already translating everything to internal representations, I'd like to know in what senses I'm saying profile P's analog to evidence verification is "treat X as Y".

draft-ietf-rats-corim.md
```diff
@@ -1216,37 +1224,55 @@ Additional details about how a key was provisioned or is protected may be assert
...
 Attest Key triples instruct a Verifier to perform key validation checks, such as revocation, certificate path construction & verification, or proof of possession.
 The Verifier SHOULD verify keys contained in Attest Key triples.
 
-Attest Key triples endorse that the keys were securely provisioned to the named (identified via an `environment-map`) Attesting Environment.
+Attest Key triples endorse that the keys were securely provisioned to the named Attesting Environment.
 Additional details about how a key was provisioned or is protected may be asserted using Endorsements such as `endorsed-triples`.
 
 Depending on key formatting, as defined by `$crypto-key-type-choice`, the Verifier may take different steps to locate and verify the key.
-If a key has usage restrictions that limits its use to Evidence signing, Verifiers SHOULD check for key use that violates usage restrictions.
+If a key has usage restrictions that limits its use to Evidence signing (e.g., see Section 5.1.5.3 in [DICE.cert]).
```
thomas-fossati, line 1263:
```suggestion
If a key has usage restrictions that limits its use to Evidence signing (e.g., see Section 5.1.5.3 in {{DICE.cert}}).
```

draft-ietf-rats-corim.md
```diff
@@ -1216,37 +1224,55 @@ Additional details about how a key was provisioned or is protected may be assert
...
+* `authorized-by`: An optional list of `$crypto-key-type-choice` keys that identifies the authorities that asserted the `key-list` in the target Evidence or Reference Values.
+
 #### Attest Key Triple {#sec-comid-triple-attest-key}
 
-An Attest Key triple record relates one or more cryptographic keys to an Attesting Environment.
-The cryptographic keys are wielded by an Attesting Environment that collects measurements from a Target Environment.
-The cryptographic keys sign Evidence.
+An Attest Key triple (`attest-key-triples` in {{sec-comid-triples}}) relates one or more cryptographic keys to an Attesting Environment (as identified by `environment` and `mkey`).
+The cryptographic attestation keys are wielded by an Attesting Environment.
+Attestation keys may be asserted via Evidence, Reference Values, or Endorsements.
```
yogeshbdeshpande, line 1252:
```suggestion

```

draft-ietf-rats-corim.md
```diff
@@ -1216,37 +1224,55 @@ Additional details about how a key was provisioned or is protected may be assert
...
+
+The attestation keys may have been used to sign Evidence or may be held in reserve for use at a later time.
 
 Attest Key triples instruct a Verifier to perform key validation checks, such as revocation, certificate path construction & verification, or proof of possession.
 The Verifier SHOULD verify keys contained in Attest Key triples.
 
-Attest Key triples endorse that the keys were securely provisioned to the named (identified via an `environment-map`) Attesting Environment.
+Attest Key triples endorse that the keys were securely provisioned to the named Attesting Environment.
```
deeglaze, line 1259: I would like this to be the introductory line of the section.

deeglaze: Same for the device identity key section above.

nedmsmith:
Additional use case background on verify-key-triple:
A centralized RoT architecture has a single (monolithic) RoT that is either trusted or not. A distributed RoT architecture has multiple trusted components that are linked cryptographically. The bootstrap flow enables the RoT components but doesn't require each to be an AE. As RoT components they are not updatable / measurable other than that their public keys can be reported as evidence. 

User / Verifier will never see data / evidence signed by these keys. However, if a key is revoked, it would be useful information to a RP who is managing security risk to know if such revocation happened. This is not an architecture unique to Intel as CXL devices will have this type of key. 

The RefVal triple and concise-evidence triple can validate existence of the keys. However, it is up to downstream processing to determine if any of the keys pose an additional risk as the Verifier would not automatically discover existence of a compromised key. 

As an optional to implement triple, a verifier could step up to do these checks (possibly as a way to achieve better scalability vs. pushing it off to downstream RPs? Some other endorsement specification could be used to achieve similar objectives, but that wouldn't change the performance / security dynamics. 

Completely ignoring it is another option, but trades security for performance where the cost of doing the security check is easily quantifiable. It seems clear that there will be customers interested in doing the checking. The question is what role should corim play and when.

draft-ietf-rats-corim.md
```diff
@@ -1203,10 +1210,11 @@ The first `series` entry that successfully matches the `selection` criteria term
 
 #### Device Identity Triple {#sec-comid-triple-identity}
 
-A Device Identity triple record relates one or more cryptographic keys to a device identity.
-The cryptographic keys are bound to or associated with a Target Environment that is within the device.
-The device identifier may be part of the Target Environment's `environment-map` or may be part of some other device identity credential, such as a certificate.
-The cryptographic keys are expected to be used to authenticate the device.
+A Device Identity triple (`identity-triples` in {{sec-comid-triples}}) relates one or more cryptographic keys to a device identity.
+The identity keys are bound to or associated with a Target Environment (as identified by `environment` and `mkey`—see below) within the device.
```
deeglaze, line 1214:
```suggestion
The identity keys are bound to or are associated with a Target Environment (as identified by `environment` and `mkey`—see below) within the device.
```

draft-ietf-rats-corim.md
```diff
@@ -2179,16 +2264,16 @@ Additionally, the creation of Attestation Results is out-of-scope for this docum
 Phase 5: Verifier Augmentation
 
 Claims related to Verifier-applied consistency checks are asserted under the authority of the Verifier.
-For example, the `attest-key-triple-record` may contain a cryptographic key to which the Verifier applies certificate path construction and validation.
-Validation may reveal an expired certificate.
-The Verifier implementation might generate a certificate path validation exception that is handled externally, or it could generate a Claim that the certificate path is invalid.
+For example, the Verifier may supply evidence freshness nonces to the Attester to be included in Evidence.
+If a Verifier nonce is used, the Verifier may augment the ACS with a nonce Claim using Verifier authority.
```
deeglaze, line 2268: What is a nonce Claim? Is this something like but isn't exactly `eat_nonce`? Given that the Verifier is allowed to add anything it wants with its authority, maybe we should drop this sentence and allow the reader to decide what claims they want to add.

nedmsmith: This is offered as an example. It is already assumed that the reader / implementer will decide what claims they want to add. Since the previous exemplary text is addressed as part of the triples, it isn't a good example.

nedmsmith (commented):
multiple comments.

deeglaze (approved):

henkbirkholz (approved):

yogeshbdeshpande (approved):
LGTM!

thomas-fossati (approved):

PR #351, opened 2024-11-27
Clarify that triples may be ignored/discarded

### LINKED ISSUE #350: Clarify that normative triples that the Verifier does not understand MAY be ignored
There will be extensions in later specs, and there are some heavy base spec triples that may not always be needed. We should state what the expected behavior is.

deeglaze:
Fixes #350

### DIFF
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -1506,6 +1506,8 @@ The CoRIM profile MUST provide a description of the expected Verifier behavior f
 
 Verifier implementations MUST exhibit the same externally visible behavior as described in this specification.
 They are not required to use the same internal representation or evaluation order described by this specification.
+They are not required to implement support for all triples, profile permitting.
+A Verifier MAY discard a triple based on policy.
 
 ## Appraisal Procedure {#sec-appraisal-procedure}
 


draft-ietf-rats-corim.md
```diff
@@ -1506,6 +1506,8 @@ The CoRIM profile MUST provide a description of the expected Verifier behavior f
 
 Verifier implementations MUST exhibit the same externally visible behavior as described in this specification.
 They are not required to use the same internal representation or evaluation order described by this specification.
+They are not required to implement support for all triples, profile permitting.
```
yogeshbdeshpande, line 1509:
```suggestion
A Verifier may not implement all the Triples specified in this specification. It depends on its Appraisal Policy for Evidence as mandated by its profile.
```

draft-ietf-rats-corim.md
```diff
@@ -1506,6 +1506,8 @@ The CoRIM profile MUST provide a description of the expected Verifier behavior f
...
+A Verifier MAY discard a triple based on policy.
```
yogeshbdeshpande, line 1510:
```suggestion

```

deeglaze:
Discarding triples is left to later phases and not normative for up to phase 4.
Ignoring new triples is a matter of what is NTI, which @henkbirkholz is going to write what is NTI.

draft-ietf-rats-corim.md
yogeshbdeshpande: Pending M T I work!

PR #362, opened 2025-01-20
protected-header-map: kid: bit -> byte

pranjalkole:
`bstr` is a byte string according to the CDDL RFC

### DIFF
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -438,7 +438,7 @@ The following describes each child item of this map.
 
 * `content-type` (index 3): A string that represents the "MIME Content type" carried in the CoRIM payload.
 
-* `kid` (index 4): A bit string which is a key identity pertaining to the CoRIM Issuer.
+* `kid` (index 4): A byte string which is a key identity pertaining to the CoRIM Issuer.
 
 * `corim-meta` (index 8): A map that contains metadata associated with a signed CoRIM.
   Described in {{sec-corim-meta}}.


thomas-fossati (approved):
thanks @pranjalkole for spotting this!

yogeshbdeshpande (approved):
Thanks for spotting this!

deeglaze (approved):

yogeshbdeshpande:
@nedmsmith @henkbirkholz : Simple change, we can merge this today, if you both approve!

thomas-fossati:
> Simple change, we can merge this today, if you both approve!

Yes.

There's probably another conversation to have around the "kid" topic: why did we make it mandatory?

yogeshbdeshpande:
> > Simple change, we can merge this today, if you both approve!
> 
> Yes.
> 
> There's probably another conversation to have around the "kid" topic: why did we make it mandatory?

Sure, do you want to track it with another so as to discuss in next CoRIM Meeting ?

deeglaze:
> > Simple change, we can merge this today, if you both approve!
> 
> Yes.
> 
> There's probably another conversation to have around the "kid" topic: why did we make it mandatory?

Don’t you need to bind the key to the contents so you can reissue a certificate for the signer and still have the contents understood? My endorsement format just had the whole key certificate as part of the signed contents and that makes it harder since you have to reissue both the cert and all the endorsements.

thomas-fossati:
> > > Simple change, we can merge this today, if you both approve!
> > 
> > 
> > Yes.
> > There's probably another conversation to have around the "kid" topic: why did we make it mandatory?
> 
> Don’t you need to bind the key to the contents so you can reissue a certificate for the signer and still have the contents understood?

I don't understand.  The _contents_ of what?  If the CoRIM's, they are (should be) independent of the CoRIM signer, no?

If one needs to reissue a short-term cert for the same signing key, it can do so easily.  Clearly, if you need to handle many keys, having a kid for fast lookup can speed up the functionality, but that seems to me like a matter of local policy (say you only have one key around) rather than forcing a requirements on everybody.

> My endorsement format just had the whole key certificate as part of the signed contents and that makes it harder since you have to reissue both the cert and all the endorsements.

May I ask you why you want the signer key's cert chain to be also signed?

deeglaze:
Moving discussion to Issue #363

yogeshbdeshpande:
> Moving discussion to Issue #363

Thank you @deeglaze

nedmsmith (approved):
LGTM

PR #370, opened 2025-01-29
Rename CoBOM to CoTL (Concise Tag List)

### LINKED ISSUE #369: Rename CoBOM to a suitable name
As per feedback from IETF RATS session, the name CoBOM which expands to Concise Bill of Materials is colliding with SBOM and also does not clearly reflect the scope of what is defined in CoBOM.

This needs to be changed.

I would propose a better name like

CoBOM ==> Changed to CoTL (Concise Tag List)

yogeshbdeshpande:
Fixes #369

### DIFF
--- a/cddl/concise-tl-tag.cddl
+++ b/cddl/concise-tl-tag.cddl
@@ -1,0 +1,6 @@
+concise-tl-tag = {
+  &(tag-identity: 0) => tag-identity-map
+  &(tags-list: 1) => [ + tag-identity-map ],
+  &(tl-validity: 2) => validity-map
+  * $$concise-tl-tag-extension
+}
--- a/cddl/cbor-tags.txt
+++ b/cddl/cbor-tags.txt
@@ -1,7 +1,7 @@
 tagged-corim-map = #6.501(corim-map)
 tagged-concise-swid-tag = #6.505(bytes .cbor concise-swid-tag)
 tagged-concise-mid-tag = #6.506(bytes .cbor concise-mid-tag)
-tagged-concise-bom-tag = #6.508(bytes .cbor concise-bom-tag)
+tagged-concise-tl-tag = #6.508(bytes .cbor concise-tl-tag)
 tagged-ueid-type = #6.550(ueid-type)
 tagged-svn = #6.552(svn)
 tagged-min-svn = #6.553(min-svn)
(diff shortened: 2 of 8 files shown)


nedmsmith:
were there no examples that tested the cobom cddl? If not, there probably should be.

yogeshbdeshpande:
> were there no examples that tested the cobom cddl? If not, there probably should be.

No not to my knowledge!

deeglaze:
This bike shed color is acceptable.

thomas-fossati:
> were there no examples that tested the cobom cddl? If not, there probably should be.

Added #371

yogeshbdeshpande:
> > were there no examples that tested the cobom cddl? If not, there probably should be.
> 
> Added #371

Thanks for creating this one, I shall try to include this probably in this PR itself...

cddl/concise-tl-tag.cddl
```diff
@@ -0,0 +1,6 @@
+concise-tl-tag = {
```
thomas-fossati, line 1: Warning: bikeshedding comment.

I'd prefer `concise-tag-list` because I'm not sure that CoTL is a "tag" in the sense that other Co{MID,SWID,TS} are.

yogeshbdeshpande: This would then beak the convention (we have as of today) that CoRIM contains a list of `tags`!

nedmsmith: I'm also OK with adding a new code point for `concise-tag-list` to `corim-map` as I agree with Thomas that I don't think these are the same type thing as comid, coswid, etc.
```
corim-map = {
...
? &(tag-list: 6) => [ + concise-tl-tag ]
...
}
```

yogeshbdeshpande: Lets separate two discussions.
1. Rename CoBOM to CoTL --> This is addressed here
2. Discuss whether CoTIL is a tag in itself or another data structure ? Discussed via Issue #372

nedmsmith (approved):
LGTM - but probably should have some examples.

nedmsmith (approved):

deeglaze (approved):

henkbirkholz (approved):

Issue #358, opened 2025-01-17
Signing one artifact for multiple profiles


deeglaze:
The CoRIM schema disallows multiple profiles, but also does not have a means of representing a bundle of CoRIMs each with their own profiles. This is troublesome when signing reference measurements for a firmware binary that is measured in different ways depending on the technology. Intel TDX, AMD SEV-SNP, and ARM CCA all have different ways of constructing a digest of a machine initial state that a firmware file like an IGVM describes how to load. I have no specified way to have a single document to deliver with the firmware that endorses each of the measurements that it could be seen with.

Either we bind this with CMW and a specific cmwc_t that specifies something like tag-identity => #6.<TN for corim>(<<corim-map>>)?

deeglaze:
The IGVM format allows us to create a single file with multiple headers on how to load the firmware depending on the TEE technology, so we should be able to have 1 CoSWID and multiple CoMIDs of diffferent profiles for the same firmware. My problem with the corim-locator-map is there is no format for a local bundle. But I did suggest we could put them all in a CMW if we agree on a cmwc_t for local bundling. I could be fine with that too, but it's more cumbersome given I'll need to carry multiple signatures instead of 1, which PQC can make pretty hefty.

Lifecycle management would be that we release a new firmware binary in the event of a problem with any technology, but we would bind TEE-specific information to the CoMID instead of the CoSWID. We don't really replace CoRIMs so much as we let old ones expire.

yogeshbdeshpande:
Thank you @deeglaze  for clarification. 
So far the definition of a Profile, implies the Profile defines the particular semantics/interpretation of not just the outer CoRIM Level but also implies semantics of internal CoMIDs and CoSWID it contains. Considering the scope, I would like to know, if assuming that there are fixed possibilities of Attestation Technologies, could `N` finite number of Profile be possibly defined by the maintainer of the Outer CoRIM blob?

To explain, what I mean, let us say, a Owner Organisation-X has four possibilities.

Possibility (A) : 1 - CoRIM - 2 CoMIDs one for TDX and one for SEV-NP 
Possibility (A):  Profile 1 - `tag:X.com:2025:tdx+sev-snp#1.0.0`

Possibility (B) : 1 - CoRIM - 3 CoMIDs one for TDX, one for SEV-SNP and one for Arm CC
Possibility (B):  Profile 2 - `tag:X.com:2025:tdx+sev-snp+cca#1.0.0`

[...]

PR #376, opened 2025-02-12
Rename tagged-corim-map to tagged-unsigned-corim

deeglaze:
This allows for better interpreting the tag as an entrypoint to a CoRIM document. This fixes the hanging reference to unsigned-corim-map noted in Issue #374.

PR #377, opened 2025-02-12
Language for signer inheritance

deeglaze:
With the proposed additions to CMW, this fixes Issue #358.


draft-ietf-rats-corim.md
```diff
@@ -267,6 +267,8 @@ For more detail, see {{sec-corim-profile-types}}.
 A CoRIM can be signed ({{sec-corim-signed}}) using COSE Sign1 to provide end-to-end security to the CoRIM contents.
 When CoRIM is signed, the protected header carries further identifying information about the CoRIM signer.
 Alternatively, CoRIM can be encoded as a #6.501 CBOR-tagged payload ({{sec-corim-map}}) and transported over a secure channel.
+If a CoRIM is contained within a larger signed document or message, there MUST be a single signer assigned to the scope of the CoRIM as part of the document's specification or protocol conveying the message, respectively.
```
yogeshbdeshpande, line 270: I would phrase the statement something like this.

`A CoRIM can be used without signature protection (please see unsigned-corim-map),

For example, an application may use TLS to transmit an unsigned CORIM and specify that the signer
is the public key in the Certificate Message.

Alternatively an unsigned CoRIM can be a part of a RATS Message which itself is signed (example a signed CMW).
In such cases the authority of CoRIM is delegated to the signer encapsulating the CoRIM.`

yogeshbdeshpande (changes requested):
We also need to re-instate unsigned -corim-map and possibly the tag to identify it again!

yogeshbdeshpande (commented):
Just realised PR #376, so modifying the review!

yogeshbdeshpande (commented):
suggested changes

draft-ietf-rats-corim.md
```diff
@@ -267,6 +267,8 @@ For more detail, see {{sec-corim-profile-types}}.
...
+For example, an application may use TLS to transmit an unsigned CoRIM and specify that the signer is the public key in the Hello message.
```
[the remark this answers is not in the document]
deeglaze: I didn't want to make a circular citation since CMW references CoRIM.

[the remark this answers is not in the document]
thomas-fossati: No problem with that.  These are "weak" links since the references are both informative.

draft-ietf-rats-corim.md
```diff
@@ -481,6 +481,14 @@ Described in {{sec-common-validity}}.
 {::include cddl/unprotected-corim-header-map.cddl}
 ~~~
 
+## Signer authority of securely conveyed unsigned CoRIM {#sec-conveyed-signer}
+
+An unsigned (#6.501-tagged) CoRIM may be contained in a larger signed document.
```
steven-bellock, line 486: "document" usually means "specification". Maybe "payload" instead of "document"?

draft-ietf-rats-corim.md
```diff
@@ -266,7 +266,7 @@ For more detail, see {{sec-corim-profile-types}}.
 
 A CoRIM can be signed ({{sec-corim-signed}}) using COSE Sign1 to provide end-to-end security to the CoRIM contents.
 When CoRIM is signed, the protected header carries further identifying information about the CoRIM signer.
-Alternatively, CoRIM can be encoded as a #6.501 CBOR-tagged payload ({{sec-corim-map}}) and transported over a secure channel.
+Alternatively, an unsigned CoRIM can be encoded as a #6.501 CBOR-tagged payload ({{sec-corim-map}}) and transported over a secure channel and maintain a notion of a CoRIM signer ({{sec-conveyed-signer}}).
```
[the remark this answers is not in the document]
deeglaze: Are the parentheses necessary for the pointer?

draft-ietf-rats-corim.md
deeglaze: Accepted Ned's changes. Thanks.

draft-ietf-rats-corim.md
```diff
@@ -481,6 +481,14 @@ Described in {{sec-common-validity}}.
 {::include cddl/unprotected-corim-header-map.cddl}
 ~~~
 
+## Signer authority of securely conveyed unsigned CoRIM {#sec-conveyed-signer}
+
+An unsigned (#6.501-tagged) CoRIM may be contained in a larger signed document.
+An unsigned CoRIM may be conveyed over a secure channel.
+In both cases, there is a method of integrity protection that relies on some authentication.
+The role of "the CoRIM signer" MUST be specified in both cases, with the same constraint on the signed CoRIM that there is a single signer.
```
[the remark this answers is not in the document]
nedmsmith: > The role MUST be added? By the sender or receiver? This will be hard to enforce. Anyone transmitting a CoRIM has to transform it first?

I'm trying to reflect the normative that applies to corim-signer-map which is not optional when COSE-sign1-corim is used. The corim author (typically is also the signer). If "sender" means author, then those are the semantics I was envisioning. The corim author can modify the corim.entities normally. Originally, only corim.entities contained manifest-signer property but was removed when we defined COSE-sign1-corim. Given we're supporting other signature formats, it makes sense to put it back in. However, the CDDL doesn't make it mandatory so we need normative prose. It wouldn't be incorrect if manifest signer role info was always populated, even if COSE-sign1-corim was used. But it would be in two places at that point. A COSE-sign1-corim signer could remove `manifest-signer` from corim.entities if an upstream process supplied it (to remove unnecessary bytes).

If the `manifest-creator` knows who the `manifest-signer` is supposed to be then the manifest creator can populate corim.entities. If the signer is a COSE-sign1-corim entity, then they can choose to remove the rundundant bytes or just layer on the corim-meta-map as part of the COSE header. 

If the manifest creator doesn't know the manifest signer yet, then I don't see an alternative other than for the signer to modify the image. Possibly, the implementation has to define a template structure that pre-allocates space for the fix up?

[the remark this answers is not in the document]
deeglaze: If we have multiple builders all authenticated to an orchestrator and builders send unsigned reference values over a secure channel to the orchestrator, then the orchestrator sends the collection of reference values in a CMW to a signer, then it seems like I'm breaking your MUST rule because of the two hops. Is that an intentional restriction?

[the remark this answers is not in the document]
nedmsmith: > Is that an intentional restriction?

I don't think so. The first hop sounds more like authoring activity, in which case there might be multiple role entities added before it gets to the orchestrator. Since the orchestrator is the last to touch it before it goes to the Verifier that seems closest to original intent. The original discussions assumed COSE signing and stopped short of trying to capture deep suppy chain workflow use cases. Basically, if the orchestrator is the last stop before handing it over to a Verfier, then any of the other entities that may have been involved can be listed as `entities` but the signer is trusted to assemble the list of entities. Doing supply chain provenance (aka in-toto) was a non-goal). 

BTW: I'm interpreting your use of "CMW to a signer" as meaning to a Verifier. If you meant "signer" to be part of the supply chain before the CMW arrives at a Verifier, then the "signer" fills in the `entities` list based on the defined roles reserving the manifest-signer role for itself.

draft-ietf-rats-corim.md
```diff
@@ -490,6 +492,27 @@ The role of "the CoRIM signer" MUST be specified in both cases, with the same co
 
 It is out of scope of this document to specify a method of delegating the signer role in the case that an unsigned CoRIM is conveyed through multiple secured links with different notions of authenticity without end-to-end integrity protection.
 
+### CoRIM bundles
+
+A method of signing a bundle of CoRIMs together is through a signed RATS Conceptual Message Wrapper (CMW) {{-cmw}}.
+Let `tag:{{&SELF}}:bundle` name a collection type that when signed MUST include the `corim-meta` protected header.
+The `corim-meta` header ensures that each CoRIM in the bundle has an identified signer.
+
+The CMW MAY use any label for its CoRIMs.
```
nedmsmith, line 501: The following section seems too terse to follow. Possibly, more of the CDDL context is needed?

nedmsmith (commented):
See comments inline.

draft-ietf-rats-corim.md
```diff
@@ -490,6 +492,27 @@ The role of "the CoRIM signer" MUST be specified in both cases, with the same co
 
 It is out of scope of this document to specify a method of delegating the signer role in the case that an unsigned CoRIM is conveyed through multiple secured links with different notions of authenticity without end-to-end integrity protection.
 
+### CoRIM bundles
+
+A method of signing a bundle of CoRIMs together is through a signed RATS Conceptual Message Wrapper (CMW) {{-cmw}}.
+Let `tag:{{&SELF}}:bundle` name a collection type that when signed MUST include the `corim-meta` protected header.
```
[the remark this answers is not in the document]
deeglaze: "The CMW collection signature structure" sounds strange to me, since it's the COSE signing structure.

draft-ietf-rats-corim.md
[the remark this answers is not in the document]
deeglaze: I don't like the requirement to modify the internals of the map. It's unnecessary in the case of my bundle use, where the signature must be a COSE envelope and must use the corim-meta statement.

draft-ietf-rats-corim.md
deeglaze: Added.

draft-ietf-rats-corim.md
```diff
@@ -267,6 +267,8 @@ For more detail, see {{sec-corim-profile-types}}.
 A CoRIM can be signed ({{sec-corim-signed}}) using COSE Sign1 to provide end-to-end security to the CoRIM contents.
 When CoRIM is signed, the protected header carries further identifying information about the CoRIM signer.
 Alternatively, CoRIM can be encoded as a #6.501 CBOR-tagged payload ({{sec-corim-map}}) and transported over a secure channel.
+If a CoRIM is contained within a larger signed document or message, there MUST be a single signer assigned to the scope of the CoRIM as part of the document's specification or protocol conveying the message, respectively.
+For example, an application may use TLS to transmit an unsigned CoRIM and specify that the signer is the public key in the Certificate message.
```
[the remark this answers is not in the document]
deeglaze: Yogesh please double-check this against the current status.

draft-ietf-rats-corim.md
```diff
@@ -539,6 +541,41 @@ Described in {{sec-common-validity}}.
...
+An unsigned CoRIM may be a payload within a secure channel.
+The CoRIM signer authority is taken from the authenticated credential of the same entity that originates the CoRIM.
+A CoRIM role entry that contains the `manifest-signer` role MUST be added to `corim-entity-map`.
+
+It is out of scope of this document to specify a method of delegating the signer role in the case that an unsigned CoRIM is conveyed through multiple secured links with different notions of authenticity without end-to-end integrity protection.
+
+### CoRIM bundles
+
+A method of signing a bundle of CoRIMs together is through a signed RATS Conceptual Message Wrapper (CMW) {{-cmw}}.
+The COSE_Sign1 signature format can be used with a CMW collection.
```
nedmsmith, line 556:
```suggestion
The COSE_Sign1 signature format can be used with a Collection CMW.
```

draft-ietf-rats-corim.md
```diff
@@ -539,6 +541,41 @@ Described in {{sec-common-validity}}.
...
+The COSE protected header can include a CMW collection type name.
```
nedmsmith, line 557:
```suggestion
The COSE protected header can include a Collection CMW type name.
```

draft-ietf-rats-corim.md
```diff
@@ -539,6 +541,41 @@ Described in {{sec-common-validity}}.
...
+The collection type name SHALL be of the form: `tag:{{&SELF}}:bundle`.
+The signing operation MUST include the `corim-meta` in the COSE_Sign1 `protected-header` parameter.
+The `corim-meta` statement ensures that each CoRIM in the bundle has an identified signer.
+
+~~~ cddl
+{::include cddl/cmw-corim-bundle.cddl}
+~~~
+
+The CMW MAY use any label for its CoRIMs.
```
nedmsmith, line 566:
```suggestion
The Collection CMW MAY use any label for its CoRIMs.
```

draft-ietf-rats-corim.md
```diff
@@ -539,6 +541,41 @@ Described in {{sec-common-validity}}.
...
+If there is a hierarchical structure to the CoRIM bundle, the base entry point SHOULD be labeled `0` in CBOR or `"base"` in JSON.
+It is RECOMMENDED to use to label a CoRIM with its tag-id in string format, where `uuid-type` string format is specified by [RFC4122].
```
nedmsmith, line 568:
```suggestion
It is RECOMMENDED to label a CoRIM with its tag-id in string format, where `uuid-type` string format is specified by [RFC4122].
```

draft-ietf-rats-corim.md
```diff
@@ -539,6 +541,41 @@ Described in {{sec-common-validity}}.
...
+CoRIMs distributed in a bundle MAY declare their interdependence `dependent-rims` with local resource indicators.
```
nedmsmith, line 569:
```suggestion
CoRIMs distributed in a bundle SHOULD declare their interdependence using `dependent-rims` with local resource indicators.
```

draft-ietf-rats-corim.md
```diff
@@ -539,6 +541,41 @@ Described in {{sec-common-validity}}.
...
+It is RECOMMENDED that a CoRIM with a `uuid-type` tag-id be referenced with URI `urn:uuid:`_tag-id-uuid-string_.
+It is RECOMMENDED that a CoRIM with a `tstr` tag-id be referenced with `tag:{{&SELF}}:local,`_tag-id-tstr_.
+It is RECOMMENDED for a `corim-locator-map` containing local URI to afterwards list a nonzero amount of reachable URLs as remote references.
```
nedmsmith, line 572:
```suggestion
It is RECOMMENDED for a `corim-locator-map` containing local URIs to afterwards list a nonzero number of reachable URLs as remote references.
```

nedmsmith: I don't know what "afterwards" means.

draft-ietf-rats-corim.md
```diff
@@ -539,6 +541,41 @@ Described in {{sec-common-validity}}.
...
+
+The following example demonstrates these recommendations for bundling CoRIMs that are subject to different profiles.
```
nedmsmith, line 574:
```suggestion
The following example demonstrates these recommendations for bundling CoRIMs with a common signer but have different profiles.
```

draft-ietf-rats-corim.md
nedmsmith: Why is this not SHOULD?

draft-ietf-rats-corim.md
```diff
@@ -539,6 +541,41 @@ Described in {{sec-common-validity}}.
...
+The collection type name SHALL be of the form: `tag:{{&SELF}}:bundle`.
+The signing operation MUST include the `corim-meta` in the COSE_Sign1 `protected-header` parameter.
+The `corim-meta` statement ensures that each CoRIM in the bundle has an identified signer.
+
+~~~ cddl
+{::include cddl/cmw-corim-bundle.cddl}
+~~~
+
+The CMW MAY use any label for its CoRIMs.
+If there is a hierarchical structure to the CoRIM bundle, the base entry point SHOULD be labeled `0` in CBOR or `"base"` in JSON.
```
nedmsmith, line 567: I'm confused, shouldn't the label be a UUID? I don't see how its used in the example.

nedmsmith (commented):
see comments

draft-ietf-rats-corim.md
deeglaze: If there's a root CoRIM, there's nothing pointing to it, and you need a notion of where to start drilling down in the graph.

draft-ietf-rats-corim.md
deeglaze: because it's redundant. The tags are in the values of the collection CMW's records, so you have TAG-ID => {TAG-ID, ...}

draft-ietf-rats-corim.md
deeglaze: Afterwards just means the local uris come first and the reachable urls come after in the list

draft-ietf-rats-corim.md
```diff
@@ -539,6 +541,41 @@ Described in {{sec-common-validity}}.
 {::include cddl/unprotected-corim-header-map.cddl}
 ~~~
 
+## Signer authority of securely conveyed unsigned CoRIM {#sec-conveyed-signer}
+
+An unsigned (#6.501-tagged) CoRIM may be a payload in an enveloping signed document.
```
yogeshbdeshpande, line 546: Is line 546 referring to the CoRIM bundle use case ? If yes, could you please add a reference to the CoRIM Bundle subsection as we have two cases, listed here but explanation spreaded across

yogeshbdeshpande (approved):
In general LGTM, we can discuss few minor points on Wednesday and close this!

draft-ietf-rats-corim.md
deeglaze: the bundle is described in a subsection a few sentences later. Does it need a link?

draft-ietf-rats-corim.md
henkbirkholz: this is the way (currently)

draft-ietf-rats-corim.md
```diff
@@ -539,6 +541,41 @@ Described in {{sec-common-validity}}.
...
+A CoRIM role entry that contains the `manifest-signer` role MUST be added to `corim-entity-map`.
+
+It is out of scope of this document to specify a method of delegating the signer role in the case that an unsigned CoRIM is conveyed through multiple secured links with different notions of authenticity without end-to-end integrity protection.
+
+### CoRIM bundles
+
+A method of signing a bundle of CoRIMs together is through a signed RATS Conceptual Message Wrapper (CMW) {{-cmw}}.
+The COSE_Sign1 signature format can be used with a Collection CMW.
+The COSE protected header can include a CMW collection type name.
+The collection type name SHALL be of the form: `tag:{{&SELF}}:bundle`.
```
yogeshbdeshpande, line 558: CMW contains unsigned-corim-map

yogeshbdeshpande, line 558: We should use the terminology of CMW Collection Profile

yogeshbdeshpande (commented):
Need more text on the CDDL types

draft-ietf-rats-corim.md
```diff
@@ -550,31 +550,33 @@ A CoRIM role entry that contains the `manifest-signer` role MUST be added to `co
...
 
-### CoRIM bundles
+### CoRIM collections
 
-A method of signing a bundle of CoRIMs together is through a signed RATS Conceptual Message Wrapper (CMW) {{-cmw}}.
-The COSE_Sign1 signature format can be used with a Collection CMW.
-The COSE protected header can include a Collection CMW type name.
-The collection type name SHALL be of the form: `tag:{{&SELF}}:bundle`.
+Several CoRIMs may share the same signer and use locally-resolvable references to each other by using a signed RATS Conceptual Message Wrapper (CMW) {{-cmw}}.
+The CMW collection type is similar to a profile in its way of restricting the shape of the CMW collection.
```
nedmsmith, line 556:
```suggestion
The Collection CMW type (i.e., "__cmwc_t") is similar to a profile in its way of restricting the shape of the collection.
```

draft-ietf-rats-corim.md
```diff
@@ -550,31 +550,33 @@ A CoRIM role entry that contains the `manifest-signer` role MUST be added to `co
...
 It is out of scope of this document to specify a method of delegating the signer role in the case that an unsigned CoRIM is conveyed through multiple secured links with different notions of authenticity without end-to-end integrity protection.
 
-### CoRIM bundles
+### CoRIM collections
 
-A method of signing a bundle of CoRIMs together is through a signed RATS Conceptual Message Wrapper (CMW) {{-cmw}}.
-The COSE_Sign1 signature format can be used with a Collection CMW.
-The COSE protected header can include a Collection CMW type name.
-The collection type name SHALL be of the form: `tag:{{&SELF}}:bundle`.
+Several CoRIMs may share the same signer and use locally-resolvable references to each other by using a signed RATS Conceptual Message Wrapper (CMW) {{-cmw}}.
```
nedmsmith, line 555: Can a Collection CMW be an unsigned payload within a secure session or by some other enveloping technology such as PKIX? If yes, then this language may be too constraining. Does "signed RATS CMW" means COSE_Sign1 or could it be interpreted more broadly? For example, it could read: "Several CoRIMs may share the same signer (e.g., as a payload to a secure session protocol or signature format such as COSE) and use locally-resolvable references to each other by using a RATS Conceptual Message Wrapper (CMW) {{-cmw}}." Note reference to COSE is exemplary.

draft-ietf-rats-corim.md
```diff
@@ -550,31 +550,33 @@ A CoRIM role entry that contains the `manifest-signer` role MUST be added to `co
...
+The CMW collection type is similar to a profile in its way of restricting the shape of the CMW collection.
+The collection type for a CoRIM a collection SHALL be `tag:{{&SELF}}:corim`.
```
nedmsmith, line 557:
```suggestion
The collection type for a CoRIM collection SHALL be `tag:{{&SELF}}:corim`.
```

draft-ietf-rats-corim.md
```diff
@@ -539,6 +541,43 @@ Described in {{sec-common-validity}}.
 {::include cddl/unprotected-corim-header-map.cddl}
 ~~~
 
+## Signer authority of securely conveyed unsigned CoRIM {#sec-conveyed-signer}
+
+An unsigned (#6.501-tagged) CoRIM may be a payload in an enveloping signed document.
+An unsigned CoRIM may be a payload within a secure channel.
```
yogeshbdeshpande, line 547:
```suggestion
```

draft-ietf-rats-corim.md
deeglaze: I don't think so. The CoRIM that has dependent-rims is more clearly the starting point. If you're handed just a collection, you don't know which is the starting point.

draft-ietf-rats-corim.md
deeglaze: Suggest a change please.

deeglaze: /cc @thomas-fossati

nedmsmith (approved):
LGTM

draft-ietf-rats-corim.md
```diff
@@ -100,9 +100,11 @@ normative:
 
 informative:
   RFC7942:
+  RFC9277:
```
thomas-fossati, line 103: is this ref used?

draft-ietf-rats-corim.md
```diff
@@ -539,6 +541,44 @@ Described in {{sec-common-validity}}.
 {::include cddl/unprotected-corim-header-map.cddl}
 ~~~
 
+## Signer authority of securely conveyed unsigned CoRIM {#sec-conveyed-signer}
+
+An unsigned (#6.501-tagged) CoRIM may be a payload in an enveloping signed document.
+The CoRIM signer authority is taken from the authenticated credential of the same entity that originates the CoRIM.
```
thomas-fossati, line 547:
```suggestion
The CoRIM signer authority is taken from the authenticated credential of the entity that originates the CoRIM.
```

draft-ietf-rats-corim.md
```diff
@@ -539,6 +541,44 @@ Described in {{sec-common-validity}}.
...
+A CoRIM role entry that contains the `manifest-signer` role MUST be added to `corim-entity-map`.
+
+It is out of scope of this document to specify a method of delegating the signer role in the case that an unsigned CoRIM is conveyed through multiple secured links with different notions of authenticity without end-to-end integrity protection.
+
+### CoRIM collections
+
+Several CoRIMs may share the same signer (e.g., as payload in a different COSE-signed message) and use locally-resolvable references to each other by using a RATS Conceptual Message Wrapper (CMW) {{-cmw}}.
+The CMW collection type is similar to a profile in its way of restricting the shape of the CMW collection.
```
thomas-fossati, line 555:
```suggestion
The Collection CMW type is similar to a profile in its way of restricting the shape of the CMW collection.
```

draft-ietf-rats-corim.md
```diff
@@ -539,6 +541,44 @@ Described in {{sec-common-validity}}.
...
+The collection type for a CoRIM collection SHALL be `tag:{{&SELF}}:corim`.
```
thomas-fossati, line 556:
```suggestion
The Collection CMW type for a CoRIM collection SHALL be `tag:{{&SELF}}:corim`.
```

draft-ietf-rats-corim.md
```diff
@@ -539,6 +541,44 @@ Described in {{sec-common-validity}}.
...
+
+If using other signing envelope formats, the CoRIM signing authority MUST be specified, e.g., by adding the `manifest-signer` role to every CoRIM, or by using a protected header analogous to `corim-meta`.
+
+~~~ cddl
+{::include cddl/cmw-corim-collection.cddl}
+~~~
+
+The Collection CMW MAY use any label for its CoRIMs.
+If there is a hierarchical structure to the CoRIM Collection CMW, the base entry point SHOULD be labeled `0` in CBOR or `"base"` in JSON.
+It is RECOMMENDED to label a CoRIM with its tag-id in string format, where `uuid-type` string format is specified by [RFC4122].
```
thomas-fossati, line 571: RFC4122 is obsoleted by https://www.rfc-editor.org/rfc/rfc9562

```suggestion
It is RECOMMENDED to label a CoRIM with its tag-id in string format, where `uuid-type` string format is specified by {{RFC9562}}.
```

draft-ietf-rats-corim.md
```diff
@@ -539,6 +541,44 @@ Described in {{sec-common-validity}}.
...
+
+An unsigned (#6.501-tagged) CoRIM may be a payload in an enveloping signed document.
+The CoRIM signer authority is taken from the authenticated credential of the same entity that originates the CoRIM.
+A CoRIM role entry that contains the `manifest-signer` role MUST be added to `corim-entity-map`.
+
+It is out of scope of this document to specify a method of delegating the signer role in the case that an unsigned CoRIM is conveyed through multiple secured links with different notions of authenticity without end-to-end integrity protection.
+
+### CoRIM collections
+
+Several CoRIMs may share the same signer (e.g., as payload in a different COSE-signed message) and use locally-resolvable references to each other by using a RATS Conceptual Message Wrapper (CMW) {{-cmw}}.
```
yogeshbdeshpande, line 554:
```suggestion
Several CoRIMs may share the same signer (e.g., as collection payload in a different signed message) and use locally-resolvable references to each other by using a RATS Conceptual Message Wrapper (CMW) {{-cmw}}.
```

yogeshbdeshpande (approved):
LGTM! Lets Merge this!

yogeshbdeshpande:
@deeglaze Lets merge this PR, needs fixing compilation issues, due to referencing. If you can please fix, it, today, we can merge this today/Monday morning UK. Thanks for the great work!

draft-ietf-rats-corim.md
deeglaze: Not since I changed to cbor-record. Will remove.

Issue #388, opened 2025-02-19
Normative language about binary encoding with informative data structures


deeglaze:
Given the appraisal process as currently specified has multiple requirements for comparing CBOR binary encodings, I'm worried we will have a hard time transitioning people to the information model when subtle encoding differences are expected to change verifier behavior.

In order to follow up application/rim+{cbor,cose} with application/rim+{json,jws} (JoRIM) for JSON-based infrastructure, we need to do something about this. I was told, "yeah I ain't don' all that" when suggesting COSE-signed CBOR documents to neighboring teams that will need to produce externally consumable reference values.

The wording for the encoding is currently
> Before performing the binary comparison, a Verifier SHOULD convert both environment-map fields into a form which meets CBOR Core Deterministic Encoding Requirements [STD94].

Then we have informative language, not normative since ECT is informative,
> If any field which is present in the condition ECT environment-map is not binary identical to the corresponding ACS entry field, then the environments do not match.

> When comparing two $crypto-key-type-choice fields for equality, a Verifier SHALL treat them as equal if their deterministic CBOR encoding is binary equal.

> First, a Verifier SHALL locate the entries in the ACS entry element-list which have a matching element-id as the condition ECT element-map

>The comparison MUST return false if the CBOR encoding of the digests entry in the condition ECT or the ACS value with the same codepoint is incorrect (for example if fields are missing or the wrong type).

My recommendation is that since the encoding canonicalization is SHOULD, and comparisons about binary equality of internal representations are SHALL, we need to specify the abstract semantics normatively, not the informative concrete semantics.

Given the JC feature multiplexing that comes with providing a JSON or CBOR encoding for data in CDDL, we can't really say two values are the same only if they are the same element in the grammar's [free algebra](https://en.wikipedia.org/wiki/Free_algebra) (sorry), since JSON and CBOR are still syntactically distinct.
I think it might be fair to say that the the translation of values that are used for "exact matching" into an internal representation should be "equivalence preserving" given an input equivalence class of input values with equal canonical encodings.

We'd then need to take care to define the equivalence between JSON and CBOR representations :-/

andrew-draper:
> we can't really say two values are the same only if they are the same element 

The text is carefully written, and does not include the word "only"

It says that they are equal if their deterministic CBOR encodings are the same, which is another way of saying that they are equal if converting each CBOR object into a tree of constituent unencoded parts returns the same tree in both cases.

Who is specifying JoRIM?

[...]

Issue #375, opened 2025-02-12
BER key type


nedmsmith:
When Evidence is included in X.509 certs or when an X.509 cert is the signer of Evidence (other CM), the signer key is in BER. Given the way to map an external representation to internal by copying, there needs to be a BER form of $crypto-key-type-choice. 

Alternatively, there needs to be a convention for converting keys to an internal representation. But since the CoRIM internal representation is optional. It isn't pragmatic to rely on the internal representation convention for authoring conditions. 

It may be reasonable for implementations to support key mapping routines and make a best effort attempt at matching based on  this. Does it make sense to make it mandatory for implementations to do best-effort key conversion? 

Alternatively, does it make sense for CoRIM to standardize an internal representation for keys independent of the current (optional) internal representation convention?

deeglaze:
First of all, I'm glad you've come back around to more flexible key comparison. You had previously not allowed for `$crypto-key-type-choice` to "match" anything but binary-equal values, despite there not being a specified way to go from some sense of "CoRIM signer" to the ECT's signer representation for comparison to then be fixed.

I think mandatory + best effort = optional however. We should be explicit for the MTI and OTI key comparison algorithms. We should have MTI for EC2, EdDSA, and RSA with the  
I can be okay with this as long as we're saying keys are comparable purely by their key type and their key type parameters, no other subjectpublickey metadata. We have to be explicit about what those are for a baseline set of algorithms however.

We can use the examples that ("EC2",x,y,crv) ("OKP",x,y,crv), ("RSA",n,e), ("HSS-LMS",pub) are pretty much all you need for common PKI. 

MTI EC2 and EdDSA Curve comparisons
```
P-256		1	= 1.2.840.10045.3.1.7
P-384		2	= 1.3.132.0.34
P-521		3	= 1.3.132.0.35
X25519		4	= 1.3.101.110
X448		5	= 1.3.101.111
Ed25519		6	= 1.3.101.112
Ed448		7	= 1.3.101.113
secp256k1	8	= 1.3.132.0.10
```

OTI curve comparisons include non-fixed curves where you get all 7 parameters, or fixed curves not listed above, e.g.,
```
brainpoolP256r1 256 = 1.3.36.3.3.2.8.1.1.7 [RFC5639]
brainpoolP320r1 257 = 1.3.36.3.3.2.8.1.1.9 [RFC5639]
[...]

PR #393, opened 2025-02-21
Encoding equality to equivalence, and key equivalence

deeglaze:
Issues #375, #388.


cddl/crypto-key-type-choice.cddl
nedmsmith, line 1: I don't see how the expanded typing makes a difference wrt lossy/non-lossy properties of each type. We could have just documented them as one or the other.

yogeshbdeshpande: Agree with Ned, we do not need to change the types here, we can just state the properties, without been explicit on the type. May be we are missing something, @deeglaze here?

deeglaze: I can change them back, sure

draft-ietf-rats-corim.md
```diff
@@ -2518,7 +2528,7 @@ A Verifier SHALL treat two algorithm identifiers as equal if they have the same
 If both an integer and a string representation are defined for an algorithm then entities creating ECTs SHOULD use the integer representation.
 If condition ECT and ACS entry use different names for the same algorithm, and the Verifier does not recognize that they are the same, then a downgrade attack is possible.
 
-The comparison MUST return false if the CBOR encoding of the `digests` entry in the condition ECT or the ACS value with the same codepoint is incorrect (for example if fields are missing or the wrong type).
+The comparison MUST return false if the determisitic encoding of the `digests` entry in the condition ECT or the ACS value with the same codepoint is incorrect (for example if fields are missing or the wrong type).
```
steven-bellock, line 2531:
```suggestion
The comparison MUST return false if the deterministic encoding of the `digests` entry in the condition ECT or the ACS value with the same codepoint is incorrect (for example if fields are missing or the wrong type).
```

steven-bellock (commented):
If lossy / lossless comparison becomes a thing then there should be an informative section that describes the concepts and provides an example or two.

nedmsmith (approved):
LGTM

draft-ietf-rats-corim.md
```diff
@@ -2466,11 +2464,11 @@ If any of the fields does not match, then the condition ECT does not match the A
 The Verifier SHALL compare each field which is present in the condition ECT `environment-map` against the corresponding field in the ACS entry `environment-map` using binary comparison.
 Before performing the binary comparison, the Verifier SHOULD convert both `environment-map` fields into a form which meets CBOR Core Deterministic Encoding Requirements {{-cbor}}.
 
-If all fields which are present in the condition ECT `environment-map` are present in the ACS entry and are binary identical, then the environments match.
+If all fields which are present in the condition ECT `environment-map` are present in the ACS entry and are equivalent, then the environments match.
```
yogeshbdeshpande, line 2467: Should we not qualify what equivalent means?

yogeshbdeshpande (approved):
LGTM, just check a minor point...

draft-ietf-rats-corim.md
deeglaze: I think that we can revert the equivalence text for ECTs since there's already a translation to them.

draft-ietf-rats-corim.md
```diff
@@ -2285,7 +2283,7 @@ See {{sec-authority}}.
 
 If the merged `measurement-values-map` contains duplicate codepoints and the
 measurement values are equivalent, then duplicate claims SHOULD be omitted.
-Equivalence typically means values MUST be binary identical.
+Equivalence means the values MUST be binary identical unless they are from a profile extension that defines its own equivalence relation.
```
yogeshbdeshpande, line 2286: Move this to a top level Equivalence definition, and check, whether it is applicable in all sections!

PR #395, opened 2025-02-23
Add EAT measured component to `measurement-values-map`

### LINKED ISSUE #381: EAT Measured Components as measurement values
[EAT Measured Components (MC)](https://datatracker.ietf.org/doc/draft-ietf-rats-eat-measured-component/) are an extension to the EAT type system that can be used to model "measurable" objects of an attester's target environment, i.e., objects whose state can be sampled and digested.

Examples of such _measured components_ include the invariant part of firmware loaded in memory at startup time, a run-time integrity check (RTIC), a file system object, or a CPU register.

MCs are defined as extensions of the EAT [`Measurement`](https://www.ietf.org/archive/id/draft-ietf-rats-eat-31.html#section-4.2.16) claim, alongside "payload" CoSWIDs.
Unlike CoSWID, these do not require an anchoring file system, making them more suitable for early boot components.

Arm [PSA and CCA software components](https://www.ietf.org/archive/id/draft-tschofenig-rats-psa-token-24.html#section-4.4.1) can be straightforwardly mapped to MCs.
Future attester technologies utilizing EAT will be able to use MCs seamlessly.

MCs reuse some parts of the CoRIM type system (namely, [`digest`](https://www.ietf.org/archive/id/draft-ietf-rats-corim-06.html#section-7.7)) and have CBOR and JSON serialisations.

This issue tracks their addition to the [`measurement-values`](https://www.ietf.org/archive/id/draft-ietf-rats-corim-06.html#section-5.1.4.1.4.2) map.

thomas-fossati:
Fix #381

### DIFF
--- a/cddl/measurement-values-map.cddl
+++ b/cddl/measurement-values-map.cddl
@@ -15,6 +15,6 @@ measurement-values-map = non-empty<{
   ? &(name: 11) => text
   ? &(cryptokeys: 13) => [ + $crypto-key-type-choice ]
   ? &(integrity-registers: 14) => integrity-registers
-  ? &(measured-component: 15) => measured-component
+  ? &(measured-component: 16) => measured-component
   * $$measurement-values-map-extension
 }>
--- a/cddl/examples/comid-measured-component.diag
+++ b/cddl/examples/comid-measured-component.diag
@@ -23,7 +23,7 @@
       [
         / measurement-map / {
           / comid.mval / 1: {
-            / comid.measured-component / 15: {
+            / comid.measured-component / 16: {
               / id / 1: [
                 / name / "boot loader X",
                 / version / { 0: "1.2.3rc2" }


cddl/measurement-values-map.cddl
```diff
@@ -15,5 +15,6 @@ measurement-values-map = non-empty<{
   ? &(name: 11) => text
   ? &(cryptokeys: 13) => [ + $crypto-key-type-choice ]
   ? &(integrity-registers: 14) => integrity-registers
+  ? &(measured-component: 15) => measured-component
```
deeglaze, line 18: 15 conflicts with the linear privlevel measurement PR.

thomas-fossati: Oops, apologies.  Will fix.

deeglaze (approved):

yogeshbdeshpande (approved):
LGTM!

nedmsmith:
I find the compositional aspects of "measured components" confusing. Given the mkey is already a compositional element with is sub-component of an environment, what does it mean to have a sub-sub-element called a "component"? It seems the basic idea of using triples to enumerate various components or sub-components is being side-stepped toward composition expressed in terms of nesting. 

Most of the values in a measured component already exist in the measurement-values-map like digest, version, flags / raw-value. Mkey can support text identifiers which seems to map to `component-id`. 

It seems like a straight forward mapping without defining another layer of composition inside of measurement-values-map.

draft-ietf-rats-corim.md
```diff
@@ -970,6 +971,8 @@ The following describes each member of the `measurement-values-map`.
 
 * `integrity-registers` (index 14): A group of one or more named measurements associated with the environment.  Described in {{sec-comid-integrity-registers}}.
 
+* `measured-component` (index 15): A measured component defined in {{Section 4.2 of -rats-eat-mc}}.
```
deeglaze, line 974: What is the matching semantics for this component? If I want to write a conditional endorsement over a piece of your measured component, what are those rules?

If you don't want to have to compose with conditional endorsements, I would recommend just making eat-measured-component triples their own thing.
If you do want composition, I think you'll need to do some transformation so component-id is split into the name and version measurements, measurement promoted to digest measurement value, flags promoted to raw-value, and signers promoted to a new measurement-value-map codepoint that has defined matching semantics.

I can appreciate not wanting to have to transform it for the endorsement. I do think you still need to define the matching semantics, and this may even be something you could specify as a request to the IANA to include the codepoint in the new measurement-values-map registry form the rats-eat-mc RFC.
Adding more matching rules to the base document for what's mostly a rejiggering of existing measurements seems a little wasteful of reviewer attention, but you have more experience with RFCs than I do.

deeglaze (changes requested):

draft-ietf-rats-corim.md
thomas-fossati: > What is the matching semantics for this component?

Exact match.

deeglaze: Even for signers? Seems rough, but okay.

thomas-fossati: Brutal but honest ;-)

andrew-draper (changes requested):
I don't think we should approve this PR in its current form.
I agree with comments above that there are no matching rules, writing those matching rules will I think expose some gaps in the PR which need to be fixed.
The gap I am most concerned about is that the `measured-component` object contains both a name and a value, but this object has been added to `mval` which is  for values only.

Issue #359, opened 2025-01-17
Update examples for raw-value


andrew-draper:
Need to add issues for PR https://github.com/ietf-rats-wg/draft-ietf-rats-corim/pull/294 (adding `raw-value-compare`)

yogeshbdeshpande:
@andrew-draper Request if we can progress on this , for the coming week, this will help greatly bringing this down to closure!

yogeshbdeshpande:
@andrew-draper : please try to close this, in coming week

yogeshbdeshpande:
@andrew-draper : This issue still needs some work, is this correct understanding?

yogeshbdeshpande:
PR 71 and PR #397 completes this issue!

PR #397, opened 2025-02-23
Create example file for all types of raw value

andrew-draper:
This fixes issue #359

### DIFF
--- a/cddl/examples/comid-raw-value.diag
+++ b/cddl/examples/comid-raw-value.diag
@@ -32,7 +32,7 @@
             }
           }
       ]
-    ] [
+    ], [
       / environment-map / {
         / comid.class / 0 : {
           / comid.class-id / 0 :
@@ -52,7 +52,7 @@
             }
           }
       ]
-    ] [
+    ], [
       / environment-map / {
         / comid.class / 0 : {
           / comid.class-id / 0 :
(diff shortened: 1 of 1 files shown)


deeglaze (approved):

cddl/examples/comid-raw-value.diag
```diff
@@ -0,0 +1,78 @@
...
+              h'67b28b6c34cc40a19117ab5b05911e37'
+            ),
+          / comid.vendor / 1 : "ACME Inc.",
+          / comid.model / 2 : "ACME RoadRunner",
+          / comid.layer / 3 : 1
+        }
+      },
+      [
+        / measurement-map / {
+          / This version compares the first 16 bits using deprecated syntax /
```
yogeshbdeshpande, line 69: I think, if we are deprecating a use of a type, we should not exemplify it?

yogeshbdeshpande (commented):
Please review my comment

cddl/examples/comid-raw-value.diag
andrew-draper: I think we should make sure it keeps working until it is fully deleted. We can delete the example when we delete the usage.

yogeshbdeshpande: I was advocating best usage via examples, of how to use it correctly! Anyway lets discuss this during the call!

nedmsmith (approved):
LGTM

PR #411, opened 2025-03-19
Add relevance of JSON for CoRIM

### LINKED ISSUE #412: Add JSON Clarifications to CoRIM document
> Recapping:
_Need to clearly highlight that JSON is also supported as encoding format for Triples and one should provide Examples as to how JSON Encoding will look like? (Focus on CBOR only to avoid scope creep and move more quickly to WGLC)_
> 
> 
> 
> Maybe the next step is to create separate issues for each and someone volunteers to do the work? 

 _Originally posted by @nedmsmith in [#195](https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/195#issuecomment-2099059870)_

yogeshbdeshpande:
Fixes #412


nedmsmith:
Do we want to include something that is defined by a profile of CoRIM in the base spec? Why not let the profile make the JSON compatibility statement?

nedmsmith:
What issue is this PR addressing so there is motivation and context discussion?

yogeshbdeshpande:
Yes, this should be the part of base spec!

This is in response to the `JSON Compatibility` Feedback we received from last IETF

nedmsmith:
> Yes, this should be the part of base spec!
> 
> This is in response to the `JSON Compatibility` Feedback we received from last IETF

Can we create an issue that captures that feedback and discusses options for how to address it?

yogeshbdeshpande:
> > Yes, this should be the part of base spec!
> > This is in response to the `JSON Compatibility` Feedback we received from last IETF
> 
> Can we create an issue that captures that feedback and discusses options for how to address it?

Done: https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/412

yogeshbdeshpande:
Will not address his PR!

Issue #415, opened 2025-03-28
Handling of conditional endorsement when there are multiple identical devices


andrew-draper:
Consider an compound system using two attestation environments and two target environments.

In the example, the AEs and TEs use identical hardware, but may be running different firmware.

This diagram shows a simplified view of the system.
```mermaid
graph TD
    Verifier["Verifier"]

    subgraph Compound["Compound system"]

    subgraph DICE1["Component #1"]
        TE1z["SPDM TE"]
        TE1y["SPDM AE"]
    end
    TE1y ==> TE1z

    subgraph DICE2["Component #2"]
        TE2y["SPDM AE"]
        TE2z["SPDM TE"]
    end
    TE2y ==> TE2z

    end

    style Compound fill:#CCC,stroke:#CCC

    Verifier -...-> TE1y
    Verifier -...-> TE2y
```
The Verifier communicates directly with an attestation environment within each component.

The attestation environments return measurements of each TE, for example AE#1 might return a hash indicating the firmware running in TE#1:
```cbor-diag
  [
    / environment-map / {
      / class / 0 : {
        / class-id / 0 : / 2.16.1.2.3.4 / 111(h'6001020304'),
        / vendor / 1 : "example.com"
      },
      / instance / 1 : 560(h'12345678')
    },
    / measurement-map / [ {
      / mval / 1 : {
        / digests / 2 : [
          [
            / sha-384 / 7,
            h'15E77D6F133252F1DB7044901313884F2977D2109B33C79F33E079BFC78865255C0FB733C240FDDA544B8215D7B8F815'
          ]
        ]
      }
    } ]
  ]
```

The measurements of TE#2 are similar, except for the instance ID and the digest in the measurement-values-map.

The evidence import phase will add the measurements for TE#1 and TE#2 to the ACS.

The CoRIM author has published several CoRIMs, each of which maps one hash to a version number, for example:

```comid
  / triples / 4 : {
    / conditional-endorsement-triples / 10 : [
      / conditions: / [ 
        [
          / environment-map / {
            / class / 0 : {
              / class-id / 0 : / 2.16.1.2.3.4 / 111(h'6001020304'),
              / vendor / 1 : "example.com"
            }
          },
          / measurement-map / [ {
            / mval / 1 : {
              / digests / 2 : [
                [
                  / sha-384 / 7,
[...]

PR #416, opened 2025-04-02
Domain edits

yogeshbdeshpande:


cddl/intrep-ect.cddl
```diff
@@ -4,6 +4,7 @@ ECT = {
   ? authority: [ + $crypto-key-type-choice ]
   cmtype: cm-type
   ? profile: $profile-type-choice
+  ? domain-id: environment-map
```
nedmsmith, line 7: Shouldn't this be a list of environment-map since an ECT could represent a domain that has as its members other domains or environments? An expectation would be that the domain-id list would be populated with environment-map values that are ECTs in the ACS. If not, that might indicate a consistency problem.

nedmsmith (commented):
See inline comment

cddl/intrep-ect.cddl
yogeshbdeshpande: No, in this model, the Domain ECT carries a new cm-type known as domain-member. Note I have only modelled domain membership triples, for now! When the Env-list array of (Env) specified in the Domain Membership Triple internal rep cm-type: `domain-member` matches the ECT list for cm-type Evidence,  then the domain-id is moved to the ACS for each of the ECT list for Domain-Env. the domain-id is the Identifier for the domain for which this ECT is a member!

draft-ietf-rats-corim.md
```diff
@@ -2219,6 +2254,32 @@ The following transformation steps are applied for both the `identity-triples` a
...
+
+{: dmt2-enum}
+* DMT.`identity`, $domain-identity-type-choice `environment-map` as under:
+
+> > **copy**(`environment-map`, `dm`.`identity`.`environment-map`)
+
+{: dmt2-enum}
+* For each e in DMT.`members`:
+
+> > **copy**(e.`environment-map`, `dm`.`identity`.`environment-map`)
```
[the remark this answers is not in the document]
yogeshbdeshpande: Well spotted! I agree with the comment!

cddl/intrep-ect.cddl
nedmsmith: > the domain-id is the Identifier for the domain for which this ECT is a member

Can a member be a member of multiple domains? The expressiveness of the membership triple allows this.

yogeshbdeshpande: Good question, I am thinking, while I write. !

yogeshbdeshpande: Yes, the statement in general is true, (i.e. an ECT Member can be a member of multiple domains). However if I look this from the prism of ACS which is an Appraisal Context for a Specific Attester, in that Context Data Structure, i.e. ACS, the ECT for Domain members, can ONLY have that specific Domain -Identity which is relevant for that Appraisal Context. So it needs only one domain-identity, unless I am missing something!

draft-ietf-rats-corim.md
```diff
@@ -2219,6 +2254,32 @@ The following transformation steps are applied for both the `identity-triples` a
...
+> > **copy**(`environment-map`, `dm`.`identity`.`environment-map`)
+
+{: dmt2-enum}
+* For each e in DMT.`members`:
+
+> > **copy**(e.`environment-map`, `dm`.`members`.`environment-map`)
+
+
+* If the Endorsement conceptual message has a profile, the profile identifier is copied to the `dm`.`members`.`profile` field.
+
```
nedmsmith, line 2282: Should there be a section in "ACS augmentation using CoMID triples" that describes how to process the dm structure so that it results in changes or additions to the ACS entries? For example, for all the members in dm.members, the ACS is searched to find environment-maps that match and the ACS.EMT.domain field is updated to include dm.members.environment-map. If the search returns NULL then what? Should a new EMT be added to the ACS? 

What happens if there are multiple matches? Do all of them get their EMT.domain fields updated? 

What happens if there is a partial match, for example only class-map, instance-id, or group-id (or combinations of the three) match?

yogeshbdeshpande: If the match fails, for a DMT, it is not added to ACS.

If none of the DMT succeeds in a `complete match` and a Verifier Policy requires DMT then it SHOULD fail Verification!

For partial match also, the result is a Failure!

I was planning to add this, description, once we (CoRIM team) is aligned in general to this concept!

yogeshbdeshpande: I am assuming that a `profile` will populate `DMT: Environment` structure identical to Evidence ECT's Environment structure.

cddl/intrep-ect.cddl
nedmsmith: Load balancing, redundancy/resiliency are use cases to consider. CXL and other bus architectues support multi-host (multiple bus master) operation where a CXL device can respond to multiple hosts. It likely wouldn't scale if attestation had to be reported for every CXL call. But during operation both hosts will be used. If a CXL device's domain membership is restricted to a single host, which does it pick? Alternatively, a top-down approach would reveal that a VM / TDX TD might allow CXL multi-host mastering and include environment-maps for each in a domain membership triple. If a CXL device also has a domain membership triple that identifies only one, then does that mean the top down statement that identifies both is wrong or that the bottom up statement is wrong? 
There is a difference between Attestation Env "collection" of measurements and declaring domain membership to a bus master or a lead Attester (BMC host) that is assembing Evidence for comms over a network vs. actually "collecting" measurements. There is a natural end to "collection" which is the top of a DICE layering scheme or the end of a PCR sequence. In the case of SGX/TDX the Quoting E/TD can describe the full layering in Evidence, but only on a per-tenant basis and only if the tenant requests it. Nevertheless, the tenant can have access to a multi-host CXL bus where each host will collect Evidence from the same device. It isn't clear if the device should attempt to report that it knows it's host(s) or if only the host(s) should describe the reachable device(s).
The description also should consider whether a class-based reference domain can match an evidence expression that includes instance-based environment (see https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/415)

draft-ietf-rats-corim.md
```diff
@@ -2351,6 +2430,21 @@ For each `rv` entry, the `condition` ECT is compared with an ACS ECT, where the
...
+Domain Membership Triples allow an Endorser (for example an Integrator) to issue an authoritative statement about the composition of an Attester as a collection of Environments.
+If the Verifier Appraisal policy requires Domain Membership, then membership triple sets a reference composition of an Attester in a Verifier database.
+At the time of Verification, the reference is then matched with actual composition as reported by an Attester in the Evidence.
+
+Domain Membership Triples are first transformed into an internal representation following the steps mentioned in {{sec-dm-trans}} leading to a representation as as specified in {{sec-dm-membership}}.
+
+For each `ECT` entry in the list of `members`, the environment elements (example Class ID) is compared with the equivalent environment elements in ACS Entry with `cmtype` as `evidence`.
+The process continues, till every environment in the member list matches the entry in ACS.
+If all the members match then the `environment` from the domain-id field, is copied to `domain-id` field for each of the member ECT.
+All the member entries are then added in the ACS, with `authority` specified in the  `domain-id` field.
```
nedmsmith, line 2444: I didn't expect the membership triple would result in new entries into the ACS. Rather, the existing entries would be updated to populate the empty `domain-id` field.

draft-ietf-rats-corim.md
```diff
@@ -2351,6 +2430,21 @@ For each `rv` entry, the `condition` ECT is compared with an ACS ECT, where the
...
+
+Domain Membership Triples allow an Endorser (for example an Integrator) to issue an authoritative statement about the composition of an Attester as a collection of Environments.
+If the Verifier Appraisal policy requires Domain Membership, then membership triple sets a reference composition of an Attester in a Verifier database.
+At the time of Verification, the reference is then matched with actual composition as reported by an Attester in the Evidence.
+
+Domain Membership Triples are first transformed into an internal representation following the steps mentioned in {{sec-dm-trans}} leading to a representation as as specified in {{sec-dm-membership}}.
+
+For each `ECT` entry in the list of `members`, the environment elements (example Class ID) is compared with the equivalent environment elements in ACS Entry with `cmtype` as `evidence`.
+The process continues, till every environment in the member list matches the entry in ACS.
+If all the members match then the `environment` from the domain-id field, is copied to `domain-id` field for each of the member ECT.
```
nedmsmith, line 2443: The domain-id is a type-choice that includes `environment-map`.  `environment` refers to a field in an ECT. It isn't clear what is intended. It appears the `dm.members` list ECTs are used as conditions to search the ACS. 
Why do all of the ACS entries have to be matched before any of them can be matched? If a member entry is matched (corroborated) then it isn't clear if entry into the ACS is the way to represent that. But one goal of membership triple is to represent the absence of expected members. This can be represented by adding the expected member ECTs using the authority of the dm triples in corim. It is unclear how the ACS should represent corroborated members. 

Additionally, the domain identity ECT also needs to be added to the ACS with membership triple signer authority.  

Note: Since ACS evidence entries could have populated the domain-id using a membership triple in evidence, the membership triple in corim that matches should verify that the domain-id in evidence ECT is the same as domain-id in dm ECT. If they disagree then actual state doesn't match expected state. However, it is unclear how the ACS represents corroborated members.

nedmsmith (commented):
see multiple comments

nedmsmith:
Not having the description of evidence first means we may not be on the same page wrt the state of the ACS when a DM triple is processed. I expect that the `domain-id` in evidence ECT will be populated by the Attesting Env that reported Evidence for the Target Env (aka the `member`). For example, a DICE layering would be processed such that each layer represents a domain and a domain member where layer n+1 is the domain that contains layer n as a member.  The domain-member triple in evidence isn't necessarily needed IMHO.

The dm triple in corim wants to corroborate what is in evidence, but also assert to the ACS when an expected member is missing. Corroboration will add a duplicate ECT as was found in Evidence ECTs but mark it as a RV ECT and authority will be that of the RVP. Applying similar semantics, the dm processing will add a duplicate ECT if found in Evidence but assert it as a DM cmtype with authority of the DM signer. However, if there are additional expected members in the dm, but not in evidence. Adding entries for them in the ACS will falsely appear as if the evidence contained them. An additional check is needed that pairs up the evidence ECTs with the DM ECTs in order to detect unpaired ECTs. The int-rep structure doesn't appear to have a way to represent the pairing result. For example, a real-world implementation might use a pointer that references paired / corroborated records while null pointers represent uncorroborated ECTs. It's unclear how far down an implementation path the draft should go to represent ACS state. Possibly, it is good enough to document the need to do the pairing check and leave it at that? 

Nevertheless, there is a concern that for other conditional endorsements, the potential for endorsing something that has a twin in the ACS might result in only one twin being updated. If the ACS can contain multiple instances of ECTs for the same component, are we sure there is a merge routine that reliably relates/combines them such that the result is an accurate representation of the state of the Attester? This may not be unique to dm triples. So maybe it is a separable issue?

draft-ietf-rats-corim.md
yogeshbdeshpande: The Domain Membership triple adds ACS Entries which has `cmtype` set to `domain-member`

yogeshbdeshpande: This ONLY happens when all the domain members in the Staging area are compared with Evidence ECT and is a perfect match.

Then the new ACS Entries are added...

draft-ietf-rats-corim.md
yogeshbdeshpande: Why do all of the ACS entries have to be matched before any of them can be matched? 

Basically ECT Entries in the Staging areas from Domain Membership all has to match the ACS Entries with cmtype as Evidence, All has to Match otherwise, the Reported Evidence has few members missing, which should FAIL the Appraisal..

nedmsmith (commented):
multiple comments

draft-ietf-rats-corim.md
```diff
@@ -1457,19 +1460,27 @@ trustworthiness properties of the subject domain exists.
...
+{::include cddl/domain-type-choice.cddl}
+~~~
 
 #### Domain Membership Triple {#sec-comid-triple-domain-membership}
 
 A Domain Membership triple assigns domain membership to environments.  The
 subject identifies a domain ({{sec-comid-domain-type}}) that has a predicate
-relationship to the object containing one or more environments.  Endorsed
-environments ({{sec-comid-triple-endval}}) membership is conditional upon
+relationship to the object containing one or more environments or one of more domains.
```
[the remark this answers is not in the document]
yogeshbdeshpande: Agree, we should change this! Thanks for the review!

cddl/intrep-domain-mem.cddl
```diff
@@ -0,0 +1,4 @@
+dm = [ + {
+  identity: ECT
```
yogeshbdeshpande, line 2:
```suggestion
  domain-id: ECT
```

draft-ietf-rats-corim.md
yogeshbdeshpande: Added a separate section, hence resolving the comment now!

draft-ietf-rats-corim.md
yogeshbdeshpande, line 1471:
```suggestion
relationship to the object where the object contains one or more members.
```

draft-ietf-rats-corim.md
```diff
@@ -219,6 +219,9 @@ Class ID:
 : An identifier for an Environment that is shared among similar Environment instances, such as those with the same hardware assembly.
 See also {{Section 4.2.4 of -eat}}.
 
+Domain:
+: A context to represent a collection of Environments that are grouped togther to represent an entity, for example a Composite Device {{Section 3.3 of -rats-arch}}
```
henkbirkholz, line 223:
```suggestion
: A context to represent a collection of Environments that are grouped togther to represent an entity, for example a product that is a Composite Device {{Section 3.3 of -rats-arch}}
```

draft-ietf-rats-corim.md
```diff
@@ -1457,19 +1460,24 @@ trustworthiness properties of the subject domain exists.
...
 
 #### Domain Membership Triple {#sec-comid-triple-domain-membership}
 
-A Domain Membership triple assigns domain membership to environments.  The
-subject identifies a domain ({{sec-comid-domain-type}}) that has a predicate
-relationship to the object containing one or more environments.  Endorsed
-environments ({{sec-comid-triple-endval}}) membership is conditional upon
-successful matching of Reference Values ({{sec-comid-triple-refval}}) to
-Evidence.
+A Domain Membership triple assigns a domain identifier to member environments. The
```
henkbirkholz, line 1469:
```suggestion
A Domain Membership triple associates a domain identifier with corresponding member environments. The
```

draft-ietf-rats-corim.md
```diff
@@ -1267,7 +1270,7 @@ The signed integer range representation is an inclusive range unless either `min
 
 ##### Domain Types {#sec-comid-domain-type}
 
-A domain is a context for bundling a collection of related environments and their measurements.
+A domain is a context for bundling a collection of related environments.
```
henkbirkholz, line 1273:
```suggestion
A Domain is a context for bundling a collection of related environments.
```

draft-ietf-rats-corim.md
```diff
@@ -1457,19 +1460,24 @@ trustworthiness properties of the subject domain exists.
...
-subject identifies a domain ({{sec-comid-domain-type}}) that has a predicate
-relationship to the object containing one or more environments.  Endorsed
-environments ({{sec-comid-triple-endval}}) membership is conditional upon
-successful matching of Reference Values ({{sec-comid-triple-refval}}) to
-Evidence.
+A Domain Membership triple assigns a domain identifier to member environments. The
+triple subject identifies a domain ({{sec-comid-domain-type}}) that has a membership predicate
+relationship to the object where the object contains one or more members.
+When predicates are domains, it allows Endorsers to construct a hierarchical model
+where a domain can be represented using a collection of well defined domain identifiers.
```
henkbirkholz, line 1473:
```suggestion
triple subject identifies a Domain ({{sec-comid-domain-type}}) that has a membership predicate relationship to the object where the object contains one or more members.
When predicates are Domains, it allows Endorsers to construct a hierarchical model
where a Domain can be represented using a collection of well defined domain identifiers.
```

thomas-fossati, line 1473: Not very clear, I suggest the following rewrite:
```suggestion
A Domain Membership triple links a domain identifier to its member Environments.
The triple’s object lists all the Environments in the domain, and the domain is identified by the Environment in the triple’s subject.
This allows the recursive construction of the topology of a Composite Device, where multiple lower-level domains can be aggregated into higher-level domains.
```

thomas-fossati (commented):
partial review

draft-ietf-rats-corim.md
```diff
@@ -1772,6 +1780,8 @@ Environment-Claims Tuples (ECT) have five attributes:
 
 * Profile : The profile that defines this tuple. If no profile is used, this attribute is omitted.
 
+* Domain Identifier : The Identity of the Domain to which a given ECT belongs.
```
thomas-fossati, line 1783:
```suggestion
* Domain Identifier : The Domain to which a given ECT belongs.
```

draft-ietf-rats-corim.md
```diff
@@ -1868,18 +1881,51 @@ If the `selection` criteria is not satisfied, then evaluation procedes to the ne
...
+|           | `domain-id`     | n/a         |
 | addition  | `environment`   | Mandatory   |
 |           | `element-list`  | Mandatory   |
 |           | `authority`     | Mandatory   |
 |           | `cmtype`        | Mandatory   |
 |           | `profile`       | Optional    |
+|           | `domain-id`     | n/a         |
 {: #tbl-ev-ect-optionality title="Endorsed Values and Endorsed Values Series tuples requirements"}
 
+### Internal Representation of Domain Membership {#sec-dm-membership}
```
thomas-fossati, line 1899: In the internal representation “block”, the anchor convention is `sec-ir-...` 

```suggestion
### Internal Representation of Domain Membership {#sec-ir-dm}
```

thomas-fossati (commented):
soem more comments

draft-ietf-rats-corim.md
nedmsmith: Domain shouldn't take the place of Reference Value triples. This description should assume the RV triples processing has completed already. That means there are `reference` ECTs in the ACS when membership triples are evaluated.

nedmsmith (commented):
multiple comments

draft-ietf-rats-corim.md
nedmsmith: > Membership triple adds ACS Entries

It might help to separate the steps that track who said what from enforcement steps that detect a bad conversation. 

Membership triple is an unconditional endorsement meaning the endorser can say what the topology of the device is without being censored / influenced by what the Attester says. There are 3 ways to originate new claims, evidence, conditional endorsement (where the addition environment is not part of the condition), and domain triples. Membership can assert a composition even if the Attester doesn't attest. Example: a laptop vendor VZ asserts that model-Z has an IO controller, keyboard, trackpad, and fingerprint sensor. VZ can produce a rim for model-Z that relates the collection of components. If the Verifier authenticates the Attester using a device credential that identifies it as model-Z, the verifier can find corims for model-Z which say "There is a collection of io-ctlr, kbd, tp, fp-rdr with a domain named DZ". This could be the first set of ECTs in the ACS as the composition isn't conditional on Evidence. 

Lets say the Attester does produce Evidence that has a measurement for io-ctlr, kbd, and tp but not fp-rdr. The domain composition from VZ is still correct. However, the inferred domain composition from Evidence is dissimilar. This difference can be detected using the right combination of queries and comparisons. The enforcement action might be policy driven where the absense of fp-rdr isn't a security issue (maybe fp-rdrs are flakey). Afterall, trying to infer a collection from Evidence is "reading in" more than was intended.

Lets say the Attester (IO-ctlr) wants to be explicit about the domain and includes membership triples in Evidence. Assume the fp-rdr is flakey and fails to return evidence. The IO ctlr membership assertion won't agree with the Evidence set (which lacks fp-rdr). This shows internal inconsistency but doesn't signal an attack since fp-rdr is known to be flakey. We assume an experienced attacker would cause the IO-ctlr to lie about the missing fp-rdr. 

A policy might expect fp-rdr to be missing from Evidence and conclude that IO-ctlr expected fp-rdr, but that's because the fp-rdr is flakey.

Lets say another policy is more sophisticated where if the VZ asserts presence of fp-rdr then override the policy where the IO-ctlr says there should be a fp-rdr but lack of fp-rdr is due to it being flakey. Now it makes sense to flag the missing fp-rdr as a security flaw. 

But in all these cases, there is never a reason to remove entries from the "who said what" record or for the processing of the ACS to fail.

nedmsmith: > Rather, the existing entries would be updated to populate the empty domain-id field.

Note that merging of records in the ACS falls under ACS optimization strategies. The comment above assumed some sort of optimization is being made to minimize the set of ECTs that make claims about any particular environment (component).

nedmsmith: I think it would be OK if the membership processing algorithm simply added the membership ECTs to the ACS unconditionally. 

If we think there is an ACS consistency algorithm that is universal, that considers membership, evidence and possibly reference ECTs, then maybe it makes sense for a later phase (phase 5?) to apply the analysis / optimization.

cddl/intrep-ect.cddl
```diff
@@ -18,4 +19,5 @@ cm-type =  &(
   attestation-results: 3
   verifier: 4
   policy: 5
+  domain-member: 6
```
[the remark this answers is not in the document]
yogeshbdeshpande: Generic Endorsements does not fit this case, as their is a definite meaning to this, identifier which has far reaching implications then a normal Endorsement

draft-ietf-rats-corim.md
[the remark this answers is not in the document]
yogeshbdeshpande: I agree, will change the text to better explain this!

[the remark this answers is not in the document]
yogeshbdeshpande: It may not just be a Composite Device, it could be a Layered Attester also,

draft-ietf-rats-corim.md
yogeshbdeshpande: Taken all the points from here and incorporated a generic term entity with an example of Composite Device!

draft-ietf-rats-corim.md
yogeshbdeshpande: I agree with the fact that ACS Entries for Domain membership can be added un-conditionally. However, Verifier can ONLY start populating the ACS when it receives the Evidence. The ACS cannot be un-conditionally be populated with ANY DM triples, without Evidence setting the context of ACS with Evidence.

draft-ietf-rats-corim.md
```diff
@@ -2351,6 +2425,21 @@ For each `rv` entry, the `condition` ECT is compared with an ACS ECT, where the
...
+At the time of Verification, the reference is then matched with actual composition as reported by an Attester in the Evidence.
+
+Domain Membership Triples are first transformed into an internal representation following the steps mentioned in {{sec-ir-dm-trans}} leading to a representation as specified in {{sec-ir-dm}}.
+
+For each `ECT` entry in the list of `members`, the environment elements (example Class ID) is compared with the equivalent environment elements in ACS Entry with `cmtype` as `evidence`.
+The process continues, untill every environment in the member list matches the entry in ACS.
+If all the members match then the `environment` from the domain-id field, is copied to `domain-id` field for each of the member ECT.
+All the member entries are then added in the ACS, with `authority` specified in the  `domain-id` field.
+
+If the match fails, then the process is repeated using next Domain Membership entry in the Verifier. If none of the entries match, then the Verification has failed.
```
deeglaze, line 2441:
```suggestion
If the match fails, then the process is repeated using next Domain Membership entry in the Verifier.
If none of the entries match, then the Verification has failed.
```

draft-ietf-rats-corim.md
```diff
@@ -2351,6 +2425,21 @@ For each `rv` entry, the `condition` ECT is compared with an ACS ECT, where the
...
+Domain Membership Triples allow an Endorser (for example an Integrator) to issue an authoritative statement about the composition of an Attester as a collection of Environments.
+If the Verifier Appraisal policy requires Domain Membership, then membership triple sets a reference composition of an Attester in a Verifier database.
+At the time of Verification, the reference is then matched with actual composition as reported by an Attester in the Evidence.
+
+Domain Membership Triples are first transformed into an internal representation following the steps mentioned in {{sec-ir-dm-trans}} leading to a representation as specified in {{sec-ir-dm}}.
+
+For each `ECT` entry in the list of `members`, the environment elements (example Class ID) is compared with the equivalent environment elements in ACS Entry with `cmtype` as `evidence`.
+The process continues, untill every environment in the member list matches the entry in ACS.
+If all the members match then the `environment` from the domain-id field, is copied to `domain-id` field for each of the member ECT.
+All the member entries are then added in the ACS, with `authority` specified in the  `domain-id` field.
```
deeglaze, line 2439:
```suggestion
All the member entries are then added in the ACS, with `authority` specified in the `domain-id` field.
```

draft-ietf-rats-corim.md
```diff
@@ -1868,18 +1879,51 @@ If the `selection` criteria is not satisfied, then evaluation procedes to the ne
...
 
+### Internal Representation of Domain Membership {#sec-ir-dm}
+
+An internal representation of Domain Membership uses the `dm` relation, which is a list of ECTs that describes the composition of an Attester with an identity of the Domain specified in the `identity` ECT.
+
+~~~ cddl
+{::include cddl/intrep-domain-mem.cddl}
+~~~
+
+The Domain member Environments are compared with the Environments specified in the list of ECTs in ACS whose `cmtype` matches Evidence.
```
deeglaze, line 1905:
```suggestion
The Domain member Environments are compared with the Environments specified in the list of ECTs in ACS whose `cmtype` matches `&(evidence: 2)`.
```

deeglaze: This sentence is quite confusing. What are the environment comparisons supposed to do?

draft-ietf-rats-corim.md
```diff
@@ -1868,18 +1879,51 @@ If the `selection` criteria is not satisfied, then evaluation procedes to the ne
...
+If the `environment` for every entry of the members of `dm` matches an entry in the ACS ECTs then the Domain identity is copied to every member
+ECT. The `members` are added to the ACS with authority specified in the identity.
```
deeglaze, line 1907:
```suggestion

If for all environments `e` in `dm`.`members`, the ACS contains an ECT with environment `e` with `cmtype: &(evidence: 2)`, then copy the `dm`.`domain-id` to each matching ECT.
```

> The `members` are added to the ACS with authority specified in the identity.

What is "the identity"?
When you add just the environments to the ACS, you're saying what about them, exactly? What part is the condition what part is the addition? Are you always adding new ECTs with nothing else in them but the environment and authority? No claims or cmtype?

draft-ietf-rats-corim.md
```diff
@@ -2351,6 +2425,21 @@ For each `rv` entry, the `condition` ECT is compared with an ACS ECT, where the
 
 If the ECTs match except for authority, the `rv` `addition` ECT authority is added to the ACS ECT authority.
 
+#### Processing Domain Membership {#sec-process-dm}
+
+Domain Membership Triples allow an Endorser (for example an Integrator) to issue an authoritative statement about the composition of an Attester as a collection of Environments.
+If the Verifier Appraisal policy requires Domain Membership, then membership triple sets a reference composition of an Attester in a Verifier database.
```
deeglaze, line 2431: What about the policy requiring domain membership has to do with the interpretation of the membership triple?

draft-ietf-rats-corim.md
```diff
@@ -2351,6 +2425,21 @@ For each `rv` entry, the `condition` ECT is compared with an ACS ECT, where the
...
+At the time of Verification, the reference is then matched with actual composition as reported by an Attester in the Evidence.
+
+Domain Membership Triples are first transformed into an internal representation following the steps mentioned in {{sec-ir-dm-trans}} leading to a representation as specified in {{sec-ir-dm}}.
+
+For each `ECT` entry in the list of `members`, the environment elements (example Class ID) is compared with the equivalent environment elements in ACS Entry with `cmtype` as `evidence`.
```
deeglaze, line 2436:
```suggestion
A list of mutable references to matching ACS Entries for a `dm` is constructed in an iterative process over `dm`.`members`.
For every environment in `dm`.`members`, select ECTs from the ACS where `cmtype: &(evidence: 2)` and whose environment matches.
If there are no such ECTs, then the `dm` processing fails.
If there is at least one such ECT, then add all to the list.

At the end of list construction, for each ECT in the list of references, copy the `dm`.`domain-id` to the ECT's `domain-id` field.
If the field is already present and denotes a different domain, then this is an error.
```
I think this could be clearer. I don't think I understood from previous texts that an environment could belong to at most one domain.

draft-ietf-rats-corim.md
```diff
@@ -1772,6 +1778,8 @@ Environment-Claims Tuples (ECT) have five attributes:
 
 * Profile : The profile that defines this tuple. If no profile is used, this attribute is omitted.
 
+* Domain Identifier : The Domain to which a given ECT belongs.
```
deeglaze, line 1781: Why can an environment only belong to one domain? What if I need to change a domain-id naming convention and keep both the old and new around for migration's sake?

It seems like you could have a lot of superimposed topologies that are acceptable, such that a single identifier is too restrictive.

draft-ietf-rats-corim.md
andrew-draper: It is unclear to me where the authority comes from. Is it the entity which signed the domain membership triple? Or the entity which signed the evidence? Or something else?

Please clarity.

draft-ietf-rats-corim.md
yogeshbdeshpande: We have to start with only one domain, to begin with...

A new DM triple should over-ride an existing domain concept.!

draft-ietf-rats-corim.md
yogeshbdeshpande: I have taken your edits to the main comparison section, as I agree the changes `you proposed` are more clear!

draft-ietf-rats-corim.md
```diff
@@ -219,6 +219,9 @@ Class ID:
 : An identifier for an Environment that is shared among similar Environment instances, such as those with the same hardware assembly.
 See also {{Section 4.2.4 of -eat}}.
 
+Domain:
+: A topological description to represent an entity, for example a Composite Device ({{Section 3.3 of -rats-arch}}), in terms of its constituent Environments or Domains.
```
thomas-fossati, line 223: Defining Domain in terms of Domains is like looking in an infinity mirror :-)

draft-ietf-rats-corim.md
```diff
@@ -2351,6 +2455,20 @@ For each `rv` entry, the `condition` ECT is compared with an ACS ECT, where the
...
 
+#### Processing Domain Membership {#sec-process-dm}
+
+Domain Membership Triples allow an Endorser (for example an Integrator) to issue an authoritative statement about the composition of an Attester as a collection of Environments.
+If the Verifier Appraisal policy requires Domain Membership, then membership triple is used to match a reference composition of an Attester, expressed via membership triple with the actual composition represented via Evidence at the time of Appraisal.
+
+Domain Membership Triples are first transformed into an internal representation following the steps mentioned in {{sec-ir-dm-trans}} leading to a representation as specified in {{sec-ir-dm}}.
+
+A list of mutable references to matching ACS Entries for a `dm` is constructed in an iterative process over `dm`.`members`.
+For every environment in `dm`.`members`, select ECTs from the ACS where `cmtype: &(evidence: 2)` and whose environment matches.
```
nedmsmith, line 2466:
```suggestion
For every environment in `dm`.`members`, select Evidence ECTs from the ACS (i.e., `cmtype: &(evidence: 2)`) where the environments match.
```

nedmsmith (commented):
see suggested wording change

nedmsmith (approved):

draft-ietf-rats-corim.md
yogeshbdeshpande: Removed it  from here and moved it to Processing of DMT section

draft-ietf-rats-corim.md
yogeshbdeshpande: Changes done!

draft-ietf-rats-corim.md
yogeshbdeshpande: The authority transformation is now explained under: `Domain Membership Triples Transformation`. Moved from here!

yogeshbdeshpande:
@thomas-fossati : @henkbirkholz @deeglaze Incorporated your comments. Please approve so that we can make progress!

draft-ietf-rats-corim.md
```diff
@@ -219,6 +219,11 @@ Class ID:
 : An identifier for an Environment that is shared among similar Environment instances, such as those with the same hardware assembly.
 See also {{Section 4.2.4 of -eat}}.
 
+Composite Attester: A Composite Attester is either a Composite Device ({{Section 3.3 of -rats-arch}}) or a Layered Attester ({{Section 3.2 of -rats-arch}}) or any composition involving a combination of one or more Composite Devices or Layered Attesters.
```
thomas-fossati, line 222:
```suggestion
Composite Attester:
: A Composite Attester is either a Composite Device ({{Section 3.3 of -rats-arch}}) or a Layered Attester ({{Section 3.2 of -rats-arch}}) or any composition involving a combination of one or more Composite Devices or Layered Attesters.
```

cddl/domain-type-choice.cddl
```diff
@@ -1,4 +1 @@
-$domain-type-choice /= uint
-$domain-type-choice /= text
-$domain-type-choice /= tagged-uuid-type
-$domain-type-choice /= tagged-oid-type
+$domain-type-choice /= environment-map
```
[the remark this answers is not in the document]
thomas-fossati: Unresolving conversation since this comment has not been addressed.

draft-ietf-rats-corim.md
```diff
@@ -2224,6 +2271,66 @@ The following transformation steps are applied for both the `identity-triples` a
 
 * If the Endorsement conceptual message has a profile, the profile is copied to the `ev`.`addition`.`profile` field.
 
+#### Domain Membership Triples Transformation {#sec-ir-dm-trans}
+
+This section describes how the external representation of a Domain Membership Triple (DMT) {{sec-comid-triple-domain-membership}} is transformed into its CoRIM internal representation {{sec-ir-dm}}.
```
thomas-fossati, line 2276:
```suggestion
This section describes how the external representation of a Domain Membership Triple (DMT) {{sec-comid-triple-domain-membership}} is transformed into its CoRIM internal representation ({{sec-ir-dm}}).
```

draft-ietf-rats-corim.md
```diff
@@ -2236,7 +2343,7 @@ If a triple condition does not match, then the Verifier continues to process oth
 
 ### ACS Requirements {#sec-acs-reqs}
 
-At the end of the Evidence collection process Evidence has been converted into an internal represenetation suitable for appraisal.
+At the end of the Evidence collection process Evidence has been converted into an internal representation suitable for appraisal.
```
thomas-fossati, line 2346:
```suggestion
At the end of the Evidence collection process, the Evidence has been converted into an internal representation suitable for appraisal.
```

draft-ietf-rats-corim.md
```diff
@@ -2356,6 +2463,20 @@ For each `rv` entry, the `condition` ECT is compared with an ACS ECT, where the
 
 If the ECTs match except for authority, the `rv` `addition` ECT authority is added to the ACS ECT authority.
 
+#### Processing Domain Membership {#sec-process-dm}
+
+Domain Membership Triples allow an Endorser (for example an Integrator) to issue an authoritative statement about the composition of an Attester as a collection of Environments.
```
thomas-fossati, line 2468:
```suggestion
Domain Membership Triples allow an Endorser (for example, an Integrator) to issue an authoritative statement about the composition of an Attester as a collection of Environments.
```

draft-ietf-rats-corim.md
```diff
@@ -2356,6 +2463,20 @@ For each `rv` entry, the `condition` ECT is compared with an ACS ECT, where the
 
 If the ECTs match except for authority, the `rv` `addition` ECT authority is added to the ACS ECT authority.
 
+#### Processing Domain Membership {#sec-process-dm}
+
+Domain Membership Triples allow an Endorser (for example, an Integrator) to issue an authoritative statement about the composition of an Attester as a collection of Environments.
+If the Verifier Appraisal policy requires Domain Membership, then membership triple is used to match a reference composition of an Attester, expressed via membership triple with the actual composition represented via Evidence at the time of Appraisal.
```
thomas-fossati, line 2469:
```suggestion
If the Verifier Appraisal policy requires Domain Membership, the Domain Membership Triple is used to match an Attester's reference composition with the actual composition represented by Evidence.
```

draft-ietf-rats-corim.md
```diff
@@ -2356,6 +2463,20 @@ For each `rv` entry, the `condition` ECT is compared with an ACS ECT, where the
 
 If the ECTs match except for authority, the `rv` `addition` ECT authority is added to the ACS ECT authority.
 
+#### Processing Domain Membership {#sec-process-dm}
+
+Domain Membership Triples allow an Endorser (for example, an Integrator) to issue an authoritative statement about the composition of an Attester as a collection of Environments.
+If the Verifier Appraisal policy requires Domain Membership, the Domain Membership Triple is used to match an Attester's reference composition with the actual composition represented by Evidence.
+
+Domain Membership Triples are first transformed into an internal representation following the steps mentioned in {{sec-ir-dm-trans}} leading to a representation as specified in {{sec-ir-dm}}.
```
thomas-fossati, line 2471:
```suggestion
Domain Membership Triples are first transformed into an internal representation following the steps described in {{sec-ir-dm-trans}}, resulting in the representation specified in {{sec-ir-dm}}.
```

draft-ietf-rats-corim.md
```diff
@@ -2356,6 +2463,20 @@ For each `rv` entry, the `condition` ECT is compared with an ACS ECT, where the
...
+
+Domain Membership Triples allow an Endorser (for example, an Integrator) to issue an authoritative statement about the composition of an Attester as a collection of Environments.
+If the Verifier Appraisal policy requires Domain Membership, the Domain Membership Triple is used to match an Attester's reference composition with the actual composition represented by Evidence.
+
+Domain Membership Triples are first transformed into an internal representation following the steps described in {{sec-ir-dm-trans}}, resulting in the representation specified in {{sec-ir-dm}}.
+
+A list of mutable references to matching ACS Entries for a `dm` is constructed in an iterative process over `dm`.`members`.
+For every environment in `dm`.`members`, select Evidence ECTs from the ACS ( i.e. `cmtype: &(evidence: 2)`) where environment match.
+If there are no such ECTs, then the `dm` processing fails.
+If there is at least one such ECT, then add all the ECTs to the list.
```
thomas-fossati, line 2476:
```suggestion
If there is at least one such ECT, then add all the matching ECTs to the list.
```

draft-ietf-rats-corim.md
```diff
@@ -2356,6 +2463,20 @@ For each `rv` entry, the `condition` ECT is compared with an ACS ECT, where the
...
+Domain Membership Triples allow an Endorser (for example, an Integrator) to issue an authoritative statement about the composition of an Attester as a collection of Environments.
+If the Verifier Appraisal policy requires Domain Membership, the Domain Membership Triple is used to match an Attester's reference composition with the actual composition represented by Evidence.
+
+Domain Membership Triples are first transformed into an internal representation following the steps described in {{sec-ir-dm-trans}}, resulting in the representation specified in {{sec-ir-dm}}.
+
+A list of mutable references to matching ACS Entries for a `dm` is constructed in an iterative process over `dm`.`members`.
+For every environment in `dm`.`members`, select Evidence ECTs from the ACS ( i.e. `cmtype: &(evidence: 2)`) where environment match.
+If there are no such ECTs, then the `dm` processing fails.
+If there is at least one such ECT, then add all the matching ECTs to the list.
+At the end of list construction, for each ECT in the list of references, copy the `dm`.`domain-id` to the ECT's `domain-id` field.
```
thomas-fossati, line 2477:
```suggestion
After creating the list, for each ECT in the list of references, copy the `dm`.`domain-id` to the ECT's `domain-id` field.
```

draft-ietf-rats-corim.md
```diff
@@ -2356,6 +2463,20 @@ For each `rv` entry, the `condition` ECT is compared with an ACS ECT, where the
...
+
+Domain Membership Triples allow an Endorser (for example, an Integrator) to issue an authoritative statement about the composition of an Attester as a collection of Environments.
+If the Verifier Appraisal policy requires Domain Membership, the Domain Membership Triple is used to match an Attester's reference composition with the actual composition represented by Evidence.
+
+Domain Membership Triples are first transformed into an internal representation following the steps described in {{sec-ir-dm-trans}}, resulting in the representation specified in {{sec-ir-dm}}.
+
+A list of mutable references to matching ACS Entries for a `dm` is constructed in an iterative process over `dm`.`members`.
+For every environment in `dm`.`members`, select Evidence ECTs from the ACS ( i.e. `cmtype: &(evidence: 2)`) where environment match.
+If there are no such ECTs, then the `dm` processing fails.
+If there is at least one such ECT, then add all the matching ECTs to the list.
```
thomas-fossati, line 2476:
```suggestion
Otherwise, add all the matching ECTs to the list.
```

draft-ietf-rats-corim.md
```diff
@@ -2356,6 +2463,20 @@ For each `rv` entry, the `condition` ECT is compared with an ACS ECT, where the
...
+If the Verifier Appraisal policy requires Domain Membership, the Domain Membership Triple is used to match an Attester's reference composition with the actual composition represented by Evidence.
+
+Domain Membership Triples are first transformed into an internal representation following the steps described in {{sec-ir-dm-trans}}, resulting in the representation specified in {{sec-ir-dm}}.
+
+A list of mutable references to matching ACS Entries for a `dm` is constructed in an iterative process over `dm`.`members`.
+For every environment in `dm`.`members`, select Evidence ECTs from the ACS ( i.e. `cmtype: &(evidence: 2)`) where environment match.
+If there are no such ECTs, then the `dm` processing fails.
+Otherwise, add all the matching ECTs to the list.
+After creating the list, for each ECT in the list of references, copy the `dm`.`domain-id` to the ECT's `domain-id` field.
+If the field is already present and denotes a different domain, then this is an error.
```
thomas-fossati, line 2478:
```suggestion
If the field is already present and contains a different domain, then this is an error.
```

thomas-fossati (commented):
I have made some minor clarifying/declunkifying cleanup, and raised #440 to track an unresolved comment.

IMO, the only remaining issue to address before merging is https://github.com/ietf-rats-wg/draft-ietf-rats-corim/pull/416/files#r2103246835, for which I wasn’t able to make a concrete proposal (too late, too tired, sorry!).

draft-ietf-rats-corim.md
```diff
@@ -2224,6 +2271,66 @@ The following transformation steps are applied for both the `identity-triples` a
 
 * If the Endorsement conceptual message has a profile, the profile is copied to the `ev`.`addition`.`profile` field.
 
+#### Domain Membership Triples Transformation {#sec-ir-dm-trans}
+
+This section describes how the external representation of a Domain Membership Triple (DMT) {{sec-comid-triple-domain-membership}} is transformed into its CoRIM internal representation ({{sec-ir-dm}}).
```
thomas-fossati, line 2276:
```suggestion
This section describes how the external representation of a Domain Membership Triple (DMT) ({{sec-comid-triple-domain-membership}}) is transformed into its CoRIM internal representation ({{sec-ir-dm}}).
```

draft-ietf-rats-corim.md
```diff
@@ -2224,6 +2271,66 @@ The following transformation steps are applied for both the `identity-triples` a
...
+#### Domain Membership Triples Transformation {#sec-ir-dm-trans}
+
+This section describes how the external representation of a Domain Membership Triple (DMT) ({{sec-comid-triple-domain-membership}}) is transformed into its CoRIM internal representation ({{sec-ir-dm}}).
+
+{:dmt-enum: counter="dmt1" style="format Step %d."}
+
+{: dmt-enum}
+* Allocate a `dm` entry.
+
+* Set the conceptual message type for the domain-id ECT and each member ECT to 6 (`domain-member`).
```
thomas-fossati, line 2283: Not sure I follow: at step 1, we allocate the `dm` entry, which is supposedly empty (since we haven’t explained yet how to populate it).  Therefore, here we instruct on setting the `cm-type` field of ECTs that don’t exist?

draft-ietf-rats-corim.md
```diff
@@ -2224,6 +2271,66 @@ The following transformation steps are applied for both the `identity-triples` a
...
+
+{:dmt4-enum: counter="dmt4" style="format %i"}
+
+{: dmt4-enum}
+* **copy**(`domain-member`, `dm`.`domain-id`.`cmtype`)
```
thomas-fossati, line 2288: The types of this assignment are not compatible.

draft-ietf-rats-corim.md
```diff
@@ -2224,6 +2271,66 @@ The following transformation steps are applied for both the `identity-triples` a
...
+
+{: dmt4-enum}
+* For each ECT e in DMT.`members`:
+
+> > **copy**(`domain-member`, `dm`.`members`[e].`cmtype`)
```
thomas-fossati, line 2293: ditto

thomas-fossati (changes requested):
Having taken a slightly deeper look into the DMT transformation algorithm, I think it needs some TLC before we can merge this.

draft-ietf-rats-corim.md
```diff
@@ -1872,18 +1886,47 @@ If the `selection` criteria is not satisfied, then evaluation procedes to the ne
...
 |           | `element-list`  | Mandatory   |
 |           | `authority`     | Mandatory   |
 |           | `cmtype`        | Mandatory   |
 |           | `profile`       | Optional    |
+|           | `domain-id`     | n/a         |
 {: #tbl-ev-ect-optionality title="Endorsed Values and Endorsed Values Series tuples requirements"}
 
+### Internal Representation of Domain Membership {#sec-ir-dm}
+
+An internal representation of Domain Membership uses the `dm` relation, which is a list of ECTs that describes the composition of an Attester with an identity of the Domain specified in the `identity` ECT.
```
yogeshbdeshpande, line 1906:
```suggestion
An internal representation of Domain Membership uses the `dm` relation, which is a list of ECTs expressed in `members` that describes the composition of an Attester with an identity of the Domain specified in the `domain-id` ECT.
```

draft-ietf-rats-corim.md
yogeshbdeshpande: Why? Both RHS and LHS are cm-type?

draft-ietf-rats-corim.md
yogeshbdeshpande: Why? Both RHS and LHS are cm-type?

yogeshbdeshpande:
Will be addressed using: https://github.com/ietf-rats-wg/draft-ietf-rats-corim/pull/454

PR #434, opened 2025-05-07
Support multi instance endorsement

andrew-draper:
Fixes issue #415

### DIFF
--- a/cddl/instance-id-type-choice.cddl
+++ b/cddl/instance-id-type-choice.cddl
@@ -2,3 +2,4 @@ $instance-id-type-choice /= tagged-ueid-type
 $instance-id-type-choice /= tagged-uuid-type
 $instance-id-type-choice /= $crypto-key-type-choice
 $instance-id-type-choice /= tagged-bytes
+$instance-id-type-choice /= instance-copy-type
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -2409,7 +2409,7 @@ If key verification succeeds for any _key_:
 
 Otherwise, do not add the `addition` ECT to the ACS.
 
-#### Copying instance field from a condition
+#### Copying instance field from a condition {#sec-comid-instance-copy}
 
 A CoRIM author may need to create a conditional endorsement which applies to all devices of a particular type, regardless of their instance field.
 If the first `stateless-environment` in the `conditions` field does not contain an `instance` codepoint then it will match against any environment with the same `class`.
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -916,6 +916,8 @@ UEID, UUID, variable-length opaque byte string ({{sec-common-tagged-bytes}}), or
 
 ~~~ cddl
 {::include cddl/instance-id-type-choice.cddl}
+
+{::include cddl/instance-copy-type.cddl}
 ~~~
 
 ##### Environment Group {#sec-comid-group}
@@ -2407,6 +2409,21 @@ If key verification succeeds for any _key_:
 
 Otherwise, do not add the `addition` ECT to the ACS.
 
+#### Copying instance field from a condition
+
+A CoRIM author may need to create a conditional endorsement which applies to all devices of a particular type, regardless of their instance field.
+If the first `stateless-environment` in the `conditions` field does not contain an `instance` codepoint then it will match against any environment with the same `class`.
+
+If the endorsement `environment-map` does not contain an `instance` field then the endorsements will be added without an instance field.
+This may cause problems if the ACS contains evidence from multiple target environments of the same type.
+
+The endorsement `enviromnent-map`/`instance` can be tagged to indicate that it is an `instance-copy`, with an integer argument.
+This type instructs the verifier to copy the `instance` field from the referenced condition to the endorsement.
+
+The integer argument is used to select which instance field to copy.
+The value 0 indicates that the `instance` field from the ACS entry which matched the first `environment-map` in the same triple as the `instance-copy` environment should be copied;
+1 indicates the ACS entry which matched the second `environment-map` etc.
+
 ### Examples for optional phases 5, 6, and 7 {#sec-phases567}
 
 Phases 5, 6, and 7 are optional depending on implementation design.
@@ -2774,7 +2791,8 @@ IANA is requested to allocate the following tags in the "CBOR Tags" registry {{!
 |     562 | `bytes`             | tagged-pkix-asn1der-cert-type, see {{sec-crypto-keys}}        | {{&SELF}} |
 |     563 | `tagged-masked-raw-value` | tagged-masked-raw-value, see {{sec-comid-raw-value-types}} | {{&SELF}} |
 |     564 | `array`             | tagged-int-range, see {{sec-comid-raw-int}}                   | {{&SELF}} |
-| 565-599 | `any`               | Earmarked for CoRIM                                           | {{&SELF}} |
+|     565 | `instance-copy`     | instance-copy, see {{sec-comid-instance-copy}}                | {{&SELF}} |
+| 566-599 | `any`               | Earmarked for CoRIM                                           | {{&SELF}} |
 
 Tags designated as "Earmarked for CoRIM" can be reassigned by IANA based on advice from the designated expert for the CBOR Tags registry.
 
--- a/cddl/instance-copy-type.cddl
+++ b/cddl/instance-copy-type.cddl
@@ -1,0 +1,1 @@
+instance-copy-type = #6.565(uint)


draft-ietf-rats-corim.md
```diff
@@ -2407,6 +2409,21 @@ If key verification succeeds for any _key_:
...
 
+#### Copying instance field from a condition {#sec-comid-instance-copy}
+
+A CoRIM author may need to create a conditional endorsement which applies to all devices of a particular type, regardless of their instance field.
+If the first `stateless-environment` in the `conditions` field does not contain an `instance` codepoint then it will match against any environment with the same `class`.
+
+If the endorsement `environment-map` does not contain an `instance` field then the endorsements will be added without an instance field.
+This may cause problems if the ACS contains evidence from multiple target environments of the same type.
+
+The endorsement `enviromnent-map`/`instance` can be tagged to indicate that it is an `instance-copy`, with an integer argument.
```
deeglaze, line 2420:
```suggestion
The endorsement's `environment-map`.`instance` can be tagged to indicate that it is an `instance-copy`, with an integer argument.
```

draft-ietf-rats-corim.md
```diff
@@ -2407,6 +2409,21 @@ If key verification succeeds for any _key_:
 
 Otherwise, do not add the `addition` ECT to the ACS.
 
+#### Copying instance field from a condition {#sec-comid-instance-copy}
+
+A CoRIM author may need to create a conditional endorsement which applies to all devices of a particular type, regardless of their instance field.
+If the first `stateless-environment` in the `conditions` field does not contain an `instance` codepoint then it will match against any environment with the same `class`.
+
+If the endorsement `environment-map` does not contain an `instance` field then the endorsements will be added without an instance field.
```
deeglaze, line 2417:
```suggestion
If the endorsement `environment-map` does not contain an `instance` field, then the endorsements will be added without an instance field.
```

deeglaze: This doesn't match my understanding. You add endorsements to the ECTs that are selected by the condition, which can also have an existing instance field. Your words suggest that there'd be an ECT without an instance field with these endorsements.

When you discuss "the endorsement" you're talking about not the endorsement triples. You're talking about conditional endorsement (series) triples' non-condition component. It's a bit too imprecise.

draft-ietf-rats-corim.md
```diff
@@ -2407,6 +2409,21 @@ If key verification succeeds for any _key_:
...
+This may cause problems if the ACS contains evidence from multiple target environments of the same type.
+
+The endorsement `enviromnent-map`/`instance` can be tagged to indicate that it is an `instance-copy`, with an integer argument.
+This type instructs the verifier to copy the `instance` field from the referenced condition to the endorsement.
+
+The integer argument is used to select which instance field to copy.
```
deeglaze, line 2423: This is a different paragraph, so you've lost the context about what this argument is. Arguments are for function calls. Fields are for objects/maps. I'm not sure what you're talking about.

Instance-copy-type references "which instance field" to copy, which doesn't make sense given that instance is a field of environment, and there's no multiplicity to select amongst, no?

The following description is now imposing additional semantics to the appraisal algorithm that the ACS's order and/or the condition matching order is relevant to the interpretation of the triple, which is overconstrained and underspecified in the appraisal algorithm description section.

draft-ietf-rats-corim.md
```diff
@@ -2407,6 +2409,21 @@ If key verification succeeds for any _key_:
 
 Otherwise, do not add the `addition` ECT to the ACS.
 
+#### Copying instance field from a condition {#sec-comid-instance-copy}
+
+A CoRIM author may need to create a conditional endorsement which applies to all devices of a particular type, regardless of their instance field.
```
deeglaze, line 2414:
```suggestion
A CoRIM author may need to create a conditional endorsement which applies to all devices of a particular type, regardless of their instance field.
```
We don't talk about devices and their types. We only talk about environments and their class/group/instance. What does this mean?

draft-ietf-rats-corim.md
```diff
@@ -2407,6 +2409,21 @@ If key verification succeeds for any _key_:
...
+If the first `stateless-environment` in the `conditions` field does not contain an `instance` codepoint then it will match against any environment with the same `class`.
```
deeglaze, line 2415:
```suggestion
If the first `stateless-environment` in the `conditions` field does not contain an `instance` codepoint, then it will match against any environment with the same `class`.
```
This is roughly already the case, modulo `class` being `class` and/or `group`.

deeglaze (changes requested):

draft-ietf-rats-corim.md
andrew-draper:
```suggestion
A CoRIM author may need to create a conditional endorsement which applies to all measurements which have the same `global` and/or `class` field within their `environment-map`, regardless of their `instance` field.
```

draft-ietf-rats-corim.md
andrew-draper: The text 3 lines ago said that this applies to conditional endorsements. A conditional endorsement can add the endorsement to any `environment-map` + `mkey` location.

draft-ietf-rats-corim.md
nedmsmith: We didn't define `global`. Where is this coming from?

draft-ietf-rats-corim.md
nedmsmith: > this applies to conditional endorsements

I don't think it applies to series endorsements since `conditional-series-record`.`addition` is only `measurement-map`. The ECT(s) updated are those  that satisfy the `condition` and `selection` criteria.

Only `conditional-endorsement-triple-record` can create a completely new ECT or update a different ECT from the ones that satisfied the `condition`.

draft-ietf-rats-corim.md
nedmsmith: > `stateless-environment`

This isn't in the cddl. Possibly it should say "environment"?

In general, matching applies to all the environment fields. If there is a class, then it will match the class even if there is also an instance or group. The interesting case is if there isn't a class but you want the addition to apply to a class of ECTs.

cddl/examples/comid-cend-inst-copy.diag
```diff
@@ -0,0 +1,84 @@
...
+        ]
+      ], /*** end stateful-environment-record ***/
+      [  /*** endorsements ***/
+        [ /*** endorsed-triple-record ***/
+          / environment-map #2 / {
+            / class / 0 : {
+              / class-id / 0 : / tagged-oid-type / 111(h'5502C000'),
+              / vendor / 1 : "ACME Inc.",
+              / model / 2 : "ACME RoadRunner Firmware",
+              / layer / 3 : 1
```
nedmsmith, line 64: Since record #0 has layer=0, the addition of 565(0) will apply to ECTs with layer=1. Seems to contradict the comment in line 66.

cddl/examples/comid-cend-inst-copy.diag
```diff
@@ -0,0 +1,84 @@
...
+            },
+            / instance / 1 : 565(0) / *** Insert instance value which matched against stateful-environment-record #0 *** /
```
nedmsmith, line 66: Since the matching rule for record #0 didn't include an instance-id, the entire class of ECTs was matched. However, that doesn't mean these ECTs didn't also have instance-id (it just wasn't part of the matching criteria). 
The addition creates a new ECT with class-id that's the same as record #1 and instance-id 565(0) that's also the same as record #1 and adds a version claim.

It isn't clear what significance record #0 has, but that may be opaque to the Verifier. 

The ACS now has a couple (at least) record #1 ECTs where one has version and the others presumably don't. These could be merged into a single ECT (assuming authority is the same for all values) without changing the state of the ACS. If they did have version (and the values differed) the merge would be lossy. 

In general, I don't see how 565() is treated differently from any other instance-id type. 

If the objective is to create "update" semantics for conditional endorsement triple, where the current semantics are "addition", then doing that via CBOR tag in a type definition seems like the wrong approach.

cddl/examples/comid-cend-inst-copy.diag
```diff
@@ -0,0 +1,84 @@
...
+        ], 
+        [ / *** stateful-environment-record entry #1 *** /
+          / environment-map / {
+            / class / 0 : {
+              / class-id / 0 : / tagged-oid-type / 111(h'5502C000'),
+              / vendor / 1 : "ACME Inc.",
+              / model / 2 : "ACME RoadRunner Firmware",
+              / layer / 3 : 1
+            },
+            / instance / 1 : 565(0) / *** Match against ACS entry with instance which stateful-environment-record #0 matched *** /
```
nedmsmith, line 43: Does instance zero (i.e. 565(0)) have special significance? Otherwise, how does Endorser know the instance id if it was dynamically asserted into the Evidence / ACS?

deeglaze: Yes, as a condition, this is a binding position for a variable named 0.
As an endorsement, 565(0) is meant to be replaced with the instance that was bound to that variable during condition matching.

nedmsmith (commented):
multiple

cddl/examples/comid-cend-inst-copy.diag
```diff
@@ -0,0 +1,84 @@
...
+    [
+      [
+        [ / *** stateful-environment-record entry #0 *** /	
+          / environment-map / {	
+            / class / 0 : {
+              / class-id / 0 : / tagged-oid-type / 111(h'5502C000'),	
+              / vendor / 1 : "ACME Inc.",	
+              / model / 2 : "ACME RoadRunner Bootloader",
+              / layer / 3 : 0
+            }
```
andrew-draper, line 21:
```suggestion
            },
```

cddl/examples/comid-cend-inst-copy.diag
andrew-draper: Instance slot usage does not change the class part of `environment-map`

draft-ietf-rats-corim.md
andrew-draper: Text has been changed, this comment is included in those changes.

draft-ietf-rats-corim.md
andrew-draper: Change added.

draft-ietf-rats-corim.md
andrew-draper: This text has been rewritten to use named variables instead of positional references.

cddl/examples/comid-cend-inst-copy.diag
```diff
@@ -0,0 +1,84 @@
...
+        ], 
+        [ / *** stateful-environment-record entry #1 *** /
+          / environment-map / {
+            / class / 0 : {
+              / class-id / 0 : / tagged-oid-type / 111(h'5502C000'),
+              / vendor / 1 : "ACME Inc.",
+              / model / 2 : "ACME RoadRunner Firmware",
+              / layer / 3 : 1
+            },
+            / instance / 1 : / slot-use / 566(0) / *** Match against value in instance slot #0 *** /
```
deeglaze, line 43: Use should not be in the condition. It should only be in the addition.

cddl/examples/comid-cend-inst-copy.diag
```diff
@@ -0,0 +1,84 @@
...
+      [
+        [ / *** stateful-environment-record entry #0 *** /	
+          / environment-map / {	
+            / class / 0 : {
+              / class-id / 0 : / tagged-oid-type / 111(h'5502C000'),	
+              / vendor / 1 : "ACME Inc.",	
+              / model / 2 : "ACME RoadRunner Bootloader",
+              / layer / 3 : 0
+            },
+            / instance / 1 : / slot-store / 565(0) / *** Match against any instance, store values from ACS entry in instance slot #0 *** /
```
deeglaze, line 22:
```suggestion
            / instance / 1 : / slot-bind / 565(0) / *** Match against any instance, store values from ACS entry in instance slot #0 *** /
```

draft-ietf-rats-corim.md
```diff
@@ -2372,6 +2374,25 @@ If the ECTs match ({{sec-match-condition-ect}}), the `ev` `addition` ECT is adde
...
+Both instance types contain an integer name for the variable which is set by the `instance-slot-store-type` and used by the `instance-slot-use-type`.
+
+Within each triple, each variable can only be set once, so if there are multiple `environment-map`s using `instance-slot-store-type` then they must each have a different integer name.
+
+After successfully matching a `stateful-environment-record` containing an `instance-slot-store-type` against an ACS entry, the verifier SHALL copy the instance value from that ACS entry to the corresponding slot variable.
+If the matching ACS entry does not include an instance then the slot variable is marked as initialised, but empty.
+
+When adding a conditional endorsement whose `enviroment-map`.`instance` field is an `instance-slot-use-type` to the ACS, the verifier SHALL set the `enviroment-map`.`instance` field from the corresponding slot variable.
+If the corresponding slot variable is initialised but empty then the verifier SHALL NOT add an `instance` field.
```
deeglaze, line 2391:
```suggestion
If the corresponding slot variable is initialized but empty then the verifier SHALL NOT add an `instance` field.
```

draft-ietf-rats-corim.md
```diff
@@ -2474,6 +2495,23 @@ If any field which is present in the condition ECT `environment-map` is not bina
...
+These fields are used to ensure that multiple conditions, for example within a conditional endorsement, match against the same instance value.
+They are also used to set the instance value in a conditional endorsement ACS entry to match the matched ECTs.
+
+The verifier maintains a small array of instance slots, each slot is identified using a non-negative integer.
+Before processing each triple, all slots SHALL be set to the unused state.
+
+If the condition ECT `environment-map` contains an `instance` field of type `instance-slot-store-type` then it matches against any instance value.
+The instance field in the ACS entry is copied into the instance slot with the relevant `instance-slot-number`.
+
+If the condition ECT `environment-map` contains an `instance` field of type `instance-slot-use-type` and that slot is unused then the environments do not match.
```
deeglaze, line 2509:
```suggestion
If the condition ECT `environment-map` contains an `instance` field of type `instance-slot-reference-type` and that slot is unused then the environments do not match.
```

draft-ietf-rats-corim.md
```diff
@@ -2474,6 +2495,23 @@ If any field which is present in the condition ECT `environment-map` is not bina
...
+
+If the condition ECT `environment-map` contains an `instance` field of type `instance-slot-use-type` and it is not binary identical to the value in that slot then the environments do not match.
```
deeglaze, line 2511:
```suggestion
If the condition ECT `environment-map` contains an `instance` field of type `instance-slot-reference-type` and it is not binary identical to the value in that slot then the environments do not match.
```

draft-ietf-rats-corim.md
```diff
@@ -2474,6 +2495,23 @@ If any field which is present in the condition ECT `environment-map` is not bina
...
+
+If a triple contains two or more condition ECTs which store values into the same instance slot then it is invalid and the environments do not match.
```
deeglaze, line 2513: Delete.

draft-ietf-rats-corim.md
```diff
@@ -2776,7 +2814,9 @@ IANA is requested to allocate the following tags in the "CBOR Tags" registry {{!
 |     562 | `bytes`             | tagged-pkix-asn1der-cert-type, see {{sec-crypto-keys}}        | {{&SELF}} |
 |     563 | `tagged-masked-raw-value` | tagged-masked-raw-value, see {{sec-comid-raw-value-types}} | {{&SELF}} |
 |     564 | `array`             | tagged-int-range, see {{sec-comid-int-range}}                   | {{&SELF}} |
-| 565-599 | `any`               | Earmarked for CoRIM                                           | {{&SELF}} |
+|     565 | `instance-slot-store` | instance-slot-store, see {{sec-comid-instance-copy}}    | {{&SELF}} |
+|     566 | `instance-slot-use` | instance-slot-use, see {{sec-comid-instance-copy}}         | {{&SELF}} |
```
deeglaze, line 2818:
```suggestion
|     566 | `instance-slot-use` | instance-slot-reference, see {{sec-comid-instance-copy}}         | {{&SELF}} |
```

draft-ietf-rats-corim.md
```diff
@@ -2776,7 +2814,9 @@ IANA is requested to allocate the following tags in the "CBOR Tags" registry {{!
 |     562 | `bytes`             | tagged-pkix-asn1der-cert-type, see {{sec-crypto-keys}}        | {{&SELF}} |
 |     563 | `tagged-masked-raw-value` | tagged-masked-raw-value, see {{sec-comid-raw-value-types}} | {{&SELF}} |
 |     564 | `array`             | tagged-int-range, see {{sec-comid-int-range}}                   | {{&SELF}} |
-| 565-599 | `any`               | Earmarked for CoRIM                                           | {{&SELF}} |
+|     565 | `instance-slot-store` | instance-slot-store, see {{sec-comid-instance-copy}}    | {{&SELF}} |
```
deeglaze, line 2817:
```suggestion
|     565 | `instance-slot-store` | instance-slot-bind, see {{sec-comid-instance-copy}}    | {{&SELF}} |
```

cddl/instance-id-type-choice.cddl
```diff
@@ -2,3 +2,5 @@ $instance-id-type-choice /= tagged-ueid-type
 $instance-id-type-choice /= tagged-uuid-type
 $instance-id-type-choice /= $crypto-key-type-choice
 $instance-id-type-choice /= tagged-bytes
+$instance-id-type-choice /= instance-slot-store-type
+$instance-id-type-choice /= instance-slot-use-type
```
deeglaze, line 6:
```suggestion
$instance-id-type-choice /= instance-slot-bind-type
$instance-id-type-choice /= instance-slot-reference-type
```

deeglaze (changes requested):

draft-ietf-rats-corim.md
```diff
@@ -2372,6 +2374,26 @@ If the ECTs match ({{sec-match-condition-ect}}), the `ev` `addition` ECT is adde
...
+Both instance types contain an integer name for the slot variable which is bound by the `instance-slot-bind-type` and referenced by the `instance-slot-ref-type`.
+
+Within each triple, each slot variable can only be bound to one value, so if there are multiple `environment-map`s using `instance-slot-bind-type` then subsequent uses introduce a constraint that the value to bind must be equal to the value already bound to the slot.
+
+After successfully matching a `stateful-environment-record` containing an `instance-slot-bind-type` against an ACS entry, the verifier SHALL copy the instance value from that ACS entry to the corresponding slot variable.
+If the matching ACS entry does not include an instance then the slot variable is marked as bound to the empty value.
+
+When adding a conditional endorsement whose `enviroment-map`.`instance` field is an `instance-slot-ref-type` to the ACS, the verifier SHALL set the `enviroment-map`.`instance` field from the contents of the corresponding slot variable.
```
andrew-draper, line 2390:
```suggestion
When adding a conditional endorsement whose `enviroment-map`.`instance` field is an `instance-slot-ref-type` to the ACS, and that variable is bound to a non-empty value, then the verifier SHALL set the `enviroment-map`.`instance` field from the contents of the corresponding slot variable.
```

draft-ietf-rats-corim.md
```diff
@@ -2372,6 +2374,26 @@ If the ECTs match ({{sec-match-condition-ect}}), the `ev` `addition` ECT is adde
...
+If the corresponding slot variable is bound to the empty value then the verifier SHALL NOT add an `instance` field.
+If the corresponding slot variable is not bound then the triple is invalid - the verifier SHALL NOT add the conditional endorsement to the ACS.
+
+If a conditional endorsement containing stateful environments which use `instance-slot-bind-type` matches against multiple ECTs, then each match is processed independently of the others.
+Each match uses its own slot bindings, and  each match adds a separate endorsement ECT to the ACS.
```
andrew-draper, line 2395:
```suggestion
Each match uses its own slot bindings, and each match adds a separate endorsement ECT to the ACS.
```

cddl/instance-id-type-choice.cddl
```diff
@@ -2,3 +2,5 @@ $instance-id-type-choice /= tagged-ueid-type
 $instance-id-type-choice /= tagged-uuid-type
 $instance-id-type-choice /= $crypto-key-type-choice
 $instance-id-type-choice /= tagged-bytes
+$instance-id-type-choice /= instance-slot-bind-type
+$instance-id-type-choice /= instance-slot-ref-type
```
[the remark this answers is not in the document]
deeglaze: Fair enough.

deeglaze (approved):

cddl/instance-id-type-choice.cddl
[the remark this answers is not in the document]
andrew-draper: Thanks, I will leave it as is then.

nedmsmith:
My intuition is to keep the matching semantics separate from the identifier semantics. 
E.g.:
```
;instance-slot-bind-type = #6.565(instance-slot-number)
;instance-slot-ref-type = #6.566(instance-slot-number)
cursor-slot = uint
tagged-match-instance-id-bind = #6.565(cursor-slot)
tagged-match-instance-id-ref = #6.566(cursor-slot)
$wc-match-type-choice /= tagged-match-instance-id-bind
$wc-match-type-choice /= tagged-match-instance-id-ref
; Add more $wc-match-type-choice entries suitable for wildcard group-id matching

environment-map = non-empty<{
  ? &(class: 0) => class-map
  ? &(instance: 1) => $instance-id-type-choice / $wc-match-type-choice
  ? &(group: 2) => $group-id-type-choice / $wc-match-type-choice
}>
```

I don't know if it makes sense to look at wild card matching on instance identifiers in measurement-values-map such as mac-address and ip-address etc. Possibly, Dionna's idea for generalized wild card matching on any measurement type is better. If corim goes in that direction, then maybe that is a better way to consider wildcard matching for class-map values? This should be a different PR. But possibly, the $wc-match-type-choice would be a common aspect that keeps wc matching separate from other RHS type structures.

andrew-draper:
> My intuition is to keep the matching semantics separate from the identifier semantics....

I like this description of the semantics, and will update to something like this if nobody disagrees.

> I don't know if it makes sense to look at wild card matching on instance identifiers in measurement-values-map such as mac-address and ip-address etc. Possibly, Dionna's idea for generalized wild card matching on any measurement type is better. If corim goes in that direction, then maybe that is a better way to consider wildcard matching for class-map values? This should be a different PR. But possibly, the $wc-match-type-choice would be a common aspect that keeps wc matching separate from other RHS type structures.

This is an interesting idea, but I think it should probably be a different PR. Making the change you propose above will make it easier to implement this second PR.

draft-ietf-rats-corim.md
```diff
@@ -2367,14 +2373,41 @@ Endorsements are added to the ACS if the Endorsement condition is satisifed by t
 
 #### Processing Endorsements {#sec-process-end}
 
-Endorsements are matched with ACS entries by iterating through the `ev` list.
-For each `ev` entry, the `condition` ECT is compared with an ACS ECT, where the ACS ECT `cmtype` contains either `evidence`, `reference-values`, or `endorsements`.
-If the ECTs match ({{sec-match-condition-ect}}), the `ev` `addition` ECT is added to the ACS.
+Endorsed Value Triple and Conditional Endorsement Triple share the same internal representation.
+Both types of triple are transformed into an internal representation based on `ev`, but the CBOR encoding of Endorsed Value Triple can only represent a subset of the CBOR encoding of Conditional Endorsement Triple.
+The Endorsed Values Triple contains a single `environment-map`, which means it can only match against one environment in the ACS; it can only add a single Endorsement ECT; and it uses the same enviromnent for `condition` and `addition`.
```
yogeshbdeshpande, line 2378:
```suggestion
The Endorsed Values Triple contains a single `environment-map`, which means it can only match against one environment in the ACS; it can only add Endorsements for ECT; and pertaining to one Environment
```

draft-ietf-rats-corim.md
```diff
@@ -2367,14 +2373,41 @@ Endorsements are added to the ACS if the Endorsement condition is satisifed by t
...
+
+Note that some condition values are able to match against multiple ACS-ECTs, or sets of ACS-ECTs.
+If there are multiple matches then each match is processed independently from the others.
+
+##### Copying instance and/or group fields from a condition {#sec-comid-instance-group-copy}
+
+A CoRIM author may need to create a conditional endorsement which applies to all all measurements which have the same `class` field within their `environment-map`, regardless of their `instance` and/or `group` fields.
+The `slot-bind-type` and `slot-ref-type` options in these fields can be used to achieve this.
+
+In the simplest case, the CoRIM sets the `enviroment-map`.`instance` field of a `stateful-environment-record` within the triple to hold an `slot-bind-type` and the `enviroment-map`.`instance` field of an `endorsed-triple-record` to hold an `slot-ref-type`.
```
yogeshbdeshpande, line 2394:
```suggestion
In the simplest case, the CoRIM author sets the `environment-map`.`instance` field of a `stateful-environment-record` within the triple to hold an `slot-bind-type` and the `environment-map`.`instance` field of an `endorsed-triple-record` to hold an `slot-ref-type`.
```

draft-ietf-rats-corim.md
```diff
@@ -2367,14 +2373,41 @@ Endorsements are added to the ACS if the Endorsement condition is satisifed by t
...
+Both instance types contain an integer name for the slot variable which is bound by the `slot-bind-type` and referenced by the `slot-ref-type`.
 
-#### Processing Conditional Endorsements {#sec-process-cond-end}
+Within each triple, each slot variable can only be bound to one value, so if there are multiple `environment-map`s using `slot-bind-type` then subsequent uses introduce a constraint that the value to bind must be equal to the value already bound to the slot.
 
-Conditional Endorsement Triples are transformed into an internal representation based on `ev`.
-Conditional endorsements have the same processing steps as shown in ({{sec-process-end}}).
+After successfully matching a `stateful-environment-record` containing an `slot-bind-type` against an ACS entry, the verifier SHALL copy the instance value from that ACS entry to the corresponding slot variable.
```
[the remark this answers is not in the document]
yogeshbdeshpande: It is not clear what is copied from which source to which destination?

Can you please add a clarity here...

draft-ietf-rats-corim.md
```diff
@@ -2367,14 +2373,41 @@ Endorsements are added to the ACS if the Endorsement condition is satisifed by t
...
+If the matching ACS entry does not include an instance then the slot variable is marked as bound to the empty value.
+
+When adding a conditional endorsement whose `enviroment-map`.`instance` field is an `slot-ref-type` to the ACS, and that variable is bound to a non-empty value, then the verifier SHALL set the `enviroment-map`.`instance` field from the contents of the corresponding slot variable.
```
yogeshbdeshpande, line 2402: `enviroment-map`.`instance` field from the contents of the corresponding slot variable to where ??

yogeshbdeshpande (commented):
can you please clarify 
1. The intent what is the intent of inserting slot bind type and slot ref type.
2. Also the actions are not clear who is going from which source to which destination...

draft-ietf-rats-corim.md
andrew-draper: Added similar fix. Thanks.

draft-ietf-rats-corim.md
```diff
@@ -2367,14 +2373,41 @@ Endorsements are added to the ACS if the Endorsement condition is satisifed by t
...
 #### Processing Endorsements {#sec-process-end}
 
-Endorsements are matched with ACS entries by iterating through the `ev` list.
-For each `ev` entry, the `condition` ECT is compared with an ACS ECT, where the ACS ECT `cmtype` contains either `evidence`, `reference-values`, or `endorsements`.
-If the ECTs match ({{sec-match-condition-ect}}), the `ev` `addition` ECT is added to the ACS.
+Endorsed Value Triple and Conditional Endorsement Triple share the same internal representation.
+Both types of triple are transformed into an internal representation based on `ev`, but the CBOR encoding of Endorsed Value Triple can only represent a subset of the CBOR encoding of Conditional Endorsement Triple.
+The Endorsed Values Triple contains a single `environment-map`, which means it can only match against one environment in the ACS; it can only add a single Endorsement ECT; and it uses the same enviromnent for `condition` and `addition`.
+
+After transformation into an `ev` entry, the processing steps of both triples are the same, as described below.
```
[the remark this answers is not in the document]
andrew-draper: The text is talking about what happens after the triples are transformed into `ev` form. What I'm saying is that after this transform the processing is the same.

For example, an endorsed values triple is transformed into an `ev` with:
`condition.environment = environment`
`condition.measurement = <empty>`
`addition.environment = environment`
`addition.measurement = measurement`

A conditional endorsement triple can create exactly the same `ev`.
The results of processing this `ev` will be the same regardless of whether it came from an Endorsed Values Triple or a Conditional Endorsement Triple.

andrew-draper:
I have added introductory text which describes use cases for these values

draft-ietf-rats-corim.md
```diff
@@ -2371,14 +2391,41 @@ Endorsements are added to the ACS if the Endorsement condition is satisifed by t
...
+The `tagged-var-bind` and `tagged-var-ref` options in these fields can be used to achieve this.
+
+In the simplest case, within a CoRIM, the `enviroment-map`.`instance` field of an `ev` condition holds a `tagged-var-bind` and the `enviroment-map`.`instance` field of an addition within that `ev` holds a `tagged-var-ref`.
+Both instance types contain an integer name for the bound variable, which is bound by the `tagged-var-bind` and referenced by the `tagged-var-ref`.
+
+Within each triple, each bound variable can only be bound to one value, so if there are multiple `environment-map`s using `tagged-var-bind` then subsequent uses introduce a constraint that the value to bind must be equal to the value already bound to the slot.
+
+After successfully matching the `class` field within a `stateful-environment-record` whose `instance` field is a `tagged-var-bind` against an ACS entry, the Verifier SHALL copy the `instance` value from that ACS entry to the corresponding bound variable.
```
yogeshbdeshpande, line 2417: copy the `instance` value from that ACS entry to the corresponding bound variable?

Did you mean the ACS entry from Evidence?

draft-ietf-rats-corim.md
```diff
@@ -931,10 +933,28 @@ anonymity set.
...
+##### Environment Wild-card Binding {#sec-comid-wc}
+
+The `tagged-var-bind` and `tagged-var-ref` values are used for constrained wildcard processing.
+
+~~~ cddl
+{::include cddl/wc-match-type-choice.cddl}
+~~~
+
+An `environment-map` without an `instance` field used in a condition will match against an ACS-ECT with any `instance` field.
+When the ACS contains evidence from multiple Attesters, and that evidence uses the `instance` field to indicate which Attester created the evidence, these values are used to indicate the relationships between environments.
```
andrew-draper, line 951:
```suggestion
When the ACS contains evidence from multiple Target Environments, and that evidence uses the `instance` field to indicate which Target Environment created the evidence, these values are used to indicate the relationships between environments.
```

draft-ietf-rats-corim.md
```diff
@@ -2371,14 +2391,41 @@ Endorsements are added to the ACS if the Endorsement condition is satisifed by t
...
 
-Conditional Endorsement Triples are transformed into an internal representation based on `ev`.
-Conditional endorsements have the same processing steps as shown in ({{sec-process-end}}).
+A CoRIM author may need to create a conditional endorsement which applies to all measurements which have the same `class` field within their `environment-map`, regardless of their `instance` and/or `group` fields.
+The `tagged-var-bind` and `tagged-var-ref` options in these fields can be used to achieve this.
+
+In the simplest case, within a CoRIM, the `enviroment-map`.`instance` field of an `ev` condition holds a `tagged-var-bind` and the `enviroment-map`.`instance` field of an addition within that `ev` holds a `tagged-var-ref`.
+Both instance types contain an integer name for the bound variable, which is bound by the `tagged-var-bind` and referenced by the `tagged-var-ref`.
+
+Within each triple, each bound variable can only be bound to one value, so if there are multiple `environment-map`s using `tagged-var-bind` then subsequent uses introduce a constraint that the value to bind must be equal to the value already bound to the slot.
```
andrew-draper, line 2415:
```suggestion
Within each triple, each bound variable can only be bound to one value, so if there are multiple `environment-map`s using `tagged-var-bind` then subsequent uses introduce a constraint that the value to bind must be equal to the value already bound to the variable.
```

draft-ietf-rats-corim.md
```diff
@@ -2371,14 +2391,41 @@ Endorsements are added to the ACS if the Endorsement condition is satisifed by t
...
+
+After successfully matching the `class` field within a `stateful-environment-record` whose `instance` field is a `tagged-var-bind` against an ACS entry, the Verifier SHALL copy the `instance` value from that ACS entry to the corresponding bound variable.
+If the matching ACS entry does not include an instance then the bound variable is marked as bound to the empty value.
+
+When adding an `ev` addition ECT whose `enviroment-map`.`instance` field is a `tagged-var-ref` to the ACS, if the variable is bound to a non-empty value, then the Verifier SHALL copy the value of the corresponding bound variable to the `enviroment-map`.`instance` field of the new ACS-ECT.
+If the corresponding bound variable is bound to the empty value then the Verifier SHALL NOT add an `instance` field.
+If the corresponding bound variable is not bound then the triple is invalid - the Verifier SHALL NOT add the conditional endorsement to the ACS.
```
andrew-draper, line 2422:
```suggestion
If the corresponding bound variable is not bound then the triple is invalid - the Verifier SHALL NOT add the addition ECT to the ACS.
```

draft-ietf-rats-corim.md
```diff
@@ -2371,14 +2391,41 @@ Endorsements are added to the ACS if the Endorsement condition is satisifed by t
...
+The `tagged-var-bind` and `tagged-var-ref` options in these fields can be used to achieve this.
+
+In the simplest case, within a CoRIM, the `enviroment-map`.`instance` field of an `ev` condition holds a `tagged-var-bind` and the `enviroment-map`.`instance` field of an addition within that `ev` holds a `tagged-var-ref`.
+Both instance types contain an integer name for the bound variable, which is bound by the `tagged-var-bind` and referenced by the `tagged-var-ref`.
+
+Within each triple, each bound variable can only be bound to one value, so if there are multiple `environment-map`s using `tagged-var-bind` then subsequent uses introduce a constraint that the value to bind must be equal to the value already bound to the variable.
+
+After successfully matching the `class` field within a `stateful-environment-record` whose `instance` field is a `tagged-var-bind` against an ACS entry, the Verifier SHALL copy the `instance` value from that ACS entry to the corresponding bound variable.
```
andrew-draper, line 2417:
```suggestion
After successfully matching the `class` field within a `ev` condition whose `instance` field is a `tagged-var-bind` against an ACS entry, the Verifier SHALL copy the `instance` value from that ACS entry to the corresponding bound variable.
```

draft-ietf-rats-corim.md
```diff
@@ -2371,14 +2391,41 @@ Endorsements are added to the ACS if the Endorsement condition is satisifed by t
...
 
-Conditional Endorsement Triples are transformed into an internal representation based on `ev`.
-Conditional endorsements have the same processing steps as shown in ({{sec-process-end}}).
+A CoRIM author may need to create a conditional endorsement which applies to all measurements which have the same `class` field within their `environment-map`, regardless of their `instance` and/or `group` fields.
+The `tagged-var-bind` and `tagged-var-ref` options in these fields can be used to achieve this.
+
+In the simplest case, within a CoRIM, the `enviroment-map`.`instance` field of an `ev` condition holds a `tagged-var-bind` and the `enviroment-map`.`instance` field of an addition within that `ev` holds a `tagged-var-ref`.
+Both instance types contain an integer name for the bound variable, which is bound by the `tagged-var-bind` and referenced by the `tagged-var-ref`.
+
+Within each triple, each bound variable can only be bound to one value, so if there are multiple `environment-map`s using `tagged-var-bind` then subsequent uses introduce a constraint that the value to bind must be equal to the value already bound to the variable.
```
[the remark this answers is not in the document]
andrew-draper: TODO: Find a way to describe this without using bound variable terminology

cddl/wc-match-type-choice.cddl
```diff
@@ -0,0 +1,8 @@
+$wc-match-type-choice /= tagged-var-bind
+$wc-match-type-choice /= tagged-var-ref
+
+tagged-var-bind = #6.565(var-number)
+
+tagged-var-ref = #6.566(var-number)
```
andrew-draper, line 6: TODO: Merge these tags into one called tagged-variable

draft-ietf-rats-corim.md
```diff
@@ -2371,14 +2391,41 @@ Endorsements are added to the ACS if the Endorsement condition is satisifed by t
...
+
+After successfully matching the `class` field within a `ev` condition whose `instance` field is a `tagged-var-bind` against an ACS entry, the Verifier SHALL copy the `instance` value from that ACS entry to the corresponding bound variable.
+If the matching ACS entry does not include an instance then the bound variable is marked as bound to the empty value.
+
+When adding an `ev` addition ECT whose `enviroment-map`.`instance` field is a `tagged-var-ref` to the ACS, if the variable is bound to a non-empty value, then the Verifier SHALL copy the value of the corresponding bound variable to the `enviroment-map`.`instance` field of the new ACS-ECT.
+If the corresponding bound variable is bound to the empty value then the Verifier SHALL NOT add an `instance` field.
+If the corresponding bound variable is not bound then the triple is invalid - the Verifier SHALL NOT add the addition ECT to the ACS.
```
deeglaze, line 2422:
```suggestion
If the corresponding bound variable is not bound then the triple is invalid—the Verifier SHALL NOT add the addition ECT to the ACS.
```

draft-ietf-rats-corim.md
```diff
@@ -2371,14 +2392,42 @@ Endorsements are added to the ACS if the Endorsement condition is satisifed by t
 
 #### Processing Endorsements {#sec-process-end}
 
-Endorsements are matched with ACS entries by iterating through the `ev` list.
-For each `ev` entry, the `condition` ECT is compared with an ACS ECT, where the ACS ECT `cmtype` contains either `evidence`, `reference-values`, or `endorsements`.
-If the ECTs match ({{sec-match-condition-ect}}), the `ev` `addition` ECT is added to the ACS.
+Endorsed Value Triple and Conditional Endorsement Triple share the same internal representation.
+Both types of triple are transformed into an internal representation based on `ev`, but the CBOR encoding of Endorsed Value Triple can only represent a subset of the CBOR encoding of Conditional Endorsement Triple.
```
yogeshbdeshpande, line 2396:
```suggestion
Both types of triple are transformed into an internal representation based on `ev`, but the CBOR encoding of Endorsed Values Triple can only represent a subset of the CBOR encoding of Conditional Endorsement Triple.
```

draft-ietf-rats-corim.md
```diff
@@ -2371,14 +2392,42 @@ Endorsements are added to the ACS if the Endorsement condition is satisifed by t
...
+
+After successfully matching an `ev` condition ECT whose `instance` field is a `tagged-variable` against an ACS entry, the Verifier SHALL bind the `instance` value to the `instance` value of the ACS-ECT which matched against that condition ECT.
+If the matching ACS-ECT does not include an instance then the bound `instance` value is the empty value.
+
+An `ev` addition ECT whose `enviroment-map`.`instance` field is a `tagged-variable` references the bound value.
+When creating the ECT which will be added to the ACS, if the referenced `tagged-variable` is bound to a non-empty value, then the Verifier SHALL copy that bound value to the `enviroment-map`.`instance` field of the ACS-ECT created by the `ev` addition.
+If the referenced variable is bound to the empty value then the Verifier SHALL NOT add an `instance` field.
+If the referenced variable is not bound then the triple is invalid - the Verifier SHALL NOT add the addition ECT to the ACS.
+
+If a conditional endorsement containing `ev` conditions which use `tagged-variables` matches against multiple ECTs, then each match is processed independently of the others.
```
yogeshbdeshpande, line 2426: Such muddling is not helping, `addition` then again followed by `condition` again

draft-ietf-rats-corim.md
```diff
@@ -931,10 +933,29 @@ anonymity set.
...
+~~~ cddl
+{::include cddl/wc-match-type-choice.cddl}
+~~~
+
+An `environment-map` without an `instance` field used in a condition will match against an ACS-ECT with any `instance` field.
+When the ACS contains evidence from multiple Target Environments, and that evidence uses the `instance` field to indicate which Target Environment the evidence relates to, the `tagged-variable` values are used to indicate the relationships between `environment-map`s.
+These values behave in a similar way when used in the `group` field.
+
+In a condition, the `tagged-variable` is used to indicate that two or more `environment-map`s must have the same `instance` value.
+In an addition, the `tagged-variable` value is used to indicate that the `environment-map`.`instance` field must be copied from the ACS-ECT which matched a condition within the same triple.
```
andrew-draper, line 955:
```suggestion
In an addition, the `tagged-variable` value is used to indicate that the `environment-map`.`instance` field must be copied from the ACS-ECT which matched a condition within the same triple, to the ECT added as a result of the addition.
```

draft-ietf-rats-corim.md
```diff
@@ -2456,14 +2477,72 @@ Endorsements are added to the ACS if the Endorsement condition is satisifed by t
...
+This section describes an algorithm for processing an `ev` containing two or more conditions which include `tagged-variable`s.
+
+{:dmt-enum: counter="dmt1" style="format Step %d."}
+
+{: dmt-enum}
+* For each of the conditions, find the set of ACS-ECTs which match the fields within that condition which are not `tagged-variable`s.
+
+* Iterate through the set of AES-ECTs which match the first condition. There may be multiple matches with different `instance` values.
+
+Any `tagged-variable`s within the first condition will be bound to values from the matching `instance` AES-ECTs.
```
yogeshbdeshpande, line 2529:
```suggestion
Any `tagged-variable`s within the first condition will be bound to values from the matching `instance` ACS-ECTs.
```

draft-ietf-rats-corim.md
```diff
@@ -2456,14 +2477,72 @@ Endorsements are added to the ACS if the Endorsement condition is satisifed by t
...
+
+* If any of the conditions has no matches for any addition ECT then discard that partial set.
+
+* If any of the conditions has multiple matches for an ECT then create a partial set for each output.
+
+{: dmt-enum}
+* Create a result set for each set of AES-ECTs which match all conditions, add that set to the list of successful matches.
+
+* Continue processing the triple, processing it separately for each result set.
+For example, if the triple is a Conditional Endorsement triple which uses `tagged-variable` within the addition `instance` field then a separate AES-ECT will be added for each result set.
```
yogeshbdeshpande, line 2545:
```suggestion
[^tracked-at]  https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/458
```

yogeshbdeshpande:
@andrew-draper :  Given that we need sufficient work to explain the clear steps of Conditions and Exit for the Algorithm as in the issue #458 
How about we discuss some ideas in the coming CoRIM Meeting, this week and prepare a better proposal in the coming two weeks so that we can work on it at the Hackathon and bring it on the table for RATS session in Madrid ?

This gives us better time to work on an Alternative proposal and discuss face to face with other CoRIM Design team members.

Given your familiarity with the subject,  in the background please work on a Clear Algorithm and then what is the best type 
(other than instance ) to replace it, can be a simple search/replace in the PR to use a new code point example:  `measurement-values-map` to deal with and progress this issue!

henkbirkholz (commented):
Rechecking after Thomas' well laid out "stability" argument https://github.com/ietf-rats-wg/draft-ietf-rats-corim/pull/434#pullrequestreview-2993731336

andrew-draper:
If there were issues with this then I think it would have been good to raise them before the last minute, and it would be good to see the alternative proposal.

If there was an alternative proposal then it might make sense to delay putting in this change until we can decide between the alternatives. Since there isn't an alternative proposal I would prefer to do what Thomas suggested earlier, which is to check this in and be prepared to change to a better version if that is available.

If we don't accept this change then my take is that we don't have a complete solution to this issue, and we are not ready for last call.

PR #439, opened 2025-05-21
Create .ruby-version

nedmsmith:
Newer versions of ruby generate warnings.
Should the repo direct everyone to use version 3.2.2


thomas-fossati:
> May I ask that you add a Gemfile.lock to this so we can understand exactly what the dependencies are that are at least one maintainer can claim works to build the doc?

+1

I think it’s two packages: base64 and pstore that need to be added to the Gemfile.  See also a similar fix in msg-wrap: https://github.com/ietf-rats-wg/draft-ietf-rats-msg-wrap/pull/223/commits/2d03f03ca2986ad05b071e5a84b2c6d04eae1c6f

yogeshbdeshpande:
@nedmsmith Given Thomas's PR #445 has fixed the Ci do we still need this, shall we close this now?

nedmsmith:
> has fixed the Ci do we still need this

I was getting bad behavior even with Thomas's "fix". The resolution for me resulted in an addition to the .gitignore. I've updated the PR to reflect this difference. Since the repo has .gitignore files (not sure why) does it matter that this additional entry is in there? Normally stuff shows up in .gitignore added by various people and it just gets swept up with whatever other stuff is in the PR.
At the end of the day the question is will having "/.bundle" in the .gitignore file adversely affect anyone?

nedmsmith:
I did that and discarded the stash. That seems to have resolved the problem so it is OK to close this PR.

nedmsmith:
Not an issue.

PR #441, opened 2025-05-27
Rename `raw-int` to `int-range`

### LINKED ISSUE #419: Rename `raw-int`
Since `raw-int` is intended to support ranged comparsions (and whos value can be two integers, representing the bounds of the range), it does not behave as an ordinary "raw integer", which is what the name implies. 

Suggested replacement: `ranged-int`.

thomas-fossati:
Fix #419


thomas-fossati:
cc @jraman567

deeglaze (approved):

andrew-draper (approved):
This is a better name for the type

nedmsmith (approved):
LGTM

PR #445, opened 2025-06-06
Update Gemfile

thomas-fossati:
Fix broken CI.

Alternative to #439


deeglaze (approved):
Would be great to get a lock file.

Issue #465, opened 2025-07-18
Section 9.2.2.1:  Clarify validation of evidence that includes CoRIM fields or entire CoRIM


gmandyam:
Current text states "If a CoRIM profile is supplied ..." but does not clarify whether the profile is supplied as part of attestation evidence.  Given that this statement appears in a section entitled "Cryptographic Validation of Evidence", a reader may conclude that the profile is included in the evidence.

Recommend to change to "If a CoRIM profile is supplied by the attester either as part of evidence or in a separate conceptual message ...".

Would also recommend an additional statement at the end of the section:  "The verifier may also ignore a CoRIM profile provided by the attester if it has access to a separate profile.  In other words, an attester-provided CoRIM profile will not necessarily take the place of a CoRIM profile that has been separately provided to the verifier."

yogeshbdeshpande:
The introduction text:

If a CoRIM profile is supplied, it MUST describe:

Does not belong here...

thomas-fossati:
This paragraph is paired with the second paragraph of §8:

> _If a CoRIM profile is specified, there are a few well-defined points in the procedure where Verifier behaviour depends on the profile. The CoRIM profile MUST provide a description of the expected Verifier behavior for each of those well-defined points._

This important information needs to be made clear to the reader.  While this is perhaps not the best place, we cannot just remove it.

thomas-fossati:
As agreed during the 08/27 meeting: let’s keep #473 open and address the following points:
 
* create section 4.2 focused on profiles, specifically normative constraints on profile writers
* ensure that the processing clearly highlights the plug-ins for profile-specific code

PR #473, opened 2025-08-12
Remove extraneous text in Sec. 9.2.2.1

gmandyam:
Addressing https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/465


henkbirkholz (approved):
not blocking this deletion, but please highlight in case there is a really good reason why you still need this (relatively arbitrary text).

Issue #466, opened 2025-07-18
Section 4.1.1 Identity:  clarify requirements on uniqueness


gmandyam:
Text as currently written states that the identifier "uniquely identifies a CoRIM instance" but does not define what constitutes an "CoRIM instance".

Recommend clarifying:  "A unique CoRIM instance specifies a unique CoRIM map.  Two CoRIM instances cannot be considered equivalent unless they match bitwise.  If not, they must have different CoRIM IDs."

yogeshbdeshpande:
@gmandyam to add some clarifying text

PR #474, opened 2025-08-12
Clarification on uniqueness of the CoRIM ID

gmandyam:
https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/466

### DIFF
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -421,7 +421,8 @@ A `corim-map` is unsigned, and its tagged form is an entrypoint for parsing a Co
 
 ### Identity {#sec-corim-id}
 
-A CoRIM Identifier uniquely identifies a CoRIM instance.
+A CoRIM Identifier uniquely identifies a CoRIM instance within the context of a CoRIM issuer.  In other words the CoRIM identifier is not globally unique,
+but can be used to distinguish CoRIMs that come from the same issuer.
 The base CDDL definition allows UUID and text identifiers.
 Other types of identifiers could be defined as needed.
 


draft-ietf-rats-corim.md
```diff
@@ -421,7 +421,8 @@ A `corim-map` is unsigned, and its tagged form is an entrypoint for parsing a Co
 
 ### Identity {#sec-corim-id}
 
-A CoRIM Identifier uniquely identifies a CoRIM instance.
+A CoRIM Identifier uniquely identifies a CoRIM instance within the context of a CoRIM issuer.  In other words the CoRIM identifier is not globally unique,
```
steven-bellock, line 424: Is `CoRIM issuer` synonymous with `signer-name` and/or `kid` within the CoRIM?

draft-ietf-rats-corim.md
```diff
@@ -421,7 +421,8 @@ A `corim-map` is unsigned, and its tagged form is an entrypoint for parsing a Co
...
+but can be used to distinguish CoRIMs that come from the same issuer.
```
yogeshbdeshpande, line 425:
```suggestion
but it is used to distinguish CoRIMs that come from the same issuer.
```

draft-ietf-rats-corim.md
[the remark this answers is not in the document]
yogeshbdeshpande: The catch is that when CoRIM ID is using `uuid` construct it will be globally unique, hence I would say may not be, instead of an affirmative, is not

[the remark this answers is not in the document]
henkbirkholz: If we are going down this route, alternatively:

```suggestion
A CoRIM Identifier uniquely identifies a CoRIM instance within the context of a CoRIM issuer.
In essence, the CoRIM identifier is intended to be globally unique,
```

draft-ietf-rats-corim.md
```diff
@@ -421,7 +421,9 @@ A `corim-map` is unsigned, and its tagged form is an entrypoint for parsing a Co
 
 ### Identity {#sec-corim-id}
```
yogeshbdeshpande, line 422:
```suggestion
### CoRIM Identifier {#sec-corim-id}
```

Issue #477, opened 2025-08-13
Interoperability with SCITT Transparency Services


achamayou:
It would be really useful to allow COSE_Sign1-signed CoRIM/CoRIM collections to be made transparent on a SCITT Transparency Service (https://ietf-wg-scitt.github.io/draft-ietf-scitt-architecture/draft-ietf-scitt-architecture.html).

Since they are COSE_Sign1, that's not far off, but there are two blockers that I can see:

1. The lack of support for X.509 in the Signed CoRIM schema (x5c/x5t), as specified in the SCITT Signed Statement definition: https://ietf-wg-scitt.github.io/draft-ietf-scitt-architecture/draft-ietf-scitt-architecture.html#fig-signed-statement-cddl

Aside from technical compatibility, this is of substantial practical value to enable transition from signing services and infrastructure that still uses X.509.

2. The lack of support for CWT_Claims (https://datatracker.ietf.org/doc/rfc9597/) as a mechanism to capture signer/issuer information for the COSE_Sign1-signed CoRIM: https://ietf-wg-scitt.github.io/draft-ietf-scitt-architecture/draft-ietf-scitt-architecture.html#fig-signed-statement-cddl. It would be good to enable that either as an alternative, or even as mechanism that can be composed with `corim-meta` for COSE_Sign1.

This would enable COSE_Sign1-signed CoRIM/CoRIM collections to be registered with transparency services, using the same authentication and registration policies that are used for other Signed Statements.

achamayou:
As discussed today, item 1 overlaps with #467. The agreed resolution for is to update #475 to remove the mandatory `kid` parameter from the protected header, and explicitly defer to RFC9052 (COSE) and RFC9630 (carrying and referencing X.509 certificates in COSE) to enable `x5c`/`x5t`/...

On item 2, I will propose a PR before next week's meeting to allow either `CWT_Claims` () or `corim-meta` to be used for COSE_Sign1-signed CoRIM.

PR #484, opened 2025-08-19
Remove mandatory kid in schema

achamayou:
This addresses the first point in #477 by removing the mandatory `kid` (defined in [RFC9052](https://www.rfc-editor.org/rfc/rfc9052.html#name-common-cose-header-paramete)) in protected headers. Other key identification methods are available in COSE, such as `x5t/c/u`, defined in [RFC9360](https://www.rfc-editor.org/rfc/rfc9360.html#name-x509-cose-header-parameters).

I suggest it makes sense not specialize CoRIM in that respect, nor to limit its scope to signers using a specific subset of these mechanisms. If it helps, we could add a sentence along the lines of:

```
The cryptographic material to be used by the verifier can be identified through any method supported by COSE_Sign1, such as `kid` defined in RFC9052 or x5t/x5c/x5u as defined in RFC360 for example.
```

### DIFF
--- a/cddl/protected-corim-header-map.cddl
+++ b/cddl/protected-corim-header-map.cddl
@@ -1,7 +1,6 @@
 protected-corim-header-map = {
   &(alg: 1) => int
   &(content-type: 3) => "application/rim+cbor"
-  &(kid: 4) => bstr
   &(corim-meta: 8) => bstr .cbor corim-meta-map
   * cose-label => cose-value
 }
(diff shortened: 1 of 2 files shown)


thomas-fossati (approved):
🚢 it!

henkbirkholz (approved):
okay with making kid optional and reduce cddl complexity. does not harm interop.

achamayou:
> okay with making kid optional and reduce cddl complexity. does not harm interop.

I claim it improves interoperability by reducing the set of possible COSE_Sign1 envelopes containing CoRIM that are invalid Signed CoRIM, in particular those using `x5*` for key identification. Those do not typically set a `kid`, since it would serve no purpose and since COSE Sign1 itself does not mandate it.

PR #496, opened 2025-10-02
Change Example Verifier to Reference Verifier

### LINKED ISSUE #406: Normative statements in section 9
For the first time I realized that the entirety of section 9 is titled (emphasis mine) "**Example** Verifier Algorithm" and states
>This document assumes that Verifier implementations may differ. To facilitate the description of normative Verifier behavior, this document describes the internal representation for an example Verifier and demonstrates how the data is used in the appraisal phases outlined in [Section 8.1](https://ietf-rats-wg.github.io/draft-ietf-rats-corim/draft-ietf-rats-corim.html#sec-appraisal-procedure).

Normative text should not be used in an example, however section 9 has many `MUST`s. Am I correct in thinking that
>To facilitate the description of normative Verifier behavior

means that section 9 is a non-normative guide to another to-be-written specification that will specify normative Verifier behavior?

steven-bellock:
Fix #406.


yogeshbdeshpande (approved):
LGTM!

PR #516, opened 2025-11-26
Remove duplications

yogeshbdeshpande:


thomas-fossati:
It looks like the second half of this PR has become obsolete because of the revamp of §§8 and 9.

The first half, which removes the "Verifier Reconciliation" section and replaces it with an overview of the document structure, may still be salvageable.

Yogesh, what are your plans for this?

thomas-fossati:
OBE (see #598)

PR #535, opened 2026-02-04
Update Veraison status

yogeshbdeshpande:
Update Veraison status to also include rust based implementation


yogeshbdeshpande:
thanks @thomas-fossati andy was non -commital about it, will check veraison/cover

Issue #575, opened 2026-04-15
Clarity on multiplicity of ref-claims


sgilson:
Hi folks,

I'm attempting to follow the reference verification algorithm and am stumbling on the statement that "ref-claims in a reference-triple-record can contain one or more entries. This multiplicity can have different meanings" from section 5.1.5. I've gone through https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/442, https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/523, and https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/571 (plus some of the associated PRs) and am unable to form a consistent mental model of this multiplicity. I'm likely missing a key concept, so any help in clearing up my confusion is appreciated. Referenced excerpts with additional emphasis are included at the end.

This comment best illustrates the problem I see: https://github.com/ietf-rats-wg/draft-ietf-rats-corim/pull/500#pullrequestreview-3316755691. It's stated that:

```
T = [
  E=e_0,
  [
    M{ K=k_0, V=v_0 },
    M{ K=k_0, V=v_1 }
  ]
]
```
is shorthand for:
```
T_1 = [
  E=e_0,
  [
    M{ K=k_0, V=v_0 }
  ]
]
+
T_2 = [
  E=e_0,  # Was e in original comment. Based on the context, I believe e_0 was intended.
  [
    M{ K=k_0, V=v_1 }
  ]
]
```

Section 5.1.5 seems consistent with this, however, it's not clear to me how this aligns with the reference verification algorithm.

Let's say the evidence is:

```
ET = [
  E=e_0,
  [
    M{ K=k_0, V=v_0 }
  ]
]
```

Also assume v_0 and v_1 make incompatible assertions about the same property, say `name` (e.g. v_0 matches when name=foo, v_1 matches when name=bar). The CoRIM author intends to say "both foo and bar are valid for name under mkey=k_0 and environment e_0".

We then produce the rv lists corresponding to each reference value form given above. Authority, cmtype, etc omitted for brevity:

```
# from T
rv_1 = [
  {
    condition: { 
      environment: e_0,
      element-list: [ 
        {element-id: k_0, element-claims: v_0 },
        {element-id: k_0, element-claims: v_1 },
      ]
    },
    addition: { environment: e_0 }
  }
]

# from T1 + T2
rv_2 = [
  {
    condition: { 
      environment: e_0,
      element-list: [ 
[...]

PR #576, opened 2026-04-17
Reference verifier section(s) revamp

thomas-fossati:
This is the revamped Section 8 that I presented at our last weekly meeting.

For the sake of your mental health, I strongly advise agianst looking at the raw markdown diff.
Instead, read the entire [new Section 8](https://ietf-rats-wg.github.io/draft-ietf-rats-corim/ect-restruct/draft-ietf-rats-corim.html#section-8) alongside the [rendering diff](https://author-tools.ietf.org/api/iddiff?url_2=https://ietf-rats-wg.github.io/draft-ietf-rats-corim/ect-restruct/draft-ietf-rats-corim.txt&url_1=https://ietf-rats-wg.github.io/draft-ietf-rats-corim/draft-ietf-rats-corim.txt) kindly provided by rfcdiff.

There are still a few `[TBC]`s and `[TODO]`s that we can tackle iteratively.

NOTE 1: I added [sample pseudocode](https://ietf-rats-wg.github.io/draft-ietf-rats-corim/ect-restruct/draft-ietf-rats-corim.html#figure-44) to the “SVN comparison” section.  We can add similar code to the other comparisons or remove it if we think that the prose is sufficiently clear. 

NOTE 2: I have added a realistic, complete [appraisal of a PSA attester](https://ietf-rats-wg.github.io/draft-ietf-rats-corim/ect-restruct/draft-ietf-rats-corim.html#section-8.2.6) based on the reference verifier algorithm and the CoRIM profile for PSA endorsements, defined in [draft-fdb-rats-psa-endorsements](https://datatracker.ietf.org/doc/draft-fdb-rats-psa-endorsements/). It currently sits at the bottom of §8, but it could be moved to an appendix if desired.


draft-ietf-rats-corim.md
```diff
@@ -1810,1184 +1821,1174 @@ When used as an identifier the responsible allocator entity SHOULD ensure unique
...
-DDGs need not be isomorphic to domain membership graphs.
-* Verify the DDG is acyclic.
+A key that successfully passes the above checks is said to be verified.
 
-If, in a later processing phase, an appraisal policy for trust dependency exists, the DDG can be further evaluated.
-For example, a trust dependency policy might specify a strength of function requirement for how Evidence about a TE is integrity protected by its AE.
+Add all the verified keys to the addition ECT's `key-list` and add the addition ECT to the ACS.
 
-Domain Dependency ECTs are processed using the following algorithm:
+##### Processing `dm` Relations {#sec-proc-dm}
```
yogeshbdeshpande, line 2718: Processing `dm` relations, is a double loop, where if `any` Domain Member ECT is added into the ACS, then The Outer Loop is re-iterated.

This core concept is missing in the new PR, which needs addition.

yogeshbdeshpande (changes requested):
Started reviewing, from today..,.

draft-ietf-rats-corim.md
```diff
@@ -1810,1184 +1821,1174 @@ When used as an identifier the responsible allocator entity SHOULD ensure unique
...
+It also assumes that the `dm-item`s in the `dm` relation are topologically sorted.
+This allows the algorithm to execute in one pass.
 
-* Check that the ACS.`cmtype` contains `domain-member`.
-* Check that *t* matches an ACS.`environment`.
+For each `dm` entry, the condition ECT is compared with either an ACS Element ECT with `cmtype` 0 (i.e., corroborated evidence) or a Domain ECT with `kind` 0 (i.e., member).
+<cref>
+[TBC]
+The original text says: "Domain Membership ECTs (i.e., cmtype equals domain-member) in the dm staging area are matched with ACS entries where cmtype is set to evidence, reference-values i.e. corroborated evidence, or domain-member [...]" but it seems that matching evidence ECTs is superfluous and potentially leading to the wrong result (i.e., matching uncorroborated environments).
+</cref>
```
```suggestion
For each `dm` entry, the condition ECT is compared with either an ACS Element ECT with `cmtype` 2 (i.e., evidence) or a Domain ECT with `kind` 0 (i.e., member).
```

draft-ietf-rats-corim.md
```diff
@@ -1810,1184 +1821,1174 @@ When used as an identifier the responsible allocator entity SHOULD ensure unique
...
+Domain Membership relations describe the expected topological arrangement of the Attester.
 
-* Check that the ACS.`cmtype` contains `domain-member`.
-* Check that the `dde`.`environment` matches a domain member ACS entry `environment`.
-* OR that the `dde`.`environment` matches one of it's ACS.`members`.`environment`.
+Domains are matched with ACS entries by iterating through the `dm` list.
 
-For each trustee *t* in `dde`.`trustees` (inner loop):
+The following algorithm assumes that the graph described by the condition ECTs in the `dm` relation is acyclic.
+It also assumes that the `dm-item`s in the `dm` relation are topologically sorted.
```
```suggestion
It also assumes that the `dm-item`s in the `dm` relation are topologically sorted (bottom up, from leaves to root).
```

draft-ietf-rats-corim.md
```diff
@@ -1810,1184 +1821,1170 @@ When used as an identifier the responsible allocator entity SHOULD ensure unique
...
 
-The Verifier MUST ensure that a triple including a matching condition is processed after any other triple that modifies or adds an ACS entry with an `environment-map` that is in the matching condition.
+* `profile`: The profile that defines the domain of interpretation of this tuple.
+This is the `profile` attribute of the CoRIM that contained the original triple from which this ECT was obtained.
+This is an optional attribute in an ECT.
+If no profile is used, the attribute is omitted.
+<cref>
+[TBC]
+What profile applies in such a case?
+</cref>
```
```suggestion
If no profile is used, the attribute is omitted, and the only comparison rules that apply are those specified in {{sec-comparison-rules}}.
```

draft-ietf-rats-corim.md
```diff
@@ -1810,1184 +1821,1170 @@ When used as an identifier the responsible allocator entity SHOULD ensure unique
...
+If two Element ECTs have the same `environment`, `cmtype`, `authority` and `profile` then their `element-list`s are merged.
+Any duplicates MUST be pruned.
 
-Regardless of the specific integrity protection method used, the Verifier MUST NOT process Evidence which is not successfully validated.
+<cref>
+[TBC]
+Original text also states: "If two merged `measurement-values-map` contains duplicate codepoints and the measurement values are not equivalent, then the Verifier SHALL report an error and stop validation processing."
+Are we sure this is the case?
+Can't they be interpreted as alternative states?
+</cref>
```
```suggestion
Two `measurement-values-map` containing duplicate codepoints and with non-equivalent measurement values MUST NOT be merged.
These are effectively two different acceptable states that need to be processed separately.
```

draft-ietf-rats-corim.md
```diff
@@ -1810,1184 +1821,1170 @@ When used as an identifier the responsible allocator entity SHOULD ensure unique
...
-* **copy**(`ddtr`.`profile`, `dde`.`profile`)
+Endorsed Values and Conditional Endorsed Values are matched with ACS entries by iterating through the `ev` list.
+For each `ev` entry, the condition ECT is compared with an ACS ECT with `cmtype` 0, 1 or 2 (i.e., reference-values, endorsements or evidence).
+If the two match, the addition ECT is added to the ACS.
 
-Append the domain dependency edge (`dde`) to the domain dependency graph (`ddg`).
+<cref>
+[TBC]
+Original text also states: "Some condition values can match against multiple ACS-ECTs, or sets of ACS-ECTs. If there are multiple matches, then each match is processed independently from the others."  Why should add exact copies of the same addition ECT?  Is there any specific semantics associated with clones?
+</cref>
```
```suggestion
```

draft-ietf-rats-corim.md
```diff
@@ -1810,1184 +1821,1170 @@ When used as an identifier the responsible allocator entity SHOULD ensure unique
...
+If, after visiting all the entries in the ACS, the matched entries set is not empty, the C-ECT matches the ACS.
+Conversely, if the matched entries set is empty, the C-ECT does not match the ACS.
 
-The Verifier SHALL perform all of the comparisons defined in {{sec-compare-environment}}, {{sec-compare-authority}}, and {{sec-compare-element-list}}.
+If the C-ECT contains a profile and the profile defines an algorithm for a given codepoint, the processor MUST use the algorithm defined by the profile in comparisons involving that codepoint.
+If the condition ECT contains a profile, but the profile does not define an algorithm for a particular codepoint, the processor MUST use the standard algorithm described in this document for comparisons involving that codepoint.
 
-Each of these comparisons compares one field in the condition ECT against the same field in the ACS entry.
+The specific comparisons performed depend on the type of relation being processed.
+In general, the processor will perform comparisons based on `environment` (see {{sec-compare-environment}} and `authority` (see {{sec-compare-authority}}), as well as more specialized comparisons based on the type of ECT matched.
```
```suggestion
In general, the processor will perform comparisons based on `environment` (see {{sec-compare-environment}}) as well as more specialized comparisons based on the type of ECT matched.
```

draft-ietf-rats-corim.md
```diff
@@ -1810,1184 +1821,1158 @@ When used as an identifier the responsible allocator entity SHOULD ensure unique
...
-Each of these comparisons compares one field in the condition ECT against the same field in the ACS entry.
+The specific comparisons performed depend on the type of relation being processed.
+In general, the processor will perform comparisons based on `environment` (see {{sec-compare-environment}}) as well as more specialized comparisons based on the type of ECT matched.
+Element ECTs will match on `element-list` (see {{sec-compare-element-list}}), Key ECTs will match on `key-list` (<cref>[TODO]</cref>), and Domain ECTs will typically match on `children` (see {{sec-compare-environment}}).
 
-If all of the fields match, then the condition ECT matches the ACS entry.
+<cref>
+[TBC]
+Are authority checks always carried out?
+</cref>
```
thomas-fossati, line 2776:
```suggestion
<cref>
[TBC]
Check, for each relation processing, where authority comparisons apply.
</cref>
```

draft-ietf-rats-corim.md
```diff
@@ -1810,1184 +1821,1158 @@ When used as an identifier the responsible allocator entity SHOULD ensure unique
...
 
-If any codepoint present in the condition ECT `measurement-values-map` does not match the same codepoint within the candidate entry `measurement-values-map` then Verifier SHALL remove that candidate entry from the candidate entries array.
+If the processor is unable to determine the applicable comparison algorithm for an attribute, it MUST behave as though the C-ECT does not match the ACS-ECT.
 
-#### Comparison of a single measurement-values-map codepoint {#sec-match-one-codepoint}
+Profile writers SHOULD use CBOR tags for widely applicable comparison methods to ease Verifier implementation compliance across profiles.
+<cref>
+[TBC]
+Unclear what this recommendation aims to achieve.
+</cref>
```
```suggestion
```

draft-ietf-rats-corim.md
```diff
@@ -1810,1184 +1821,1153 @@ When used as an identifier the responsible allocator entity SHOULD ensure unique
...
 
-As a second example, in PSA {{-psa-token}} the verification public key is looked up in the appraisal context using the `ueid` claim found in the PSA claims-set.
-If found, COSE Sign1 verification is performed accordingly.
+**Merge Rules.**
+If two Element ECTs have the same `environment`, `cmtype`, `authority` and `profile` then their `element-list`s are merged.
+Any duplicates MUST be pruned.
 
-Regardless of the specific integrity protection method used, the Verifier MUST NOT process Evidence which is not successfully validated.
+Two `measurement-values-map` containing duplicate codepoints and with non-equivalent measurement values MUST NOT be merged.
+These are effectively two different acceptable states that need to be processed separately.
```
sgilson: Would `element-map` be more accurate than `measurement-values-map` in this context?

thomas-fossati: Yes, definitely.  Thanks for spotting it.

draft-ietf-rats-corim.md
thomas-fossati: The comment can be resolved as we clarified this during the weekly call.  Yogesh: Feel free to unresolve if you disagree.

draft-ietf-rats-corim.md
thomas-fossati: This has been addressed in 3fdfcf7.  However, it would probably be worth calling out `element-id` as one of the matching criteria.

draft-ietf-rats-corim.md
```diff
@@ -1810,1184 +1821,1153 @@ When used as an identifier the responsible allocator entity SHOULD ensure unique
...
 
-If every entry in the condition ECT `authority` matches byte-by-byte a corresponding entry in the ACS entry `authority` field, then the authorities match.
-The order of the fields in each `authority` field do not affect the result of the comparison.
+##### Element List Comparison {#sec-compare-element-list}
 
-If any entry in the condition ECT `authority` does not have a matching entry in the ACS entry `authority` field then the authorities do not match.
+A C-ECT's `element-list` matches an ACS-ETC's `element-list` if all the `element-map`s in the C-ECT's `element-list` match the `element-map`s in the ACS-ECT's `element-list`.
 
-When comparing two `$crypto-key-type-choice` fields for equality, the Verifier SHALL treat them as equal if their deterministic CBOR encoding is binary equal.
+Any `element-map` that is present in the ACS-ECT's `element-list` but not in the C-ECT's is ignored in the comparison.
```
sgilson, line 2805: I've been looking at CoRIM modeling with others and modeling multiple accepter attester states is a sticking point (i.e.  modeling ref values such that SPDM index 1 can have digest value A or B). Can you spot my mistake in the following logic?

After ECT merging is complete, there will be exactly one ECT with a given environment, cmtype, authority, and profile. The element-list will effectively contain a set of all unique element-maps. Borrowing some of your notation from another MR, this is means that the below forms are equivalent:
```
T = [
  E=e_0,
  T=reference-value, # added the cmtype as "T"
  [
    M{ K=k_0, V=v_0 },
    M{ K=k_0, V=v_1 }
  ]
]

T_1 = [
  E=e_0,
  T=reference-value,
  [
    M{ K=k_0, V=v_0 }
  ]
]
+
T_2 = [
  E=e_0,  # Was e in original comment. Based on the context, I believe e_0 was intended.
  T=reference=value,
  [
    M{ K=k_0, V=v_1 }
  ]
]
```
So, due to merging, a CoRIM author can provide either T or T1 + T2. From now on, we'll work with the merged form T.


From here, it would seem that the final form T doesn't support the case where the author intended to represent multiple acceptable states. That is:
```
# This can be matched:
T = [
  E=e_0,
  T=evidence,
  [
    M{ K=k_0, V=v_0 },
    M{ K=k_0, V=v_1 }
  ]
]

# This is should be matched, but seemingly would not be:
T = [
  E=e_0,
  T=evidence,
  [
    M{ K=k_0, V=v_0 }
  ]
]
```
This is due to the statement that **all** element-maps in the C-ECT must match the element-maps in the ACS-ECT. `M{ K=k_0, V=v_1 }` would not match because it is (rightfully) missing from the evidence.

I may be missing a key concept here. For example, could it be the case that T1 and T2 were actually not merged because one of their element-maps were not the same? That could be a solution to my dilemma, but I'm not sure what other issues that might cause.

draft-ietf-rats-corim.md
```diff
@@ -2727,14 +2727,10 @@ Domain Membership relations describe the expected topological arrangement of the
 Domains are matched with ACS entries by iterating through the `dm` list.
 
 The following algorithm assumes that the graph described by the condition ECTs in the `dm` relation is acyclic.
-It also assumes that the `dm-item`s in the `dm` relation are topologically sorted.
+It also assumes that the `dm-item`s in the `dm` relation are topologically sorted (bottom up, from leaves to root).
```
yogeshbdeshpande, line 2725: Please also add a reference to a clean topological sorting algorithm which can be referenced here. So far our internal algorithms were solely based on the logic documented in this draft. Now we rely on foreign algorithms to do correct job, before this draft algorithms can work properly.

yogeshbdeshpande (commented):
looking at it now!

draft-ietf-rats-corim.md
thomas-fossati: I believe we have a bug in the spec.

The idea should be that alternative states must be encoded in separate triples, and that merging rules must disallow coalescing triples that represent alternative states.

Otherwise, as you have also noticed, the matching algorithm does not work because the `rv-item.condition` becomes self-contradictory.

I think that the sentence in §5.1.5:

> Each ref-claims entry can represent a different possible state of the Environment

needs to go.

draft-ietf-rats-corim.md
henkbirkholz: afaik, there is no need to pursue this change any further. @yogeshbdeshpande, could you resolve it?

draft-ietf-rats-corim.md
henkbirkholz: Now captured in this issue: https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/575

draft-ietf-rats-corim.md
henkbirkholz: Addressed

henkbirkholz (approved):
all items are addressed. When Yogesh resolves his open suggested change and there a a few more ticks, we can merge

nedmsmith:
I pushed updates to ect-restruct-updated that focus on domain dependency and membership. However, there are a number of problems with how the pseudo code has been implemented and subsequently the transformation and processing algorithms are incorrect. There is also an issue with word choice in the revised D-ECT structures as well as an error. 

Membership:
Primary issue for domain membership is it isn't possible to add a new "container" domain given a set of existing domains / environments. The logic is if environments A and B exist then a container domain C describes A and B as members of C. 

The current processing only allows C to be added if it is a child of A or B. 

The allgorithm should be:
IF `domain-id` is found in the ACS then assert `members` as ECTs into the ACS.
IF `members` are found in the ACS the assert `domain-id` as an ECT into the ACS.

Dependency:
The D-ECT excluded use of domain-id as a search criterion in the condition. This is necessary for domain dependency as the semantics of dd are that the "children" are trustees of the `domain-id` if BOTH `domain-id` and `trustees` are already a member of ANY domain. This needs to be represented in the condition. The addition only adds trustees so the addition doesn't need `environment`. The current intrep has this backwards.


The choice of "children" is unfortunate since member and trust relationships don't have to be hierarchical only acyclic. "peers" would have been a better choice.

draft-ietf-rats-corim.md
```diff
@@ -1810,1207 +1821,1312 @@ When used as an identifier the responsible allocator entity SHOULD ensure unique
...
+The following algorithm assumes that the graph described by the condition ECTs in the `dm` relation is acyclic.
+It also assumes that the `dm-item`s in the `dm` relation are topologically sorted (bottom up, from leaves to root).
+This allows the algorithm to execute in one pass.
 
-## Domain Dependency Triples {#processing-domain-dependency}
+For each `dm` entry, the condition ECT is compared with either an ACS Element ECT with `cmtype` 2 (i.e., evidence) or a Domain ECT with `kind` 0 (i.e., member).
+All other ECTs are ignored.
 
-### Internal Representation of Domain Dependency {#sec-ir-dd}
+If all the `children` environments in the condition ECT have a matching ECT in the ACS, the ECT addition is added to the ACS.
```
yogeshbdeshpande, line 2731:
```suggestion
If all the `children` environments in the condition ECT of the `dm entry` match a corresponding `environment` in an ACS ECT, then the addition ECT of the `dm entry` is added to the ACS.
```

yogeshbdeshpande (commented):
More clarity of comparisons

nedmsmith:
Following up on today's meeting. The DDT logic can be sumarized from one line in section 5: "A DDT specifies which component (identified by `domain-id`) depends on which other components (identified by `trustees`) for proper operation."

The condition isn't expressed in terms of a triple subject/object relationship. It is whether or not the DDT subject/object are all represented in the membership graph. 

There is an issue with conflating the membership ECT with the trust dependency ECT in that the condition logic for membership omits `environment` because it wants to enforce a bottom-up construction of a membership graph. 

The trust graph is not intended to be isomorphic to the membership graph. It does say it should be acyclic. From the perspective of a single triple, that just means the domain-id should not be in the trustee list. It doesn't try to specify an algorithm for cycle detection.

The problem with overloading the member and trust dep ECTs is they have different goals and semantics. In one case `environment` condition is not needed and the other case it is needed. By conflating them, a new value `kind` was created that allows if-then logic that deconflates them. But if the structures were fully orthogonal there is no need for a `kind` value. Right now the logic is:
IF kind=member then condition.environment is empty
ELSE IF kind=dependency then condition.environment is populated.

My proposal is to define another ECT specialization M-ECT for member ECT and remove D-ECT to be trust dependency specific aka T-ECT.

This will remove `kind` entirely and `chileren` can be avoided using the wording that most closely resonates (i.e., `members` for M-ECT and `trustees` for T-ECT). This will make documenting things more straight forward - which is the overall objective.

The overall changes would be:
```
ECT = E-ECT / T-ECT / K-ECT / M-ECT

// Membership 
M-ECT = {
  ECT-common
  members: [ + environment-map ]
}

member-condition-ect = ({
  authority: [ + $crypto-key-type-choice ]
  members: [ + environment-map ]
}) .within M-ECT

member-addition-ect = ({
  environment: environment-map
  authority: [ + $crypto-key-type-choice ]
  members: [ + environment-map ]
}) .within M-ECT

dm = [ + member-item ]

member-item = {
  condition: member-condition-ECT
  addition: member-addition-ECT
}

// Trust Dependency
T-ECT = {
  ECT-common
  trustees: [ + environment-map ]
}

dd = [ + trust-dep-item ]

trust-dep-item = {
  condition: T-ECT
  addition: T-ECT
}
```

`kind` is eliminated (implicit in CDDL structure naming). 
No need for conditional logic on `kind` to switch on optionality of `environment`.
The intuitive `members` and `trustees` can be used directly without introducing `children` which is biased toward hierarchies vs. graphs. 
Graph isomorphism between `members` and `trustees` is more intuitively obvious. 
A cycle detection / prevention algorthm can be a class built-in function. e.g.:
```
dd::MAKE-ACYCLIC()
```
The revised pseudocode for dd processing would be:
```
FUNC append_acs( dd: dd) -> ACS {
    dd::MAKE-ACYCLIC()
    FOREACH item IN dd:
        IF acs::IS-MEMBER(item.condition.environment)
            FOREACH trustee IN item.condition.trustees:
                IF acs::IS-MEMBER(trustee.environment):
                        CONTINUE
                ELSE GET NEXT item
            acs::APPEND(item.addition)
    RETURN acs
}
```
The pseudo code tests that both the environment and list of trustees are in the set of M-ECTs in the ACS. If yes, then add the trustor-trustee relationships. The MAKE-ACYCLIC function ensures the trustor-trustee relationships are secure before adding them to the ACS. There could be lots of impmlementation specific approaches that enforce security - but it is out of scope for the corim spec.
Note: A set of disjointed dd expressions is still acyclic.

Note: Section 5 can be reworded to be less presumptive toward hierarchies as the graph structure.

nedmsmith (changes requested):
I am suggesting changes to the internal representation to decouple membership ECT from trust dependency ECT.

yogeshbdeshpande:
> conditional logic on `kind` to switch on optionality of `environment`

@nedmsmith : After reviewing your changes, it all makes sense to me.

Can you please append the `ect-restruct-updated`  branch with these proposed changes so that we can create a new PR out of ect-restruct branch ...?

draft-ietf-rats-corim.md
```diff
@@ -1810,1207 +1821,1312 @@ When used as an identifier the responsible allocator entity SHOULD ensure unique
...
+{{fig-verifier-internal}} illustrates how the CoRIM processor fits into the wider RATS Verifier architecture.
 
 ~~~ aasvg
 {::include pics/corim-verifier.txt}
 ~~~
 {: #fig-verifier-internal artwork-align="center" title="CoRIM Processing Flow"}
 
-### Verifier Processing
+The CoRIM processor accepts Reference Values and Endorsements in the form of CoRIM documents, as well as Evidence that has been converted into a CoRIM-compatible format using transforms such as those described in {{-rats-evidence-trans}}.
+Before the appraisal can begin, all Conceptual Messages must be broken down and reshaped into a common internal representation.
```
yogeshbdeshpande, line 1901:
```suggestion
Before the appraisal can begin, all  CoRIMs must be broken down and reshaped into a common internal representation.
```

draft-ietf-rats-corim.md
```diff
@@ -1810,1207 +1821,1312 @@ When used as an identifier the responsible allocator entity SHOULD ensure unique
...
+The internal representations of Reference Values and Endorsements are stored in a staging area prior to appraisal initiation.
```
yogeshbdeshpande, line 1902: How about Attest Key Triples and Device Identity Triples & Domain Membership and Domain Dependency Triples.

draft-ietf-rats-corim.md
```diff
@@ -1810,1207 +1821,1312 @@ When used as an identifier the responsible allocator entity SHOULD ensure unique
...
+Instead, the internal representation of Evidence is used to initialize the ACS.
```
yogeshbdeshpande, line 1903:
```suggestion
The Appraisal process is initiated upon processing Evidence and transforming it into an internal representation. The ACS is then initialized  using the internal representation of Evidence.
```

draft-ietf-rats-corim.md
```diff
@@ -1810,1207 +1821,1312 @@ When used as an identifier the responsible allocator entity SHOULD ensure unique
...
+This section describes the data structures used by the CoRIM processor.
 
-See section {{sec-conc-mess}} for how Conceptual Messages are transformed from external representations to internal representations.
+#### ECT {#sec-ect}
 
-This section describes the internal structure of ECT and Appraisal Claims Set (ACS). These structures are used in the Appraisal phases
-described in detail in the following section.
+The Environment-Claim Tuple is a core internal construct of the CoRIM Verifier.
+It is used to describe a feature (or "Claim") of the appraised environment alongside relevant metadata.
+All ECTs, except those containing Evidence Claims, are typically obtained from CoMID triples.
```
yogeshbdeshpande, line 1922: Is this true, how about the CoTLs ??

draft-ietf-rats-corim.md
```diff
@@ -1810,1207 +1821,1312 @@ When used as an identifier the responsible allocator entity SHOULD ensure unique
...
+~~~ cddl
+{::include cddl/intrep-ect-refval-addition.cddl}
+~~~
+{: #fig-rv-ect-add title="Profiled ECT for Reference Values (addition)"}
 
-### Evidence Transformation
+As this is used to corroborate an Evidence ECT, its layout is identical to that of an `Evidence-addition-ECT`.
+The only differences are the values of the `element-list`, `authority` and `cmtype` attributes.
+Here, the `authority` value is that of the RVP rather than the Attester, and the `cmtype` value is `reference-values` rather than `evidence`.
+Furthermore, the `element-list` is created if a match is found during processing, rather than when the triple is transformed into the internal representation.
```
yogeshbdeshpande, line 2137: I would re-phrase it slightly better,

Furthermore the `element-list` in the addition-ECT entry is empty when the triple is transformed into its internal representation. Only during appraisal processing, if the conditions are satisfied then the element-list is created. See section (XXX Corroboration of Reference Values)

draft-ietf-rats-corim.md
```diff
@@ -1810,1207 +1821,1312 @@ When used as an identifier the responsible allocator entity SHOULD ensure unique
...
-{::include cddl/intrep-rv.cddl}
+{::include cddl/intrep-evs.cddl}
 ~~~
+{: #fig-evs title="Endorsed Values Series Relation"}
 
-The `rv` relation is a list of condition-addition pairings where each pairing is evaluated together.
-If the `condition` containing reference ECTs matches Evidence ECTs then the Evidence ECTs are re-asserted, but with RVP authority as contained in the `addition` and `cmtype` set to `reference-values`.
+The `evs` relation compares the condition ECTs with the ACS.
+<cref>
+[TBC]
```
yogeshbdeshpande, line 2165: Clarify Line 2165 during the review process

draft-ietf-rats-corim.md
```diff
@@ -1810,1207 +1821,1312 @@ When used as an identifier the responsible allocator entity SHOULD ensure unique
...
 
-### Corroboration of Reference Values {#sec-ref-corrob}
-Reference Values are matched with ACS entries by iterating through the `rv` list.
+~~~ cddl
+{::include cddl/intrep-domain-dep.cddl}
+~~~
+{: #fig-dd title="Domain Dependency Relation"}
 
-For each `rv` entry, the `condition` ECT is compared with an ACS ECT, where the ACS ECT `cmtype` contains `evidence`.
+{{fig-domain-ect-cond}} shows the profiled Domain ECT for both Domain Membership and Dependency conditions.
```
yogeshbdeshpande, line 2233: We should be careful in call Domain Condition as `ECT`, as strictly speaking it is NOT an ECT as it does not contain an Environment - but ONLY children a.ka. members.

yogeshbdeshpande (commented):
some more comments

yogeshbdeshpande:
@thomas-fossati A general observation reading the PR: Now that we have split the internal representation of Key Related Triples (they are no longer represented as EV Items), we should carefully proof read the document, wherever we mention processing of Reference Values and Endorsements to either state explicitly `including Key Triples` or make sure we have suitable clarifications to resolve the ambiguity!!!

thomas-fossati:
Picking a few points:

> The trust graph is not intended to be isomorphic to the membership graph.

Yes, but it needs to be a sub-graph.

> It does say it should be acyclic. From the perspective of a single triple, that just means the domain-id should not be in the trustee list.

Yes, but that’s not enough: the cycle detection needs to span all the DDT triples.

> The problem with overloading the member and trust dep ECTs is they have different goals and semantics. In one case `environment` condition is not needed and the other case it is needed. By conflating them, a new value `kind` was created that allows if-then logic that deconflates them. But if the structures were fully orthogonal there is no need for a `kind` value. Right now the logic is: IF kind=member then condition.environment is empty ELSE IF kind=dependency then condition.environment is populated.
> My proposal is to define another ECT specialization M-ECT for member ECT and remove D-ECT to be trust dependency specific aka T-ECT.

Going back to first principles, the idea was to extract the common structure, which DDTs and DMTs share for the most part.  But, as you say, the underlying triples have different semantics (the same goes for E-ECTs and K-ECTs, by the way), and therefore there needs to be an "internal switch" to drive the different processing logics.

Splitting them into separate data items would go counter to this reasoning and create a precedent for splitting all the E-ECTs and K-ECTs, too.

Instead, I believe this must be addressed in the prose, which I have carried over rather sloppily — apologies! —, rather than adding more CDDL or pseudo-code.

> […[
> ```
> dd::MAKE-ACYCLIC()
> ```

I’d take a step further and create a built-in `SORT-DEPENDENCIES` (or similar), which operates on different data types (D-ECTs as well as conditional endorsements) and also performs cycle detection.

yogeshbdeshpande (approved):
LGTM

nedmsmith (approved):
LGTM

PR #577, opened 2026-05-13
Fix typo in locator map CDDL

setrofim:
Fix definition of the thumbprint field inside corim-locator-map to indicate multiplicity inside the array variant of the field.


yogeshbdeshpande (approved):
LGTM!

PR #579, opened 2026-05-14
Clarify Reference Triples Encoding

### LINKED ISSUE #575: Clarity on multiplicity of ref-claims

yogeshbdeshpande:
Fixes #575


draft-ietf-rats-corim.md
```diff
@@ -1355,19 +1355,14 @@ The `reference-triple-record` has the following parameters:
...
 * `ref-claims`: Contains one or more reference measurements for the Target Environment
 
-CoMID triples ({{sec-comid-triples}}) may contain multiple `reference-triple-record` entries, each of which describes one or more possible states for a particular Target Environment.
+CoMID triples may contain multiple `reference-triple-record` entries.
+Each `reference-triple-record` describes one possible state for a particular Target Environment identified by `ref-env`.
+In other words, different reference states for a Target Environment MUST be expressed using separate `reference-triple-record` entries.
 
 The `ref-claims` in a `reference-triple-record` can contain one or more entries.
-This multiplicity can have different meanings:
+Each `ref-claims` entry represents the state of a different measured element within the Target Environment.
```
```suggestion
Each `ref-claims` entry represents the expected state of a different measured element within the expected overall state of the Target Environment. 
```

thomas-fossati (approved):

nedmsmith (approved):
OK, but there should be language that provides intuition regarding the 2 approaches (a) match by example and (b) match by rules.

henkbirkholz (approved):
see Ned's comment: https://github.com/ietf-rats-wg/draft-ietf-rats-corim/pull/579#pullrequestreview-4419410083

PR #589, opened 2026-06-04
fix: comid-7.diag example

setrofim:
Replace null (which is invalid, according to the spec) for int-range.max with Inifinity.


setrofim:
I misread the spec. It  re-maps `postive-inf` and `negative-inf` to `null`, so the example is correct.

PR #611, opened 2026-07-14
Clarify that the environment matching rules apply recursively

thomas-fossati:


draft-ietf-rats-corim.md
```diff
@@ -2894,6 +2894,7 @@ Before performing the binary comparison, the processor SHOULD convert the attrib
 If all the attributes which are present in the C-ECT `environment` (e.g., `instance-id` or `group-id`) are also present in the ACS-ECT and are binary identical, the two environments match.
 Otherwise, the environments do not match.
 
+In other words, a match succeeds when the C-ECT's `environment` is contained in the ACS-ECT's `environment` - i.e., for every field path defined in the C-ECT's `environment`, the ACS-ECT's `environment` defines the same path with an equal value.
```
yogeshbdeshpande:
```suggestion
In other words, a match succeeds when the C-ECT's `environment` is contained in the ACS-ECT's `environment` - i.e., for every member defined in the C-ECT's `environment`, the ACS-ECT's `environment` defines the same member with an equal value.
```

yogeshbdeshpande: i think, we need to find a better terminology otherwise, it sounds lik path construction

thomas-fossati: I believe that the term "path" is appropriate.
What we are talking about is effectively the sequence of fields from root to leaf within the tree defined by the environment struct.
This usage is quite common; see for example JSONPath and XPath/XQuery.

Perhaps, instead of "field path" it would be more consistent to use "attribute path", WDYT?

thomas-fossati: In discussion with Henk we agreed that the term "attribute path" will be added to the Terminology section.  It looks like the term is going to be useful in other places where matching logic is described.

thomas-fossati: See [7d75f2e](https://github.com/ietf-rats-wg/draft-ietf-rats-corim/pull/611/commits/7d75f2eeba0bfa13bce81d334797d51078c0aa90)

thomas-fossati: BTW, {{sec-compare-mvm}} and {{sec-compare-element-map}} use the same containment pattern and could be adjusted accordingly.  (I'll prepare a PR off this same branch.)

thomas-fossati: See #613

thomas-fossati:
Closing this as superseded by #613

PR #613, opened 2026-07-15
Define Attribute Path Containment and apply it where appropriate

thomas-fossati:
Extract the common "sub-shape" matching pattern into a glossary definition ("attribute path containment") and use this definition wherever the pattern appears.

Also define "range satisfaction" to anchor the idea of range matching more rigorously.

NOTE: this builds on top of (and further refines) #611


nedmsmith (approved):
LGTM

yogeshbdeshpande (approved):
lgtm

henkbirkholz (approved):

PR #625, opened 2026-08-29
Remove 'unique' wording from identifier descriptions

### LINKED ISSUE #555: Use of the word `unique`
The document is liberal in its use of the word `unique`. In many cases the word can be removed entirely without consequence.

steven-bellock:
Fix #555.

### DIFF
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -274,7 +274,7 @@ Environment-Claim Tuple (ECT):
 The ECT also contains Authority which identifies the entity that authored the ECT.
 
 Instance ID:
-: An identifier of an Environment that is unique to that Environment instance, such as the serial number of a hardware module.
+: An identifier of an Environment instance, such as the serial number of a hardware module.
 See also {{Section 4.2.1 of -eat}}.
 
 Measurement:
@@ -354,7 +354,7 @@ The CDDL definitions in this document follows the naming conventions illustrated
 
 A CoRIM is a collection of tags and related metadata in a concise CBOR {{-cbor}} encoding.
 A CoRIM can be digitally signed with a COSE {{-cose}} signature.
-A tag is a structured, machine-readable data format used to uniquely identify, describe, and manage modules or components of a system.
+A tag is a structured, machine-readable data format used to identify, describe, and manage modules or components of a system.
 
 Tags can be of different types:
 
@@ -367,7 +367,7 @@ Tags can be of different types:
 CoRIM allows for new types of tags to be added in future specifications.
 For example, Concise Trust Anchor Stores (CoTS) ({{-ta-store}}) is currently being defined as a standard CoRIM extension.
 
-Each CoRIM contains a unique identifier to distinguish a CoRIM from other CoRIMs.
+Each CoRIM contains an identifier to distinguish a CoRIM from other CoRIMs.
 
 CoRIM can also carry the following optional metadata:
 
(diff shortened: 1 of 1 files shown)


steven-bellock:
This should, hopefully, not be too controversial. But we will find out in the review.

thomas-fossati:
Follow-up to the discussion at the CoRIM weekly meeting on 09/02/2026.

Regarding replacing "uniqueness" with "collision-resistance", I believe that the definition of "Collision-Resistant Name" in [Section 2 of JWT](https://www.rfc-editor.org/info/rfc7519/#section-2) can be used as-is and applied to all CoRIM identifiers.

steven-bellock:
@thomas-fossati let me file a separate issue for that. Regardless of how `unique` is defined, this pull request can stand apart from that.

draft-ietf-rats-corim.md
```diff
@@ -695,7 +695,7 @@ The following example demonstrates these recommendations for bundling CoRIMs wit
 
 A CoMID tag contains information about hardware, firmware, or module composition.
 
-Each CoMID has a unique ID that is used to unambiguously identify CoMID instances when cross referencing CoMID tags, for example in typed link relations, or in a CoTL tag.
+Each CoMID has an ID that is used to unambiguously identify CoMID instances when cross referencing CoMID tags, for example in typed link relations, or in a CoTL tag.
```
gmandyam: If we remove the uniqueness requirement, can we ensure that the CoMID can be unambiguously identified?

gmandyam (approved):
One comment on the use of the term 'unambiguously', but I don't think it needs to block merge.

draft-ietf-rats-corim.md
```diff
@@ -274,7 +274,7 @@ Environment-Claim Tuple (ECT):
 The ECT also contains Authority which identifies the entity that authored the ECT.
```
yogeshbdeshpande, line 274: A general comment applicable to all references of unique.


We should have a common statement to state that 

All the Identifiers in this document are unique such that they are collision resistant by means of identifier uniqueness or name scoping.

yogeshbdeshpande: I am happy to approve this PR, if we make the statement either in this PR or create an issue (and hence a separate PR) to address the issue!

yogeshbdeshpande (commented):
please see comment

draft-ietf-rats-corim.md
steven-bellock: @yogeshbdeshpande see https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/627.

yogeshbdeshpande: Thank you @steven-bellock.

Just one last question, I assume you have checked all similar instances of keyword `unique`

Just want to avoid repetition of the same PR again and again!

draft-ietf-rats-corim.md
yogeshbdeshpande, line 442: Does 442 also suffer from the same `unique` plague. I think so!

yogeshbdeshpande: What about this one?

yogeshbdeshpande (commented):
Checked and I believe at 2 More places, uniqueness needs tidy up!!!

draft-ietf-rats-corim.md
steven-bellock: I think
>within the context of a CoRIM issuer.

makes it acceptable.

steven-bellock: Yes, it can probably be removed here.

draft-ietf-rats-corim.md
steven-bellock: >Each CoMID has an ID that is used to identify CoMID instances when cross referencing CoMID tags, for example in typed link relations, or in a CoTL tag.

Yes, that seems to work.

yogeshbdeshpande (approved):
LGTM! Thanks for addressing my comments!

thomas-fossati (approved):
Thanks a lot!

nedmsmith (approved):
LGTM

PR #629, opened 2026-09-04
Keys transformation and processing - issue590

### LINKED ISSUE #590: improve DIT/AKT description
We say that additional details about key protection or provisioning can be supplied via EVTs, but it's not clear to me how that would work.
We say that the keys produce endorsements, but what form do these take exactly? Also, is it one per key? This seems to mirror the lack of precise information that we also have in §8.
There are a few SHOULDs here, but none of them explain when it is OK to skip the requirement.

nedmsmith:
PR updates key triple transformation and processing sections. 
See Issue #590.
Also see [PR #28 in draft-ietf-rats-evidence-trans repo. ](https://github.com/ietf-rats-wg/draft-ietf-rats-evidence-trans/pull/28)

### DIFF
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -2943,7 +2943,7 @@ FUNC SERIES-MATCH(acs: ACS, series: SERIES)
 
 Keys relations identify cryptographic keys that require additional key verification steps before being appended to the ACS.
 Processing follows the algorithm as defined in {{sec-match-and-augment}}.
-The `keys` list is iterated for each `key-item` where the `acs::MATCH()` function uses`condition`, if specified, to search the ACS for preexisting conditions.
+The `keys` list is iterated for each `key-item` where the `acs::MATCH()` function uses`condition`, if specified, to search the ACS for preexisting keys.
 Match conditions may return a variety of results including `Evidence-addition-ECT`, `Reference-Value-addition-ECT`, `Endorsement-addition-ECT`, and `Key-addition-ECT` ECTs.
 If `Evidence-addition-ECT`, `Reference-Value-addition-ECT`, or `Endorsement-addition-ECT` are matched, the `element-list.element-map.measurement-values-map.cryptokeys` list is appended to `key-item.addition.key-list`.
 If `Key-addition-ECT` ECTs are matched, its `key-list` is appended to `key-item.addition.key-list`.
--- a/draft-ietf-rats-corim.md
+++ b/draft-ietf-rats-corim.md
@@ -2361,11 +2361,8 @@ The internal representation of Attest Key and Device Identity triples uses the `
 ~~~
 {: #fig-k title="Keys Relation"}
 
-<cref>
-[TODO]
-Specialise condition/addition ECTs.
-Define constraints.
-</cref>
+If `key-item` is used to transform Evidence, `key-item.condition` is omitted since Evidence addition is unconditional.
+Nevertheless, `key-item.addition` is merely staged for addition pending successful key verification checks.
 
 ##### Domain Memberships {#sec-ir-domain-mem}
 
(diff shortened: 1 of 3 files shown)


nedmsmith:
Observe that key-list in Key-addition-ECT could rely on being populated from keys found by the condition criteria. Hence, it may not be necessary to populate key-list in a identity or attest-key triple. However, since the identity and attest-key triples transformation treats the key-list as both condition and addition, the intrep K-ECT is required for both condition and addition. The possibly peculiar side effect is you can't specify a key as a condition without also verifying it. Maybe that is safer?

If a new key triple is defined that has two key lists (condition, addition) then it would be possible to describe path relationships. For example, there could be three key-items in `keys` relation where:
(condition: Key1, addition: Key2)
(condition: Key2, addition: Key3)
(condition: key3, addition: Key4) 

Key1 could be an EE key where Key2 is its issuer, and so forth for Keys 3 and 4. Key4 could be a trust anchor found in a CoTS entry. 

This has the property that path construction can be facilitated by the vendor who generates the corim / attester device. But it could be more flexible describing a path through a graph involving multiple entities. It is not a replacement for a cert path or a trusted route (e.g., nasr) but it seems helpful for defining related principals.

fchinchilla (approved):

draft-ietf-rats-corim.md
```diff
@@ -2778,9 +2740,11 @@ FUNC transform(
     RETURN item
 }
 ~~~
-{: #algo-key-transform title="Key Triple Transformation"}
+{: #algo-key-transform title="Key Triples Transformation"}
 
-Note that keys are added under the authority of the verifier.
+Note that keys are added under the authority of the verifier because the key verification check is performed by the Verifier.
+
+After the `keys` relations are staged, processing commences as decribed in {{sec-proc-keys}}.
```
steven-bellock, line 2747:
```suggestion
After the `keys` relations are staged, processing commences as described in {{sec-proc-keys}}.
```

draft-ietf-rats-corim.md
```diff
@@ -2975,21 +2939,18 @@ FUNC SERIES-MATCH(acs: ACS, series: SERIES)
...
-
-Keys relations identify cryptographic keys that require additional key verification steps.
+##### Processing `keys` Relations {#sec-proc-keys}
 
-Keys are matched with ACS entries by iterating through the `keys` list.
-For each `keys` entry, the condition ECT is compared with an ACS ECT with `cmtype` 1 or 2 (i.e., endorsements or evidence).
-If they match, perform the following steps for each key in the condition ECT `key-list`:
+Keys relations identify cryptographic keys that require additional key verification steps before being appended to the ACS.
+Processing follows the algorithm as defined in {{sec-match-and-augment}}.
+The `keys` list is iterated for each `key-item` where the `acs::MATCH()` function uses`condition`, if specified, to search the ACS for preexisting keys.
```
steven-bellock, line 2946:
```suggestion
The `keys` list is iterated for each `key-item` where the `acs::MATCH()` function uses `condition`, if specified, to search the ACS for preexisting keys.
```

draft-ietf-rats-corim.md
```diff
@@ -2778,9 +2740,11 @@ FUNC transform(
     RETURN item
 }
 ~~~
-{: #algo-key-transform title="Key Triple Transformation"}
+{: #algo-key-transform title="Key Triples Transformation"}
 
-Note that keys are added under the authority of the verifier.
+Note that keys are added under the authority of the verifier because the key verification check is performed by the Verifier.
```
steven-bellock, line 2745:
```suggestion
Note that keys are added under the authority of the Verifier because the key verification check is performed by the Verifier.
```

steven-bellock (changes requested):

steven-bellock (approved):

henkbirkholz (approved):

PR #632, opened 2026-09-08
Accepted environments are also domains

nedmsmith:
Updated language to clarify that accepted environments are also domains.

Note: This PR does not try to effect internal representation such that for every ECT there is a corresponding M-ECT. This would introduce some complexity as M-ECTs can't exist without both a Domain and at least one Member. Whereas, a singleton environment implies a domain without any members. For this wor work, M-ECT.members must be optional. 

An implmementation could construct MATCH() functions such that a DDT processing would satisfy the domain membership requirement if the Environment existed already. 

However, the current wording says "Domain" which is a little slight of hand. 

But if we require full intrep expression of every Environment is also a Domain, then we need processing steps for each transformation type to construct a parallel M-ECT and allow them to have unpopulated members list.


yogeshbdeshpande (approved):
LGTM!

henkbirkholz (approved):

draft-ietf-rats-corim.md
```diff
@@ -1607,8 +1608,12 @@ Consequently, the OS loader is a trustee domain of the OS.
 Alternatively, trust in a peripheral device might depend on trustworthy operation of a peripheral device's bus controller.
 The bus controller is therefore a trustee domain of the peripheral device.
 
-TDTs cannot create domains.
-Instead, TDT processing first checks that a `domain-id` has already been accepted into the ACS before adding trust dependencies.
+TDTs cannot instantiate domains.
+Instead, TDT processing first verifies that a domain-id has already been accepted into the ACS before adding any trust‑dependency triples.
+
+Environments that have been accepted into the ACE are automatically considered Domains.
```
thomas-fossati, line 1644:
```suggestion
Environments that have been accepted into the ACS are automatically considered Domains.
```

draft-ietf-rats-corim.md
```diff
@@ -1607,8 +1608,12 @@ Consequently, the OS loader is a trustee domain of the OS.
...
+
```
```suggestion
Environments that have been accepted into the ACE are automatically considered Domains.
```

thomas-fossati (commented):
LGTM, a few comments inline.

thomas-fossati (approved):

PR #694, opened 2026-09-10
Use GitHub usernames for `CODEOWNERS`

steven-bellock:
Then one does not need to update their e-mail address if it changes.


thomas-fossati (approved):

yogeshbdeshpande (approved):
lgtm

